PR #12, opened 2019-03-15
WIP: design of storage module

zaibon:
- basic explanation of the reponsability of the module
- Start of design of the intefaces of each sub-module

### DIFF
--- a/specs/storage/readme.md
+++ b/specs/storage/readme.md
@@ -55,10 +55,10 @@ type DiskModule interface {
 Responsible to do the capacity planning of the storage pools on top of the disk prepare by the disk sub-module
 
 ```go
-type StoragePool interface{
-    ID() string
-    Path() string
-    Size() string
+type Space struct {
+    ID string
+    Path string
+    Size string
 }
 
 type StoragePoolModule interface {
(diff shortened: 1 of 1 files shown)


specs/storage/readme.md
```diff
@@ -0,0 +1,111 @@
...
+Responsible to discover and prepare all the disk available on a node to be ready to use for the other sub-modules
+
+### Interface
+
+```go
+type DiskModule interface {
+    // walk over all the disks of the node
+    // create a partition table
+    // create a single partition that takes the full size of the disk
+    Prepare() (DiskInfo[],error)
```
muhamadazmy: I think we need to take into consideration a `preparation policy` for example, in case we want to do a raid setup. In that case, a more verbose interface is needed, like what disks to use and which raid level, etc...

muhamadazmy: Also i think, for easier storage namespace management, we can group all HDD disks in one pool, and all SSDs in another, in that case space reservation can go very fast and easy since u only have one storage pool of each type. and space checking can be very easy.

zaibon: Good idea for the preparation policy. That will give us more freedom for the future.

Regarding the grouping of the different disks, what you explain seems like an implementation detail. The interface define here doesn't really take in account disk. The only thing you need to give is a disk type and a size. How the module organize stuff underneath is not something the user needs to worry about.

But indeed what you propose could be a nice approach to it

PR #21, opened 2019-03-28
init system specs

muhamadazmy:


specs/init.md
```diff
@@ -33,13 +33,31 @@ it's own service, make sure it's always running by re-spawning (if needed), a `o
...
+# Configuration
+Each service must have a configuration file that defines how a service should start.
 
-> Do we need to support starting and stopping individual services ? with start/stop commands ?
\ No newline at end of file
+```yaml
+exec: full command line goes here
+oneshot: true/false # default to false
+after: # optional
+ - dependency 1
```
zaibon, line 60: How is the format of this dependency ? is it referring to another configuration file name ?

zaibon: Also maybe we want to have a `before` field, to sort service that depend on the same other service but that still need to be ordered ?

specs/init.md
```diff
@@ -33,13 +33,31 @@ it's own service, make sure it's always running by re-spawning (if needed), a `o
...
+  - Reload is given a specific service name, to reload the config in case the parameter or the env list has changed. Reload is NOT done automatically with `sync`. Must be explicitly called with certain services. 
+
+# Configuration
+Each service must have a configuration file that defines how a service should start.
 
-> Do we need to support starting and stopping individual services ? with start/stop commands ?
\ No newline at end of file
+```yaml
+exec: full command line goes here
+oneshot: true/false # default to false
```
zaibon, line 58: I think we need to think about other restart policy.
We have learned that just plainly restarting a failing service contently can end up in a pretty heavy loop that will consume a lot of resources for nothing.
Something that let us choose to try to restart the service but with a increasing pause between retries ...

specs/init.md
```diff
@@ -33,13 +33,31 @@ it's own service, make sure it's always running by re-spawning (if needed), a `o
 - All services logs are written directly to kernel ring-buffer
   - Optionally later on, one of the daemon can be responsible of reading the logs and push them somewhere else.
 - once a service update it's status, other `waiting` threads (that depends on this one) will get freed to take start.
-## Controlling:
+
+## Controlling
```
zaibon, line 38: I think this section still miss some command to restart a service. 
In the case of an upgrade, after the new binary has been downloaded, we need to be able to restart the service.

zaibon (changes requested):

zaibon:
> Once configuration is validated, a job thread is started for each defined service

Is this idea of having one thread per start job is there to allow parallel start of service and thus reduce the boot time ? I'm just wondering if it's worth it to introduce concurrency here. Concurrent code is always harder to reason about. For such a critical component, simplicity is a must. 
I guess we'll have to benchmark a bit once we start on a POC to see if we can live with sequential flow.

specs/init.md
muhamadazmy: > How is the format of this dependency ? is it referring to another configuration file name ?

yes, that should be another service name



> Also maybe we want to have a `before`

I don't believe we need that, for example
- lets have service A, B, and C
- Service C can has after A
- Service B can has After, A, and C

this mean the order should be as follows
A, C, then B

specs/init.md
muhamadazmy: Yeah, a backoff strategy will be implemented to avoid getting stuck in a busy loop.

specs/init.md
muhamadazmy: Yes, we can add that of course

muhamadazmy:
> > Once configuration is validated, a job thread is started for each defined service
> 
> Is this idea of having one thread per start job is there to allow parallel start of service and thus reduce the boot time ? I'm just wondering if it's worth it to introduce concurrency here. Concurrent code is always harder to reason about. For such a critical component, simplicity is a must.
> I guess we'll have to benchmark a bit once we start on a POC to see if we can live with sequential flow.

Not only that we need to speed up the boot, we also need to 'wait' for each service, in case it's crashed. Another approach is to start them all sequentially (with proper waits in between) and then get in a busy loop trying to wait for each service to see which one has crashed. this will also work, but can increase the boot-time.

zaibon (approved):

PR #22, opened 2019-04-01
Network module

muhamadazmy:
Network module


muhamadazmy:
After a small talk with @delandtj  i think the networking responsibilities should be different as follows:
- The network module/daemon is responsible for the networking of the host, no more.
- CNI is called by the runtime, we don't need to interfere except may be create a new plugin when needed.
- To make this consistent with the way we run VMs, the same plugins still can be invoked by the vm module to setup same network infrastructure. 
- This means of curse, we shouldn't use libvirt anymore for VMs, but @delandtj thinks that we really should do that anyway. I definitely agree with him on that.

specs/network/cases.md
```diff
@@ -0,0 +1,128 @@
...
+    // a vxlan always has an ID
+    id
+}
+
+```
+
+### QUESTIONS:
+
+  - Storage of network objects.
+  It has been said that we will use the 'blockchain' to store 'things'. A network/network resource could very well be such a 'thing', but at the moment there is no clear view on how to do that and how big these 'things' can be. Also, that 'thing' needs to be encrypted to make sure we don't enlarge a possible attack surface towards a tenant network.
```
zaibon: As far as I understand the "blockchain" used to store data won't be a "real" blockchain, but just a bunch of BCDB with some RAFT on top for consensus.
So storage wise we should not have a problem cause BCDB can store big amount of data.
Now for notification, seems we need to think how we could create something that doesn't require all the node too keep doing some pooling on the BCDB, that would requires way to much bandwidth.
Maybe we could open a FR in BCDB to have some kind of publish/subcribe mechanism so any change can be pushed the the required nodes.
Needs further thinking on this.

zaibon:
this branch got out of controle closing in favor of: https://github.com/threefoldtech/zosv2/pull/49

PR #23, opened 2019-04-02
security

muhamadazmy:


zaibon (commented):
That's a good start. 
Some remarks:
- We should not reference JWT here. We didn't decide to go for that solution and I think we should explore other ideas first before settling on something. But the concepts of hierarchic ACL is defensively something we need to have
- This "user space" idea is very interesting. Having an isolated space for a bunch of resources per user is something I like very much.

specs/security.md
```diff
@@ -0,0 +1,26 @@
+# Requirements for security
+
+> A resource is a container,  or a VM
```
zaibon: We can remove this whole paragraph since it assumes there will be an API on the node, this is not the case anymore.

specs/security.md
```diff
@@ -0,0 +1,26 @@
...
+- Any API call that is going to provision a resource that consume cpu, memory, or storage must require a PublicKey.
+- The public-key is coupled with the resource for its entire life-time, only requests that are signed with the private-key part are authorized to operate on the resource.
+- Data on disk is ALWAYS encrypted, per resource. (A key is generated once storage is required and is )
+
+> Another idea, is that the farmer access only allow him to provision a `space`, a space is a cpu,memory,storage that has it's own API.
+The Space API only allow access to the user (owner of the space). The user space API has the endpoints required to provision containers, and or virtual-machines. More or less, a `space` is a reservation of certain capacity on a `single-node`.
+
+
+## Connection Privilege
+The system will never provide a terminal (not in production anyway). The only way to ask for resources from a machine is only via it's API.
```
zaibon: Same here

zaibon (changes requested):

PR #24, opened 2019-04-02
provisioning discussion

muhamadazmy:


specs/provisioning/provisioning.md
```diff
@@ -0,0 +1,29 @@
+# TFGrid provisioning
```
muhamadazmy, line 1: I think this is the one

zaibon: yep I'll already merge this one so we get rid of the branch and can always create new PR to modify the flow if needed

zaibon (approved):

PR #25, opened 2019-04-10
DSL initial specs

muhamadazmy:


specs/dsl.md
```diff
@@ -0,0 +1,150 @@
+# DSL 
+DSL is the *domain specific language* used to communicate with a single zero-os node. The language is designed to
```
zaibon, line 2: We need to rephrase this. There will be no communication to the node itself.
Instead provisioning will be driven from transactions in a blockchain.
The DSL will still be used as a description of what needs to be provision, but the node will just read it from the blockchain. So there is not really an API to create resources.

specs/dsl.md
```diff
@@ -0,0 +1,150 @@
...
+# Content-Type
+The API is probably going to support multiple content-types, json, yaml, etc... in the rest of this document we going to use `yaml` to
+show structures, and objects for readability.
+
+# Philosophy
+Here we gonna list what the DSL is all about and how it should behave. This is an open discussion area. This is basically the ideas on 
+how the system should interface with the world. Hence it's more of the philosophy of zos, rather than the specs. The following rules is
+going to define how the entire system should behave.
+
+- The main entity of the API is a `resource`, you can create, list, inspect and destroy a resource
```
zaibon, line 15: > you can create, list, inspect and destroy a resource

That's not right, the only thing you can do is create or destroy. There is not listing or inspection.
There will be some possibility to get monitoring information for the resources a client have reserved and own. But in no case a user will be able to do a listing of the resources present on a node.

specs/dsl.md
```diff
@@ -0,0 +1,150 @@
...
+- The DSL is a language used to describe a resource on creation, and returned on inspection
+- A resource `create` operation can create a single resource, a single resource can be a single or multiple allocations
+- An allocation limited to storage, network, memory, and cpu.
+- A resource `create` must be atomic. in other words all allocations must pass before the resource is considered created.
+- A resource must have a unique id, and any set of searchable tags for fast find.
```
zaibon, line 20: similar to my last comment, tags are useless since there is not listing possible.

specs/dsl.md
```diff
@@ -0,0 +1,150 @@
...
+- Each allocation, must have unique name (in the same resource)
+- In a single resource, allocations can have dependencies, and can reference each others
+
+General resource structure
+```yaml
+id: <id>  # optional: if left empty a unique one will be generated
```
zaibon, line 26: This is always going to be generated by the node. So it's not really an optional field.

specs/dsl.md
```diff
@@ -0,0 +1,150 @@
...
+            - @data: /var/data
+        ports:
+            - @redis: 6379
+        env:
+            KEY: VALUE
+        entry-point: redis-server 
+        # more options in container allocation
+```
+
+### User Space
```
zaibon, line 110: This concept of space is most probably not going to be used.

PR #27, opened 2019-05-10
Implement Flist module

zaibon:
- [x] Implement Flister interface
  - [x] Mount
  - [x] Umount
- [x] Tests
   - [x] business logic
   - [x] RPC interface
- [x] Implement Zbus interface

### DIFF
--- a/modules/flist/flist.go
+++ b/modules/flist/flist.go
@@ -69,16 +69,9 @@ func (f *flistModule) Mount(url, storage string) (string, error) {
 		storage = defaultStorage
 	}
 
-	rc, err := downloadFlist(url)
+	flistPath, err := f.downloadFlist(url)
 	if err != nil {
 		sublog.Err(err).Msg("fail to download flist")
-		return "", err
-	}
-	defer rc.Close()
-
-	flistPath, err := f.saveFlist(rc)
-	if err != nil {
-		sublog.Err(err).Msg("fail to save flist on disk")
 		return "", err
 	}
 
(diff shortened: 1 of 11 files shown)


modules/flist/flist.go
```diff
@@ -0,0 +1,235 @@
...
+// Mount implements the Flister.Mount interface
+func (f *flistModule) Mount(url, storage string) (string, error) {
+	sublog := log.With().Str("url", url).Str("storage", storage).Logger()
+	sublog.Info().Msg("request to mount flist")
+
+	if storage == "" {
+		storage = defaultStorage
+	}
+
+	rc, err := downloadFlist(url)
```
muhamadazmy: Check how 0-core does it, because we also now cache the flists as well. so we don't download the same flist over and over.

modules/flist/flist.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+	pidPath := filepath.Join(f.pid, rnd) + ".pid"
+	logPath := filepath.Join(f.log, rnd) + ".log"
+
+	args := []string{
+		"-backend", filepath.Join(f.root, "backend", rnd),
+		"-cache", f.cache,
+		"-meta", flistPath,
+		"-storage-url", storage,
+		"-daemon",
```
muhamadazmy, line 97: `daemon` Is that a new flag? I can't find it in g8ufs implementation. Also do we really need to demonize g8ufs ? We can always just spawn the fs, then wait on both the pid file and that the mountpoint is ready.

In current 0-core the way we do it is that we wait for the "mount ready" on fuse stdout, then we know for sure that the mount is ready. I think this approach is better than waiting for 2 seconds

zaibon: > daemon Is that a new flag? I can't find it in g8ufs implementation. Also do we really need to demonize g8ufs ? We can always just spawn the fs, then wait on both the pid file and that the mountpoint is ready.

I believe the have to demonize cause otherwise the g8ufs processes are stopped if the flist module is restarted.

> In current 0-core the way we do it is that we wait for the "mount ready" on fuse stdout, then we know for sure that the mount is ready. I think this approach is better than waiting for 2 seconds

Yep this is something I need to improve, there is a todo in the code for it.

modules/flist/flist.go
zaibon: Indeed, that is the plan, I still have a todo in the downloadFlist function for that.

modules/flist/flist.go
```diff
@@ -0,0 +1,325 @@
...
+		log.Error().Err(err).Str("path", path).Msg("0-fs daemon did not stopped properly")
+		return err
+	}
+
+	// clean up working dirs
+	logPath := filepath.Join(f.log, name) + ".log"
+	backend := filepath.Join(f.root, "backend", name)
+	_ = os.RemoveAll(logPath)
+	_ = os.RemoveAll(backend)
+	_ = os.RemoveAll(path)
```
LeeSmet: why are these errors ignored?

muhamadazmy (approved):

modules/flist/flist.go
```diff
@@ -0,0 +1,328 @@
...
+	return f.saveFlist(resp.Body)
+}
+
+// saveFlist save the flist contained in r
+// it save the flist by its md5 hash
+// to avoid loading the full flist in memory to compute the hash
+// it uses a MultiWriter to write the flist in a temporary file and fill up
+// the md5 hash then it rename the file to the hash
+func (f *flistModule) saveFlist(r io.Reader) (string, error) {
+	tmp, err := ioutil.TempFile(f.flist, "*_flist_temp")
```
muhamadazmy, line 215: Please consider the size of the tmpfs on a zos system, we might not have enough space under /tmp.

modules/flist/flist.go
```diff
@@ -0,0 +1,328 @@
...
+		return "", err
+	}
+
+	hash := fmt.Sprintf("%x", h.Sum(nil))
+	path := filepath.Join(f.flist, hash)
+	if err := os.MkdirAll(filepath.Dir(path), 0770); err != nil {
+		return "", err
+	}
+
+	if os.Rename(tmp.Name(), path); err != nil {
```
muhamadazmy, line 233: os.Rename will not work across devices. since /tmp is usually a different partition or filesystem other than the final flist destination witch is usually a permanent disk partition (or cache) Rename will  fail.

Copying a file though might cause a race condition if u start to mount 2 locations with same flist. This is why in core0 I used to lock the flist file on disk first before i start the download to make sure only one "mount" is downloading the file.

muhamadazmy (changes requested):

modules/flist/flist.go
zaibon: I don't write it into `/tmp` here. The first argument of `TempFile` define the directory to use. Here it is part of the working directory of the module itself.

modules/flist/flist.go
zaibon: Like explained in the previous comment `tmp` file  is not located in `/tmp`. It is created in the same directory as its final location, so the Rename is an atomic operation. So we don't need locks and are safe from race conditions.

muhamadazmy (approved):

LeeSmet (approved):
LGTM

PR #29, opened 2019-05-19
Flist meta

muhamadazmy:


zaibon (changes requested):
That's a good start indeed.
I would like we go further identifying exactly what section of the OCI config spec is going to be allow to be chosen by the user and by 0-OS itself.

So need to go over https://github.com/opencontainers/runtime-spec/blob/master/config.md and https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md

muhamadazmy:
@zaibon yes, idd i was already working on this part right now.

zaibon:
Zerotier is not going to be in the picture either. 
I don't really see a case where we would need host networking. At best we will have a bridge on the host and a network container providing network for other container but all through the bridge. I want to keep isolation at his best as much as possible

zaibon:
It might be we encounter the need at some point during the development, but let's not already design around these low security tricks

specs/container/readme.md
```diff
@@ -38,4 +38,59 @@ type ContainerModule interface {
...
+### Flist meta
+
+Part of the `config.json` entries that can be defined byt the user during the flist build stage or when creating a container:
+
+- process.terminal
+- process.user
+- process.args
+- process.env
+- process.cwd
+- hostname
```
muhamadazmy, line 59: Hostname should be defined in runtime, hence it should be moved to the other section

specs/container/readme.md
```diff
@@ -38,4 +38,59 @@ type ContainerModule interface {
...
+- mounts (for persistent storage)
```
muhamadazmy, line 60: Mounts here probably only defines "targets" a possible data location needed by the containerized application. But mounts (sources) need to be defined again in runtime where u link actual sources to defined targets here.

muhamadazmy:
I think using containerd will hide the low level details of the config.json file. The client has a very high level interface to configure the container, while it still possible to provide a full runtime sepc.

zaibon:
Doesn't matter how high the abstraction is, we still need to know what kind of configuration we need to give to the container.
In containerd they use Opts function to configure things, but the results is the same.

specs/container/readme.md
```diff
@@ -38,4 +38,59 @@ type ContainerModule interface {
...
+
+Next to these field we will also authorized:
+
+- exposed ports (this will allow auto port forward to the application ports automatically if asked)
+
+### Runtime meta
```
muhamadazmy, line 66: I can see u dropped the pre-defined profiles idea from the specs (privileged and unprivileged), any reason why ?

specs/container/readme.md
zaibon: I'm not sure about that. Why exactly this must be chosen by the runtime ?
I think use should be able to choose the hostname of their container and use the hostname to address other containers.

specs/container/readme.md
zaibon: Indeed, mounts define by a user will refers target location inside container and the source is going to be a storage ID or something. The storage is something the user also has to reserve.

specs/container/readme.md
zaibon: I'm not discarding predefined profile, but just privileged and unprivileged is not right. There won't be such a thing a privileged container. All container must be as secure as possible. Maybe some profile can allow certain things for certain workloads. But then it is going to be very limited capabilities used with certain chosen flist.

PR #31, opened 2019-05-20
Container module

zaibon:


zaibon:
I added some link worth reading.
I have been reading quite a bit the code of containerd see what we can leverage from them.
My conclusion on this is we have 2 way of doing it:
- We try to extract the runtime package from containerd and just integrate this one. This package is everything regarding the management of container-shim and runc. So it is the lowest level I think
- We integrate containerd itself as the container runtime. They provide a really nice client to it that make container management super easy. I believe we can plug our 0-fs in replacement of the normal image management so this is probably the easiest way.

So the decision really comes down to how much code we want to bring in. Since we want to try to make 0-OS as low level as possible I would prefer the first solution. But time is also running against us, so the second option is maybe the fastest way to victory. And over time we could strip down stuff we don't need.

muhamadazmy:
@zaibon 

I actually think that leverage on the full containerd is the easiest and fastest approach. And it's probably much easier to just use the client from the container module to interface with the system.

modules/container.go
```diff
@@ -0,0 +1,53 @@
...
+	// will ever be needed since all is gonna be routing based.
+}
+
+type ContainerModule interface {
+	// Run creates and starts a container on the node. It also auto
+	// starts command defined by `entrypoint` inside the container
+	// ns: tenant namespace
+	// name: name of container
+	// flist: flist of container
+	Run(ns, name, flist string, tags, env []string, network NetworkInfo,
```
zaibon: Seems like we could wraps all these argument into a struct instead. The number of argument starts to be a bit much I think

modules/container/container.go
```diff
@@ -0,0 +1,219 @@
...
+func (c *containerModule) Run(ns string, name string, flist string, tags, env []string, network modules.NetworkInfo,
+	mounts []modules.MountInfo, entrypoint string) (id modules.ContainerID, err error) {
+	// create a new client connected to the default socket path for containerd
+	client, err := containerd.New(c.containerd)
+	if err != nil {
+		return id, err
+	}
+	defer client.Close()
+
+	args, _ := shlex.Split(entrypoint)
```
zaibon: What does this return if `entrypoint` is empty ?

modules/container/container.go
```diff
@@ -0,0 +1,219 @@
...
+
+	root, err := c.flister.Mount(flist, "")
+	if err != nil {
+		return id, err
+	}
+
+	defer func() {
```
zaibon: You should not defer Umount here.
You only unmount when you delete the container.

modules/container/container.go
```diff
@@ -0,0 +1,219 @@
...
+		ctx,
+		name,
+		containerd.WithNewSpec(opts...),
+	)
+
+	if err != nil {
+		return id, err
+	}
+
+	defer func() {
```
zaibon, line 232: Should not defer Delete here.

modules/container/container.go
```diff
@@ -0,0 +1,219 @@
...
+		if err != nil {
+			container.Delete(ctx, containerd.WithSnapshotCleanup)
+		}
+	}()
+
+	logs := path.Join("/var/log", ns)
+	os.MkdirAll(logs, 0755)
```
zaibon: need to check the error return by `MkdirAll`

modules/container/container.go
```diff
@@ -0,0 +1,219 @@
...
+		return id, err
+	}
+
+	defer func() {
+		if err != nil {
+			container.Delete(ctx, containerd.WithSnapshotCleanup)
+		}
+	}()
+
+	logs := path.Join("/var/log", ns)
```
zaibon: Not sure we want to writes logs into `/var/log`
My idea is that a module will have a dedicated root directory under which it can do whatever.
This makes it easy to managed the storage used by each module and avoid spreading files all over the tree

So I would create a log directory under the root directory of the container module and use this instead

modules/container/container.go
```diff
@@ -0,0 +1,219 @@
...
+		for {
+			signal := syscall.SIGTERM
+			if trials <= 0 {
+				signal = syscall.SIGKILL
+			}
+			task.Kill(ctx, signal)
+			trials--
+			select {
+			case <-exitC:
+				break loop
```
zaibon, line 340: Doesn't this make the flow go to the loop label and thus restart the loop?

modules/container/container.go
```diff
@@ -0,0 +1,219 @@
...
+	spec, err := container.Spec(ctx)
+	if err != nil {
+		return err
+	}
+
+	if err := container.Delete(ctx); err != nil {
+		return err
+	}
+
+	return syscall.Unmount(spec.Root.Path, syscall.MNT_DETACH)
```
zaibon:
```suggestion
	return c.flister.Umount(spec.Root.Path)
```

modules/flist/flist.go
```diff
@@ -45,6 +45,10 @@ func New(root string) modules.Flister {
 		root = defaultRoot
 	}
 
+	if err := os.MkdirAll(root, 0770); err != nil {
```
zaibon: nice catch :+1:

modules/container.go
muhamadazmy: Yup, was totally thinking the same. will do it

modules/container/container.go
muhamadazmy: it will return an empty []string list

modules/container/container.go
muhamadazmy: we defer here only if an error occurs (container failed to start) so there will not be a delete call afterwards

modules/container/container.go
muhamadazmy: this only happens if an error occurs, we make sure we delete the container if the task start fails (then followed by unmount)

modules/container/container.go
muhamadazmy: Yup, makes sense. will change that.

modules/container/container.go
muhamadazmy: Yup, will also move it to the module start up code.

modules/container/container.go
muhamadazmy: No, it's not a `goto` statement. `break <label>` will break out of the loop labeled `loop` it's used to break nested loops or in this case a select block.

muhamadazmy: https://stackoverflow.com/questions/11104085/in-go-does-a-break-statement-break-from-a-switch-select

modules/container/container.go
muhamadazmy: why u wanna drop the `DETACH` it will make the unmount. this will allow the system to actually perform the unmount once the files are not used anymore. instead of getting an  (filesystem busy) error

modules/container/container.go
zaibon: I'm really not sure about this pattern.
I think defer does evaluate the the value of err when you call defer, not when the function deferred is called.
https://golang.org/doc/effective_go.html#defer
So this means here err will always be false at this point, thus the Umount will never bee called even in the case there is an error when you create the container.

modules/container/container.go
zaibon: Same remark here

modules/container/container.go
zaibon: Look again, I'm not doing a `syscall.Unmount`, but call the `Umount` method of the flister module.

modules/container/container.go
muhamadazmy: I don't believe so, i used this pattern before many times and it works as expected. but let me test again to be sure.

muhamadazmy: Notice also i use a named return. so (err) in all function reference the same err object, hence once the err is actually populated to value it will be detected by error.

muhamadazmy: https://play.golang.org/p/_sKT9UBDwdV

zaibon:
Reading the code a second time, I wonder if we should include the flister module in this module at all.
Instead of passing the URL fo the flist to this module, shoulnd't we instead just pass the path root the rootfs ?
So we keep the separation of concern between module and doesn't introduce dependencies between low level module.

The component responsible to watch the reservation can then use both container and flist module to provision the container.
This also make it easier to handle error, cause now if the flist fail to unmount for some reason at the end of the Delete method. The method will return an error. And we'll endup in a half deleted state hard to fix.

While if the reservation watcher manage both flist and container it can have a better logic regarding failure at different level.

muhamadazmy (approved):

PR #32, opened 2019-05-21
Upgrade module

zaibon:


specs/upgrade/README.md
```diff
@@ -0,0 +1,79 @@
...
+```
+
+After upgrade:
+
+```
+root
+├── bin
+│   ├── containerd
+│   ├── flist_module_0.1.0
+│   └── runc
```
muhamadazmy, line 78: where did runc come from? I think it's missing from first block above

specs/upgrade/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+0-OS is design to provide maximum uptime for its workload, so rebooting a node should never be required to upgrade any of its component (except when we will push a kernel upgrade).
+
+The only way to get code onto a 0-OS is using flist. An upgrade flist will be composed of the new binary to install and in some case a migration binary that needs to be run before the start of the new module.
+
+The content of the upgrade flist will be cached on disk, so in the event of a power failure, the node can just restart and restart all the workloads without downloading all the new modules again.
+
+## Flist upgrade layout
+
+The files in the upgrade flist needs to be located in the filesystem tree at the same destination they would need to be in 0-OS. This allow the upgrade code to stays simple and only does a copy from the flist to the cache disk of 0-OS.
```
muhamadazmy, line 35: Copying to the cache desk is not enough IMHO. Since the base 0-fs image is a tmpfs image that is fully loaded in memory, we need to make sure the new binaries overrides the one in the base image we can do this by one of those ways:
1- Binaries need to be copied to the `roofs` of the node, this means the files need to be copied every time the machine is boated
2- May be override the PATH variable to always prefer a (/var/cache/bin) for example over (/bin) but this can be tricky in case a library upgrade is required so we will also need LD_LIBRARY_PATH. this can get messy
3- Actually install zos on the a desk on first boot, then update the files in place.

> Talking about libraries: updating binaries without the libraries will not work (unless we grantee we always have static builds). I think instead of publishing a single flist with the updated binaries a full zos image should be published and the node then can only copy the files that need to be update (comparing checksum with local files manifest, or a tool that diff from the current 0-os image) . A node that is freshly booted can just boot from the latest build.

zaibon: > Copying to the cache desk is not enough IMHO

There is a confusion here. I used bad wording. I'm not talking about the 0-fs cache. But a disk in which 0-OS has installed what it needs to do its job.

> Binaries need to be copied to the roofs of the node, this means the files need to be copied every time the machine is boated

No I want to avoid this, all the components needs to be locally available, so if a node reboot and doesn't have internet connectivity it still can run properly

> updating binaries without the libraries will not work (unless we grantee we always have static builds). 

Of course this example only present a component that is statically build. I would like to push static binaries as much as possible, but for tools that cannot be shipped static we'll have to also distribute the libraries.

> I think instead of publishing a single flist with the updated binaries a full zos image should be published and the node then can only copy the files that need to be update (comparing checksum with local files manifest, or a tool that diff from the current 0-os image)

I don't think this is necessary except maybe for the first ever release. Then update flist can only contains the new stuff. This makes update logic so much easier if you know everything in the flist needs to be copied over.

specs/upgrade/README.md
zaibon: indeed

specs/upgrade/README.md
muhamadazmy: So does this mean zos will now require installation ?

zaibon: This is all going to be automatic of course, but yeah like today, we'll have to do some steps before the node is fully operational. There is no way around it.
These steps includes:
- reserve a disk
- do a capacity check
- make sure all modules have latest version installed
- ...

modules/upgrade/file.go
```diff
@@ -0,0 +1,28 @@
+package upgrade
+
+import (
+	"os"
+	"path/filepath"
+)
+
+var ModeExecutable os.FileMode = 0111
```
LeeSmet: I think the correct filemode should be 0100, since the permissions are not actually a hierarchy (assuming it is an executable and not a script, since a script needs to be readable too iirc)

modules/upgrade/publisher.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	log.Info().Msg("upgrade found")
+	return upgrade, nil
+}
+
+func joinURL(base, path string) (string, error) {
+	u, err := url.Parse(base)
+	if err != nil {
+		return "nil", err
```
LeeSmet: Is there a reason that the default value for strings ("") is not returned here?

modules/upgrade/upgrade.go
```diff
@@ -0,0 +1,384 @@
...
+	return &UpgradeModule{
+		version: version,
+		root:    root,
+		flister: flister,
+	}, nil
+}
+
+// FIXME: not sure about the public interface of this package yet
+// most probably will need to run as a daemon too
+func (u *UpgradeModule) Run(period time.Duration, p Publisher) error {
```
LeeSmet: It seems this function never returns, so this error return can go

LeeSmet (changes requested):

modules/upgrade/publisher.go
zaibon: good catch

modules/upgrade/upgrade.go
zaibon: I'm still actually a bit unhappy about this function.
Don't you think I should maybe remove the infinite loop from the Run method, and move the recurring logic part of the main file ?
So this package only holds logic about upgrade itself. And it's up to the caller to decide how often to run it ?

modules/upgrade/file.go
zaibon: Indeed, the last 1 is probably not relevant. `0110` is enough I think, only test if the current user or group can execute ?

LeeSmet: Don't think that is 100% correct either, since it will mark a file which as `0010` permissions as executable, but if the file is owned by your user, it is actually not. Tested this a bit locally, and a file owned by my user can indeed not execute some other file I own if its marked as `0077`. 

This answer on stackexchange seems to give a pretty detailed explanation https://unix.stackexchange.com/a/411901

modules/upgrade/upgrade.go
LeeSmet: I'm not sure, the period is already decided by an argument. So the question is more, would someone ever want to not call all these actions together I guess?

zaibon: yeah but if I remove the infinite loop, then I also remove the period argument.
And the Run method is actually the only think to run when you want to check/apply a new upgrade.

LeeSmet: True, I misunderstood your earlier comment, though you wanted to remove this function and just expose the individual functions that it calls. I think that is indeed acceptable.

PR #46, opened 2019-06-11
Add meeting from 11/06/2019

maxux:


modules/container/container.go
```diff
@@ -99,8 +99,8 @@ func (c *containerModule) Run(ns string, data modules.Container) (id modules.Con
 	if data.Interactive {
```
zaibon: This file has nothing to do in this PR, please remove

zaibon (changes requested):

delandtj (approved):
Can someone also research how k8s implements and uses these pause containers ?

PR #65, opened 2019-06-18
Apply network resource Network

### LINKED ISSUE #61: network: logic to apply network resources
We need to implement the logic that applies a network resource.

zaibon:
fixes #61


modules/network/network_test.go
```diff
@@ -46,13 +57,94 @@ func TestCreateBridge(t *testing.T) {
...
+				Connected: []modules.Connected{
+					{
+						Type:   modules.ConnTypeWireguard,
+						Prefix: MustParseCIDR("2a02:1802:5e:cc02::/64"),
+						Connection: modules.Wireguard{
+							NICName:  "cc02",
+							Peer:     net.ParseIP("2001::1"),
+							PeerPort: 51820,
+							Key:      "4w4woC+AuDUAaRipT49M8SmTkzERps3xA5i0BW4hPiw=",
+							// LinkLocal: net.
```
delandtj: fe80::cc01/64 to add for testing ?

modules/network/networker.go
```diff
@@ -1,27 +1,200 @@
...
+func configureWG(storageDir, network string, netRes modules.NetResource) error {
+	var (
+		netnsName   = netnsName(network)
+		wgName      = wgName(network)
+		storagePath = filepath.Join(storageDir, network)
+		key         wgtypes.Key
+		err         error
+	)
+
+	key, err = wireguard.LoadKey(storagePath)
```
delandtj, line 206: what if the error is a storage error ?

modules/network/networker.go
```diff
@@ -1,27 +1,200 @@
...
+		if conn.Peer.To16() != nil {
+			endpoint = fmt.Sprintf("[%s]:%d", conn.Peer.String(), conn.PeerPort)
+		} else {
+			endpoint = fmt.Sprintf("%s:%d", conn.Peer.String(), conn.PeerPort)
+		}
+
+		peers[i] = wireguard.Peer{
+			PublicKey:  conn.Key,
+			Endpoint:   endpoint,
+			AllowedIPs: []string{"0.0.0.0/0"},
```
delandtj: AllowedIPs has to contain: the fe80::/ of the WG, the Prefix of the Network Resource on the bridge, an IPv4 LL addr and an IPv4 subnet on the bridge. No peer can heve the same IP (so certainly not 0/0)

modules/network/wireguard/wireguard.go
```diff
@@ -141,3 +163,28 @@ func newPeer(pubkey, endpoint string, allowedIPs []string) (wgtypes.PeerConfig,
 
 	return peer, nil
 }
+
+func GenerateKey(dir string) (wgtypes.Key, error) {
```
delandtj, line 183: hmmm... you have multiple error types emanating from that function, how to handle them ?

modules/network/networker.go
zaibon: Well in any case, If I cannot read the key from the disk, would it be because it has not been generated before or storage error, I need to create a new one.
But this means I need a way to update the resource unit when things like that happens, cause all the other node will have to upgrade their wg config with the new public key.

modules/network/networker.go
zaibon: But don't you need this for the wg interface to be able to forward all traffic in/out ? I though that without this the interface can actually only reach other peer. But the exit not cannot send public traffic to it

modules/network/wireguard/wireguard.go
zaibon: Well it up to the caller to decide what to do when something fails.

modules/network/networker.go
delandtj: Hmm.. 
I'll setup testnet.sh to reflect that properly

PR #69, opened 2019-06-20
setupwgtest.sh and User Network Object

delandtj:
- added specifications for reachable peers in network.go
  - added a container behind NRs to verify correct routing/
  forwarding.

### DIFF
--- a/modules/network.go
+++ b/modules/network.go
@@ -18,9 +18,35 @@ type Networker interface {
 
 // NetID is a type defining the ID of a network
 type NetID string
+// Type4 is the Node's IPv4 reachability:
+type Type4 int
 
+// Type6 is the Node's IPv6 reachability
+type Type6 int
+const (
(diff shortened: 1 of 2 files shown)


modules/network.go
```diff
@@ -18,9 +18,35 @@ type Networker interface {
 
 // NetID is a type defining the ID of a network
 type NetID string
+// Type4 is the Node's IPv4 reachability:
+type Type4 int
 
+// Type6 is the Node's IPv6 reachability
+type Type6 int
```
zaibon: Instead of Type, I would call this Reachability or Visibility or something more explicit. 
Specially that this live in the modules package so we have all sorts of type in there.

zaibon (changes requested):
Mostly LGTM , just one remark regarding the nomenclature of a type

modules/network.go
delandtj: Eh.. your Code, your names ;-) but make a distinction 4-6

PR #72, opened 2019-06-20
service configuration for zinit

muhamadazmy:


conf/readme.md
```diff
@@ -0,0 +1,14 @@
+
+## Services Boot Sequence
+- node-read
```
zaibon:
```suggestion
- node-ready
```

conf/readme.md
```diff
@@ -0,0 +1,14 @@
...
+  - storaged
+  - netwokrd
+- boot
+  - contd
+  - flister
+  - upgrade
+
+## Pseudo boot steps
+both node-read and boot are not actual services, but instead they are there to define a `boot stage`. for example once `node-ready` service is (ready) it means all crucial system services defined by 0-initramfs are now running. This now only contains the `udev` daemons, but it's easy to add more without the need to change all the other services.
```
zaibon:
```suggestion
both node-ready and boot are not actual services, but instead they are there to define a `boot stage`. for example once `node-ready` service is (ready) it means all crucial system services defined by 0-initramfs are now running. This now only contains the `udev` daemons, but it's easy to add more without the need to change all the other services.
```

zaibon (changes requested):

muhamadazmy:
@zaibon I got all basic services running inside the zosv2 image. Currently i am still using overlay to make sure the config (and binaries) from this branch are linked correctly into the boot image.

Currently the netword and storaged are not building correctly yet. So once these modules are working it will be easy to add the code to build and copy the config in the 0-initramfs image.

I also made sure that the default `-broker` flags have decent defaults. and fixed a bug in the upgraded module where it was using the wrong zinit socket.

conf/boot.yaml
```diff
@@ -0,0 +1,10 @@
+# ready is a pseudo boot stage to mark that all
```
zaibon:
```suggestion
# boot is a pseudo boot stage to mark that all
```

conf/node-ready.yaml
```diff
@@ -0,0 +1,15 @@
+# ready is a pseudo boot stage to mark that all
```
zaibon:
```suggestion
# node-ready is a pseudo boot stage to mark that all
```

conf/boot.yaml
zaibon: Also the rest of the comment doesn't make much sens for this pseudo stage

conf/node-ready.yaml
```diff
@@ -0,0 +1,15 @@
...
+# have been booted correctly
+
+exec: "true"
+oneshot: true
+after:
+  # list of all services that must be running before moving on
+  # note, these services are defined by the 0-initramfs repo
+  - local-modprobe
+  - udev-trigger
+  - redis
```
zaibon: duplicate from line 12.
Also I wonder if Redis should be there.
Redis is the message broker for zbus. I can change at some point. Shouldn't we move it out of here. Since these are really more hardware setup oriented.

zaibon (changes requested):

conf/flistd.yaml
```diff
@@ -0,0 +1,3 @@
+exec: flistd -broker unix://var/run/redis.sock
```
zaibon: Let stays consistent. Or we specify it for all modules daemon or for none.
Also, since some of the module requires to have some cache fs. We should maybe also set the `-root` flag of some modules daemon

zaibon (changes requested):

conf/node-ready.yaml
muhamadazmy: that's exactly the reason why it should be part of zos2 not initramfs. because we can just remove it from here (and add the new bus) and push to have a new image.

The upgrader can then take care of updating the config of running machines.

conf/flistd.yaml
muhamadazmy: yeah, i agree i already provided it to all modules.

PR #75, opened 2019-06-21
Go storage

LeeSmet:
In this first version, the storage modules initializes the storge pools according to a predefined policy (can be set using flags). Possible policies are: single (linear), raid1 and raid10. It's also possible to define how much volumes should be created (by default 0 -> as much as possible), and how much disks to use per volume. Volumes of multiple devices are created only on the same type (ssd/hdd), and the cache is attempted to be mounted on an ssd volume, if possible


modules/storage/filesystem/device.go
```diff
@@ -0,0 +1,186 @@
...
+
+	if len(devices.BlockDevices) != 1 {
+		return device, fmt.Errorf("device not found")
+	}
+
+	return devices.BlockDevices[0], nil
+}
+
+func (l *lsblkDeviceManager) PoolType(ctx context.Context, pool Pool) (DeviceType, error) {
+	devices, err := l.Devices(ctx)
```
muhamadazmy: You could have just used `l.WithLabel(ctx, pool.Name())`

modules/storage/storage.go
```diff
@@ -0,0 +1,287 @@
...
+}
+
+/**
+Initialize, must be called at least onetime each boot.
+What Initialize will do is the following:
+ - Try to mount prepared pools (if they are not mounted already)
+ - Scan free devices, apply the policy.
+ - If new pools were created, the pool is going to be mounted automatically
+**/
+func (s *storageModule) Initialize(policy modules.StoragePolicy) error {
```
muhamadazmy: As we agreed on the chat. I really think it's better if we split this into 2 parts
- The initialize part, only called on the module start, and not exposed on zbus which mounts and loads the exiting pools and ensuring the cache is mounted.
- Another method for the creation of new pools, given the policy.

modules/storage/filesystem/device.go
```diff
@@ -0,0 +1,186 @@
...
+)
+
+// Device represents a physical device
+type Device struct {
+	Type       string   `json:"type"`
+	Path       string   `json:"name"`
+	Label      string   `json:"label"`
+	Filesystem FSType   `json:"fstype"`
+	Children   []Device `json:"children"`
+	DiskType   DeviceType
```
muhamadazmy: Please add the json tags for consistency, if u want it getting filled up from the lsblk output put `json:"-"`.

modules/storage/storage.go
zaibon: Indeed initialize should not be expose over zbus if the module calls it automatically when it starts

modules/storage/filesystem/btrfs.go
```diff
@@ -0,0 +1,343 @@
...
+
+		pools = append(pools, btrfsPool(fs.Label))
+	}
+
+	return pools, nil
+}
+
+type btrfsPool string
+
+// Mounted checks is the pool is mounted
```
zaibon:
```suggestion
// Mounted checks if the pool is mounted
```

modules/storage/filesystem/btrfs.go
```diff
@@ -0,0 +1,343 @@
...
+
+	return "", false
+}
+
+func (p btrfsPool) Name() string {
+	return string(p)
+}
+
+func (p btrfsPool) Path() string {
+	return path.Join("/mnt", string(p))
```
zaibon:
```suggestion
	return filepath.Join("/mnt", string(p))
```

modules/storage/filesystem/btrfs.go
```diff
@@ -0,0 +1,343 @@
...
+}
+
+func (p btrfsPool) enableQuota(mnt string) error {
+	_, err := run(context.Background(), "btrfs", "quota", "enable", mnt)
```
zaibon, line 163: aren't you supposed to set the value of the quota too ?

modules/storage/filesystem/btrfs.go
```diff
@@ -0,0 +1,343 @@
...
+
+	var volumes []Volume
+
+	subs, err := BtrfsSubvolumeList(context.Background(), mnt)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, sub := range subs {
+		volumes = append(volumes, btrfsVolume(path.Join(mnt, sub.Path)))
```
zaibon:
```suggestion
		volumes = append(volumes, btrfsVolume(filepath.Join(mnt, sub.Path)))
```

modules/storage/filesystem/btrfs.go
```diff
@@ -0,0 +1,343 @@
...
+	return volumes, nil
+}
+
+func (p btrfsPool) AddVolume(name string) (Volume, error) {
+	mnt, ok := p.Mounted()
+	if !ok {
+		return nil, DeviceNotMountedError
+	}
+
+	mnt = path.Join(mnt, name)
```
zaibon:
```suggestion
	mnt = filepath.Join(mnt, name)
```

modules/storage/filesystem/btrfs.go
```diff
@@ -0,0 +1,343 @@
...
+func (v btrfsVolume) Volumes() ([]Volume, error) {
+	var volumes []Volume
+
+	subs, err := BtrfsSubvolumeList(context.Background(), string(v))
+	if err != nil {
+		return nil, err
+	}
+
+	for _, sub := range subs {
+		volumes = append(volumes, btrfsVolume(path.Join(string(v), sub.Path)))
```
zaibon:
```suggestion
		volumes = append(volumes, btrfsVolume(filepath.Join(string(v), sub.Path)))
```

modules/storage/filesystem/device.go
```diff
@@ -0,0 +1,186 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"time"
+
+	log "github.com/rs/zerolog/log"
+)
+
+type DeviceManager interface {
+	Device(ctx context.Context, device string) (Device, error)
```
zaibon, line 15: Please document how each method of the interface should work and what's its purpose

modules/storage/filesystem/device.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	if err := json.Unmarshal(bytes, &devices); err != nil {
+		return nil, err
+	}
+
+	return setDeviceTypes(flattenDevices(devices.BlockDevices)), nil
+}
+
+func (l *lsblkDeviceManager) WithLabel(ctx context.Context, label string) ([]Device, error) {
```
zaibon: I found `WithLabel` confusing name, isn't `ByLabel` better ?

modules/storage/filesystem/filesystem.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+// Usage struct
+type Usage struct {
+	Size uint64
+	Used uint64
+}
+
+// Volume represents a logical volume in the pool. Volumes can be nested
+type Volume interface {
```
zaibon, line 16: Please document each method of the interface

modules/storage/filesystem/filesystem.go
```diff
@@ -0,0 +1,43 @@
...
+	AddVolume(name string) (Volume, error)
+	RemoveVolume(name string) error
+	Usage() (Usage, error)
+	Limit(size uint64) error
+	Name() string
+	FsType() string
+}
+
+// Pool represents a created filesystem
+type Pool interface {
```
zaibon, line 36: Please document each method of the interface

modules/storage/filesystem/utils.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	err = json.Unmarshal(bytes, &seekTime)
+	return seekTime.Typ, seekTime.Time, err
+}
+
+func run(ctx context.Context, name string, args ...string) ([]byte, error) {
+	output, err := exec.CommandContext(ctx, name, args...).Output()
+	if err != nil {
+		if err, ok := err.(*exec.ExitError); ok {
```
zaibon, line 64: what if the error is not of type `exec.ExitError`? then you never return the error at all

modules/storage/storage.go
```diff
@@ -0,0 +1,287 @@
...
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/rs/zerolog/log"
+	"github.com/threefoldtech/zosv2/modules"
+	"github.com/threefoldtech/zosv2/modules/storage/filesystem"
+)
+
+const (
+	cacheTarget = "/var/cache"
```
zaibon, line 16: @muhamadazmy @delandtj should we keep this path ?

modules/storage/storage.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+
+	log.Debug().Msgf("Mounting cache partition in %s", cacheTarget)
+	return filesystem.BindMount(cacheFs, cacheTarget)
+}
+
+// createFs creates a filesystem with the given name and limits it to the given size
+// if the requested disk type does not have a storage pool available, an error is
+// returned
+func (s *storageModule) createFs(size uint64, name string, preferSSD bool) (filesystem.Volume, error) {
```
zaibon: What did we called this `createFs`, when it actually creates a sub volume ?

modules/storage/storage.go
```diff
@@ -0,0 +1,287 @@
...
+			} else {
+				hddPools = append(hddPools, pool)
+			}
+		}
+		possiblePools = append(ssdPools, hddPools...)
+	}
+
+	// for now take the first volume, if an ssd is prefered ssds will be sorted
+	// first
+	for _, volume := range possiblePools {
```
zaibon, line 288: I don't see where we verify that the selected volume still have enough free space.

modules/storage/storage.go
```diff
@@ -0,0 +1,287 @@
...
+// returned
+func (s *storageModule) createFs(size uint64, name string, preferSSD bool) (filesystem.Volume, error) {
+	var fs filesystem.Volume
+	var err error
+
+	possiblePools := s.volumes
+	if preferSSD {
+		ssdPools := []filesystem.Pool{}
+		hddPools := []filesystem.Pool{}
+		for _, pool := range s.volumes {
```
zaibon, line 275: Can't we avoid to go back to the disks to get the PoolType ?
I would like we try to let the disks alone as much as possible.
So when the storage module boots, it does a scanning of all the disks. Then keeps the required info in memory, so we can completely turn off all the unused disk.

zaibon (changes requested):
Mostly looks good.

Let's make sure all the warning from go vet and linting are fixed.
Also always prefer `filepath` package over just `path` when manipulation file paths.

Biggest concern with the current code is we still go too much to the disk to get information we could already know (like type of disks, ...)

I would also like to have a documentation on the different path/location used for everything (cache, mountpoints, ....)

modules/storage/filesystem/btrfs.go
LeeSmet: No, afaik the value of the quota is set when the Limit method is called, which creates an associated qgroup. This method is called when a pool is mounted, and btrfs will find the associated qgroup and apply the limits.

@muhamadazmy can you confirm this?

muhamadazmy: @LeeSmet yes, that' correct. the enable quota method will only enable quota support for this filesystem. otherwise we can't create qgroups. then later, when we are creating a subvolume, we pass the limit.

modules/storage/storage.go
muhamadazmy: I have no attachment to this path. Any other path would work. the question is, is there a reason to move away from this path?

modules/storage/filesystem/btrfs.go
zaibon: Got it, thx

modules/storage/storage.go
muhamadazmy: it's a compromise between the following:
- Caching
- Performance
- Number of operations

Caching, is usually problematic, and can go easily out of sync. I already hate that we keep the available pools in memory, since it can go out of sync with the system if another cache (some how) was created or delete behind the storage module back.

Performance is a good concern, going to the disk is indeed much slower than checking some cached in memory value, but this brings us to the last point.

Number of operations, basically how often do we need to do this ? it's only on creating a new subvolume which i think is not that many operations per hours. 

I would actually recommend we drop all memory state all together, and make sure everything is always checked during runtime.

modules/storage/storage.go
zaibon: well I want to avoid to have stuff spread all over the place, and I wanted to try to keep each modules have a clean separation.  So we don't really allow modules to write anywhere in the filesystem, but instead just in one directory only.

I think we should document all the path/location used by everything we have, once we have a clear view we can decide where to put what.

modules/storage/storage.go
muhamadazmy: Regarding turning off all unused disks, once the disk is mounted we can't turn it off. Only unformatted disks can be safely turned off. and this can be done only one time on storage module booting.

Then on creating new pool, we will have to check all "free" disks to get the type. then create the pool, and turn the remaining disks off again.

modules/storage/storage.go
zaibon: https://github.com/threefoldtech/zosv2/issues/79

modules/storage/storage.go
zaibon: The actual problem with the current code, is that we don't cache the disk type. And since we actually do some write to the disk to determine its type (seektime) then every time we create a new pool, we have to turn on the disk, execute seektime. This is quite redundant since the type of a disk will never change, Once you know it, you know it forever.

I agree with you that caching is a bitch and I would like to avoid it too.
Keeping the pools in memory doesn't really bring anything here since the type is not saved.

> since it can go out of sync with the system if another cache (some how) was created or delete behind the storage module back.

Regarding this, the real problem is that if we go out of sync, that means or we have a bug in our code, or the node is compromised. Since the storage module MUST be the only one dealing with disks...

muhamadazmy: seektime does not write to the disk. it only reads random blocks and calculate the time between the seeks to the random position. Just saying.

But anyway, since we generate the pools names, we can save some data in the pool label to extract such information without actually storing it anywhere.

so a pool label can be (instead of only a guid) be in a format like `type:id` instead. so pool type basically can be extracted from name.

@LeeSmet what do u think ?

zaibon: > seektime does not write to the disk. it only reads random blocks and calculate the time between the seeks to the random position. Just saying.

Correct, but my point stays, the disk need to be active for that.

> so a pool label can be (instead of only a guid) be in a format like type:id instead. so pool type basically can be extracted from name.

That's a nice idea actually. That would avoid caching and solve the type issue at the same time

LeeSmet: Imo, if we want the storage module to be the only thing managing disks, then caching should be fine . If the cached info is out of date with the disks that can mean 2 things: a bug in the storage module (forgot to rescan disks after modification), or something else performed an action on the disks, which is then also considered a bug as storage should be the only thing doing that. Labeling without caching would mean we need to scan for filesystems every time, does this not have a performance impact?

modules/storage/filesystem/device.go
```diff
@@ -18,8 +17,9 @@ type DeviceManager interface {
 	Devices(ctx context.Context) ([]Device, error)
 	// ByLabel finds all devices with the specified label
 	ByLabel(ctx context.Context, label string) ([]Device, error)
-	// PoolType finds the type of a storagepool
-	PoolType(ctx context.Context, pool Pool) (DeviceType, error)
+	// Scan the system for devices. This must be called after all actions which
```
zaibon, line 20: > This must be called after all actions which make persistent changes on the disk layout.

Why do you have to scan every disk any time you do something on them ?
This doesn't solve the problem. The goal was to avoid to have to read all the disk every time to know their types. 
Seems now you keep that info in memory. But then now we have a scan function that needs to be called every time? This solution still doesn't allow us to bring the unused disk down.

zaibon (approved):
Ok looks good.
I'm still not convince by the approach taken to solve the disk type issue. But let's already merge this so we don't block storage any longer. 
But @LeeSmet please open an issue regarding disk caching so we can improve this in a later PR.

modules/storage/filesystem/device.go
LeeSmet: We can avoid this by manually setting labels on disks/partitions which get included in filesystems, setting the device children if we partition it, ...

LeeSmet: @zaibon https://github.com/threefoldtech/zosv2/issues/84

zaibon:
@LeeSmet before we merge, ca you fix the linting error that comes from the storage module please ? (https://travis-ci.com/threefoldtech/zosv2/builds/116983451)
There are other error but they don't come from your code, so we can skip them for now. Thx :-)

LeeSmet:
linter should be good now :)

zaibon:
@LeeSmet @muhamadazmy merged ! Good work :+1:

PR #82, opened 2019-06-26
Provisiond

muhamadazmy:
## Provisioning daemon

- Provision daemon pulls reservations request from configured sources. Normally when this is about to be production ready should get reservation from bcdb. Currently we have 2 sources we use for debugging.
  - A fifo located at `/var/run/reservation.pipe` u can push a `json` string as defined by the `Reservation struct` 
  - A http poll source that pulls for reservation on a configurable http endpoint.
- Note that, we currently have no way to return the reservation results (success, error) or the data or provisioned items. We only print out the results in the logs. We need to report this back to 3bot at some point, but this area still shady.

## Example:
```bash
echo '
{
    "type": "volume",
    "tenant": "azmy",
    "data": {"size": 3221225472}
}' > /var/run/reservation.pipe

# you have to read the result of this reservation from the logs
# then use it in the volume-id below
# currently for demo purposes only the volume-id is just the path to the volume
# later on this gonna be an actual id to grantee ownership of the volume

echo '
{
    "type": "container",
    "id":"first", 
    "tenant": "azmy", 
    "data": {
        "flist": "https://hub.grid.tf/thabet/redis.flist",
        "entrypoint": "redis-server", 
        "mounts": [
            {
                "volume-id": "<place volume id here>",
                "mountpoint": "/opt/data"
            }
        ]
    }
}' > /var/run/reservation.pipe
```


cmds/provisiond/main.go
```diff
@@ -0,0 +1,57 @@
...
+	"os"
+
+	"github.com/rs/zerolog"
+	"github.com/rs/zerolog/log"
+
+	"github.com/threefoldtech/zbus"
+	"github.com/threefoldtech/zosv2/modules/provision"
+)
+
+const module = "container"
```
zaibon:
```suggestion
const module = "provision"
```

modules/provision/container.go
```diff
@@ -0,0 +1,97 @@
...
+			// TODO:
+			//   Network: this requires network module
+			//   Mounts: this required storage module
+			Entrypoint:  config.Entrypoint,
+			Interactive: config.Interactive,
+		},
+	)
+
+	if err != nil {
+		flistClient.Umount(mnt)
```
zaibon: Let's at least log the eventual Umount error

muhamadazmy: done

cmds/provisiond/main.go
muhamadazmy: I removed the const, it's not used since provisiond is not a zbus service. it only uses the client to do it's work

PR #88, opened 2019-06-27
Close #84: Cache devices in memory

### LINKED ISSUE #84: Storage: disk caching
Right now the storage module does a scan of the disks every time something changes the disk layout (filessytem creation, in the future maybe partitioning, ...). This prevents us from bringing down the disks completely. 

It should be possible to manually modify the in memory device when these actions are done, allowing us to shut down the disks and not rescan

LeeSmet:
Also prefer to use only indexing in loops over slices, to avoid
unneeded copies

### DIFF
--- a/modules/storage/filesystem/device.go
+++ b/modules/storage/filesystem/device.go
@@ -12,15 +12,15 @@ import (
 // DeviceManager is able to list all/specific devices on a system
 type DeviceManager interface {
 	// Device returns the device at the specified path
-	Device(ctx context.Context, device string) (Device, error)
+	Device(ctx context.Context, device string) (*Device, error)
 	// Devices finds all devices on a system
-	Devices(ctx context.Context) ([]Device, error)
+	Devices(ctx context.Context) (DeviceCache, error)
 	// ByLabel finds all devices with the specified label
-	ByLabel(ctx context.Context, label string) ([]Device, error)
-	// Scan the system for devices. This must be called after all actions which
-	// make persistent changes on the disk layout.
-	Scan(ctx context.Context) error
+	ByLabel(ctx context.Context, label string) (DeviceCache, error)
 }
+
+// DeviceCache represents a list of cached in memory devices
+type DeviceCache []*Device
 
 // FSType type of filesystem on device
 type FSType string
@@ -51,7 +51,7 @@ type Device struct {
 	ReadTime   uint64     `json:"-"`
 }
 
-// Used assumes that the device is used if it has custom label of fstype
+// Used assumes that the device is used if it has custom label or fstype or children
 func (d *Device) Used() bool {
 	return len(d.Label) != 0 || len(d.Filesystem) != 0 || len(d.Children) > 0
 }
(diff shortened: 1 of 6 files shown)


modules/storage/storage.go
```diff
@@ -33,9 +33,17 @@ type storageModule struct {
 
 // New create a new storage module service
 func New() modules.StorageModule {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
+	defer cancel()
+
+	m, err := filesystem.DefaultDeviceManager(ctx)
+	if err != nil {
+		panic(err)
```
muhamadazmy: Don't panic, instead return the error, then the `main()` can panic or `log.Fatal`

modules/storage/storage.go
```diff
@@ -203,15 +203,15 @@ func (s *storageModule) CreateFilesystem(size uint64) (string, error) {
 func (s *storageModule) ReleaseFilesystem(path string) error {
```
muhamadazmy, line 201: I think this method logic can be improved by extracting the Pool path from the `path` param. Then only scan this pool for the subvolume, instead of going over all the pools in a loop.
A filesystem path is probably somehting like `/mnt/<pool-id>/<subvolume-id>` . So we already know the pool id.

modules/storage/storage.go
```diff
@@ -167,22 +172,17 @@ func (s *storageModule) initialize(policy modules.StoragePolicy) error {
 
 	// make sure new pools are mounted
 	log.Info().Msgf("Making sure new volumes are mounted")
-	for _, pool := range newPools {
-		if _, mounted := pool.Mounted(); !mounted {
-			log.Debug().Msgf("Mounting volume %s", pool.Name())
-			if _, err = pool.Mount(); err != nil {
+	for idx := range newPools {
```
muhamadazmy, line 173: why using the `idx` here and not `for _, pool := range newPools` ? is this to avoid copying the string ?

modules/storage/filesystem/device.go
```diff
@@ -62,45 +62,49 @@ func (d *Device) Used() bool {
...
 // DefaultDeviceManager returns a default device manager implementation
-func DefaultDeviceManager() DeviceManager {
-	return &lsblkDeviceManager{
-		devices: []Device{},
+func DefaultDeviceManager(ctx context.Context) (DeviceManager, error) {
+	m := &lsblkDeviceManager{
+		devices: DeviceCache{},
 	}
+
+	err := m.scan(ctx)
```
muhamadazmy, line 74: While I like the idea of a `one time` scan of the devices on module load. This can cause an issue when if the machine disks has been replaced (hot plugged) while the machine is running. I have no idea if this is a case that we have to care about at this point.

zaibon: We will need to supports this at some point. This was planned in the original interface of the storage daemon.
I don't know yet exactly how we're going to implement this. But yes, we want to be able to hot swap disk in a node.
But we can probably planned that for later release

zaibon (approved):
After the comments from @muhamadazmy have been addressed I think we're good to merge.

modules/storage/filesystem/device.go
LeeSmet: I would assume that hot swapping disks generates some kernel message, so we can listen for those and act accordingly in the future

modules/storage/storage.go
LeeSmet: Not really _too_ important here since its just a string, but yes. Could have left it here, but I already changed it and it IS a very small performance benefit

muhamadazmy (approved):

PR #108, opened 2019-07-11
Tnodb provision

zaibon:
+ update http provision source to match tnodb change

### DIFF
--- a/tools/tnodb_mock/provision.go
+++ b/tools/tnodb_mock/provision.go
@@ -1,0 +1,105 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"log"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/gorilla/mux"
+
+	"github.com/threefoldtech/zosv2/modules/provision"
+)
+
+type reservation struct {
+	Reservation *provision.Reservation
+	Sent        bool
+}
+
+var reservationStore map[string][]*reservation
+var reservationMu sync.Mutex
+
+func reserve(w http.ResponseWriter, r *http.Request) {
+	nodeID := mux.Vars(r)["node_id"]
+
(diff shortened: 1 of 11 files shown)


tools/tnodb_mock/provision.go
```diff
@@ -0,0 +1,105 @@
...
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+	defer cancel()
+	ch := make(chan []*provision.Reservation)
+
+	// start long polling
+	go func(ctx context.Context, ch chan<- []*provision.Reservation) {
+		for {
```
muhamadazmy: `defer close(ch)`

tools/tnodb_mock/provision.go
```diff
@@ -0,0 +1,105 @@
...
+		http.Error(w, fmt.Sprintf("node %s not found", nodeID), http.StatusNotModified)
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+	defer cancel()
+	ch := make(chan []*provision.Reservation)
+
+	// start long polling
+	go func(ctx context.Context, ch chan<- []*provision.Reservation) {
```
muhamadazmy: I don't get why this has to be in a go routine, it can be just a busy loop with intermittent sleeps

muhamadazmy (commented):
I also don't get where the methods in tnodb/network.go are used?!

tools/tnodb_mock/provision.go
zaibon: Indeed, each request is already executing in its own goroutine. thx

zaibon:
> I also don't get where the methods in tnodb/network.go are used?!

They were all in the main.go before I just spitted the handlers into separate files

muhamadazmy (approved):

PR #111, opened 2019-07-15
let's reserve some more prefixes free

delandtj:
for administrative ends, (mgmt, routing , ....)
So we're not using `0000-000f`

### DIFF
--- a/tools/tnodb_mock/tnodb.go
+++ b/tools/tnodb_mock/tnodb.go
@@ -54,7 +54,7 @@ func allocate(allocation *Allocation) (*net.IPNet, error) {
 
 	// random from 1 to subnetCount-1
 	// we never hand out the network 0 cause we keep it for the routing segment of the farm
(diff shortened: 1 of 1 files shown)


tools/tnodb_mock/tnodb.go
```diff
@@ -54,7 +54,7 @@ func allocate(allocation *Allocation) (*net.IPNet, error) {
 
 	// random from 1 to subnetCount-1
 	// we never hand out the network 0 cause we keep it for the routing segment of the farm
-	nth := rand.Int63n(int64(subnetCount)-1) + 1
+	nth := rand.Int63n(int64(subnetCount)-16) + 16
```
zaibon, line 58: Please update the comment line 57 too to explain why we skip some numbers

zaibon (changes requested):

zaibon (approved):

PR #115, opened 2019-07-17
Network update

### LINKED ISSUE #113: network: Logic to request a new network resource for a node joining a network
When a node needs to join a network for the first time, a new network resource (NR) needs to be added to the tenant network object (TNo). 

To create a new NR, the TNoDB requires 2 information from the node:
- a free port
- the public wireguard key of the node for this TNo

We need to implement this communication between the node and the TNoDB. Currently the networker interface only expose a method to publish the public wireguard key https://github.com/threefoldtech/zosv2/blob/82028f15dd91604c6fd443dda2d7332b632e99f4/modules/network.go#L12-L18

This interface needs to be rethink to change `PublishWGPubKey` to a method that ask the TNoDB to add a new NR to the TNo.

zaibon:
fixes #113


cmds/networkd/main.go
```diff
@@ -30,43 +31,62 @@ func main() {
...
-	if err != nil {
-		return
+	if err := bootstrap(); err != nil {
+		log.Error().Err(err).Msg("failed to bootstrap network")
+		os.Exit(1)
 	}
 
-	time.Sleep(5 * time.Second)
+	log.Info().Msg("network bootstraped successfully")
+	var (
```
muhamadazmy, line 58: Just an idea 
if we exposed identityd over zbus. then other modules can just request the local id from the daemon with a zbus call. The nice side effect u get for free, is that the call will block until identityd is actually up and running and processing requests. so no timed waiting.

muhamadazmy (approved):

cmds/networkd/main.go
zaibon: That's my plan. When I've time I want to make identity module a daemon too. So it will be the only code responsible for the node identity etc

Issue #77, opened 2019-06-24
Design provisioning schemas


zaibon:
To do:
- [x] : go over current tnodb_mock data structure and clean it up: https://github.com/threefoldtech/zosv2/pull/139
- [x] : convert all the object to JSX schemas: https://gist.github.com/zaibon/8f8124c094073225a3475c4736eeb531
- [x] : validating that all the schema are actually valid and working in JSX

muhamadazmy:
I updated the schema in the issues description.

zaibon:
Now that we have our tnodb_mock and the network module is pretty much done.
I'll go over how the tnodb_mock looks like today, see if I cannot simplify a bit the data structure. Then once this clean up is done, I can produce the corresponding JSX schema that needs to be used for BCDC

Updated the first comment with some bullet points.

zaibon:
I created a gist with all the types that we use in the 0-OS that will need to be stored inside bcdb and their JSX schema counterpart: https://gist.github.com/zaibon/8f8124c094073225a3475c4736eeb531

Next step is validating that all the schema are actually valid and working in JSX ( I just wrote them from the top of my head)

zaibon:
everything can be found at https://github.com/threefoldtech/jumpscaleX_threebot/tree/development/ThreeBotPackages/threefold/tfgrid_workloads/models

PR #139, opened 2019-07-24
Clean up tnodb data structure

zaibon:
part of #77


delandtj (approved):
A part from the fact that so we cannot possibly find out if a network resource has local peers in a hidden network, yeah, I suppose that's correct ;-)

zaibon:
@delandtj there is also the problem of wireguard listening port that is still not solved here.

PR #140, opened 2019-07-25
Add versions to all the daemons

### LINKED ISSUE #133: update Makefile to include commit hash and version in daemon binaries
This is to allow us to easily debug daemon bugs

muhamadazmy:
Fixes #133

### DIFF
--- a/cmds/contd/go.sum
+++ b/cmds/contd/go.sum
@@ -69,6 +69,7 @@ github.com/google/shlex v0.0.0-20181106134648-c34317bd91bf/go.mod h1:RpwtwJQFrIE
 github.com/google/uuid v1.1.1 h1:Gkbcsh/GbpXz7lPftLA3P6TYMwjCLYm83jiFQZF/3gY=
 github.com/google/uuid v1.1.1/go.mod h1:TIyPZe4MgqvfeYDBFedMoGGpEw/LqOeaOT+nhxU+yHo=
 github.com/hpcloud/tail v1.0.0/go.mod h1:ab1qPbhIpdTxEkNHXyeSf5vhxWSCs/tWer42PpOxQnU=
+github.com/iancoleman/strcase v0.0.0-20190422225806-e506e3ef7365/go.mod h1:SK73tn/9oHe+/Y0h39VT4UCxmurVJkR5NA7kMEAOgSE=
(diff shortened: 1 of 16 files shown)


delandtj (approved):
Cool!

modules/version/version.go
```diff
@@ -0,0 +1,62 @@
...
+	Branch = "{branch}"
+	// Revision of the code
+	Revision = "{revision}"
+	// Dirty flag shows if the binary is built from a
+	// repo with uncommitted changes
+	Dirty = "{dirty}"
+)
+
+// Version interface
+type Version interface {
```
zaibon, line 24: No sure we need an interface for this. What does it bring here ?

cmds/contd/go.sum
```diff
@@ -69,6 +69,7 @@ github.com/google/shlex v0.0.0-20181106134648-c34317bd91bf/go.mod h1:RpwtwJQFrIE
 github.com/google/uuid v1.1.1 h1:Gkbcsh/GbpXz7lPftLA3P6TYMwjCLYm83jiFQZF/3gY=
 github.com/google/uuid v1.1.1/go.mod h1:TIyPZe4MgqvfeYDBFedMoGGpEw/LqOeaOT+nhxU+yHo=
 github.com/hpcloud/tail v1.0.0/go.mod h1:ab1qPbhIpdTxEkNHXyeSf5vhxWSCs/tWer42PpOxQnU=
+github.com/iancoleman/strcase v0.0.0-20190422225806-e506e3ef7365/go.mod h1:SK73tn/9oHe+/Y0h39VT4UCxmurVJkR5NA7kMEAOgSE=
```
zaibon, line 72: Why is this added ? I don't see any new dependency being added in this PR.

zaibon (approved):
LGTM, just 2 questions, but they are not blocking

modules/version/version.go
muhamadazmy: Nothing actually, may be in case we have another way to implement the version in the future may be

cmds/contd/go.sum
muhamadazmy: Hmm, weird it's from the schema actually, I don't know how it got here!

zaibon: Ok no worries, we can merge the schema PR first, rebase this branch, so these things will disappear

muhamadazmy: This doesn't go with rebase. I think we can ignore it since it's not a change to the `go.toml` so this is basically added by `go build`

PR #142, opened 2019-07-25
Schema extended

### LINKED ISSUE #141: schema: implement extra data types from jumpscale
We need to implement all needed data types from jumpscale schema:

- IP Address 
- IP Range
- Enums
- Dicts
- ...

muhamadazmy:
Add support for more data types in schema. This PR includes implementations for:

- IP Address (`ipaddr`)
- IP Range (`iprange`)
- Dict (`dict`)
- Enum (`enum`)

Fixes #141

### DIFF
--- a/modules/schema/generate_test.go
+++ b/modules/schema/generate_test.go
@@ -89,3 +89,58 @@ creation = (D)
 	// 	return object, nil
 	// }
 }
+
+func ExampleGenerateGolang_enums() {
+	const input = `
+@url =  person
+name = "UNKNOWN" (S)    #official name of the package, there can be no overlap (can be dot notation)
+gender = "male,female,others" (E)
+	`
+
+	schema, err := New(strings.NewReader(input))
+	if err != nil {
+		panic(err)
+	}
+
+	if err := GenerateGolang(os.Stdout, "test", schema); err != nil {
+		panic(err)
+	}
+
+	// Output:
+	// package test
+	//
+	// import "encoding/json"
+	//
+	// type Person struct {
+	// 	Name   string           `json:"name"`
+	// 	Gender PersonGenderEnum `json:"gender"`
+	// }
+	//
+	// func NewPerson() (Person, error) {
+	// 	const value = "{\"name\": \"UNKNOWN\"}"
+	// 	var object Person
+	// 	if err := json.Unmarshal([]byte(value), &object); err != nil {
+	// 		return object, err
+	// 	}
+	// 	return object, nil
+	// }
+	//
+	// type PersonGenderEnum uint8
+	//
+	// const (
+	// 	PersonGenderMale PersonGenderEnum = iota
+	// 	PersonGenderFemale
+	// 	PersonGenderOthers
+	// )
+	//
+	// func (e PersonGenderEnum) String() string {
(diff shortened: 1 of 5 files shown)


modules/schema/generate_test.go
```diff
@@ -89,3 +89,157 @@ creation = (D)
...
+	}
+
+	// Output:
+	// package test
+	//
+	// import "encoding/json"
+	//
+	// type Parent struct {
+	// 	Name string                 `json:"name"`
+	// 	Data map[string]Child       `json:"data"`
```
zaibon, line 219: I'm not sure how JSX handles this, but does dict only allows string as key ?

zaibon (approved):

modules/schema/generate_test.go
muhamadazmy: In Python any hashable object can be used as a key. In Go anything can be used as key i guess maybe except empty struct. I was asking regarding the implementation of schema in jsx. I think they only allow string l. Because the dict schema definition is onlyd (dict) not followed by element type. Supporting an element type (like what list do) i added to the go implementation.

zaibon: > In Go anything can be used as key 

I double checked that and this is actual not completely correct, see: 

> The comparison operators == and != must be fully defined for operands of the key type;

from https://golang.org/ref/spec#Map_types

I also did the check in JSX itself:
```python
SCHEMA = """ 
@url = test.schema.1 
info = (dict) 
init_dict = {"number": 468} (dict) 
""" 
schema = j.data.schema.get_from_text(SCHEMA)                                                                                                                                                                                              
obj = schema.new()                                                                                                                                                                                                                        
obj.info = {1:'hello',"world":2}                                                                                                                                                                                                          
{1: 'hello', 'world': 2}
obj._json                                                                                                                                                                                                                                 
'{\n"info": {\n"1": "hello",\n"world": 2\n},\n"init_dict": {\n"number": 468\n}\n}'
```
So it seems JSX schema use a plain python dict so any hashable value can be used as a key. Even wrost for us, this mean we can have mix type in the keys.

I think we should clearly define our limitation in the go implementation and warn the user when such data is received. So it will be the responsibility to the user to make sure to not go too crazy with these things when implementing something that needs to be use by our go implementation.

muhamadazmy: > I double checked that and this is actual not completely correct.
Yup, you are right, i completely forgot about the `==` and `!=` implementation.

I totally understand that we can't support everything that python do. I suggest an extension to the schema language as follows (I can implement this easily in Go):

For dict type, the Key, and Value types can be defined:
- `dict` : default to `Key == String`, and `Value == interface{}`
- `dictT` default to `Key == String` and defines `Value == T`
- `dictKV` defines `Key == K` and `Value == V`

I think this can be easily implemented in `python` as for Go. I also suggest that python add a short name for dict (as `M` since `D` is already taken by `Date`)

zaibon (approved):

modules/schema/generate_test.go
zaibon: Looks good. Let's already implement the key==string one, that's the only we use in our schema at the moment.

PR #143, opened 2019-07-25
provisiond: implement cryptography primitives for resource identification

### LINKED ISSUE #83: Resource IDs and ownership of resources
Reservation can (and will) happen on multiple steps. For example to start a container you probably have to do the following:

- Create a storage volume
- Allocate network resource
- Create container, and assign both the volume, and the network resource to that container. 

The problem is, on creating a container you need to pass the the information of the volume (may be the volume full path) and the network resource identity as well. But there is no way to grantee that the caller of the container creation is the rightful owner of the associated resources (volume and/or network)

We think that the IDs of the allocated resources should carry some information about the owner of the resource. so in that case, if the id of the volume and the network matches the owner of the container, the container creation should pass.

### Proposal 
an id of a resource can be a jwt token with payload that contains both the resource information needed, plus the tenant id. The jwt token is signed with the node private key so the jwt is only valid on the node where the resource is allocated. 

Example, you first allocate a volume, the volume id can hold the 

```json
{
  "path": "/pool/volume",
  "tenant": "id of the owner"
}
```
then on container creation, the jwt token is passed as the volume id. the node can then verify that the `jwt` has valid signature, then matches the tenant id to the tenant id of the container.create caller. Once verified, the volume path can be mounted inside the container.

The only drawback for using this technique is the size of the `jwt` token.

zaibon:
fixes #83


modules/provision/container.go
```diff
@@ -93,10 +92,15 @@ func ContainerProvision(ctx context.Context, reservation Reservation) (interface
...
+	if err != nil {
+		return nil, err
+	}
+
 	id, err := containerClient.Run(
-		reservation.Tenant.String(),
+		containerID,
 		modules.Container{
-			Name:   uuid.New().String(),
+			Name:   reservation.User,
```
muhamadazmy: This can cause conflict if the same user is hosting multiple containers on the same node. I suggest using the containerID.

modules/provision/crypto.go
```diff
@@ -0,0 +1,101 @@
...
+	signature, err := crypto.Sign(privateKey, buf.Bytes())
+	if err != nil {
+		return err
+	}
+	r.Signature = signature
+	return nil
+}
+
+// Verify verifies the signature of the reservation
+func Verify(r Reservation) error {
```
muhamadazmy, line 46: Don't see a reason why this can't be an instance method.

```suggestion
func (r *Reservation) Verify() error {
```

modules/provision/crypto.go
```diff
@@ -0,0 +1,101 @@
...
+
+	publicKey, err := crypto.KeyFromID(identity.StrIdentifier(r.User))
+	if err != nil {
+		return errors.Wrap(err, "failed to extract public key from user ID")
+	}
+
+	return crypto.Verify(publicKey, buf.Bytes(), r.Signature)
+}
+
+func Hash(r Reservation) ([]byte, error) {
```
muhamadazmy: Same here.

```suggestion
func (r *Reservation) Hash() ([]byte, error) {
```

modules/provision/provision.go
```diff
@@ -27,26 +27,19 @@ const (
...
 type Reservation struct {
 	// ID of the reservation
 	ID string `json:"id"`
-	// Tenant ID
-	Tenant Tenant `json:"tenant"`
-	// ReplyTo is a dummy attribute to hold the 3bot address
-	// we need to report to once the reservation is done
-	ReplyTo ReplyTo `json:"reply-to"`
+	// Identification of the user requesting the reservation
+	User string `json:"user_id"`
```
muhamadazmy, line 35: Why the `reply-to` was dropped? we still need to notify the user `3bot` on the reservation result. Do u have another idea to do this?

muhamadazmy (commented):
Except for few comments, the PR looks good. But there is one thing i don't get.

The signature now verify that a reservation is authentic and hasn't been tempered with. But it still doesn't verify that a resource ID pointed to by reservation is actually owned by the user. For example a VolumeID in a container reservation. 

Also, the volumeID now does not carry any information about the volume, hence the container provision logic is completely wrong now, assuming that the volumeID is the source of a mount. 

So to sum up and rephrase:
- The idea of using a recomputable hash as an ID is good.
- The node can't use the ID to point to a reservation, hence it can't figure out what u pointing to (volumeID to actual volume path) unless the node start tracking all active reservation locally in some sort of dict.
- The node can't verify that you are the owner of a resource that you are referring to (again the container -> volume example) unless again the node can track (or at least be able to query) all active reservations.
- Current code assumes that volume id is the same as the volume path which is wrong.

modules/provision/container.go
zaibon: yep this is mistake, I inverted container tenant namespace and container name

modules/provision/crypto.go
zaibon: was just to not overload the API of the Reservation struct. could be change to a method

modules/provision/provision.go
zaibon: This is not an information that needs to be in the reservation. We already have the userID (3bot) from which you get the current address of this 3bot. This address might change so no point having it in the reservation

zaibon:
> The signature now verify that a reservation is authentic and hasn't been tempered with. But it still doesn't verify that a resource ID pointed to by reservation is actually owned by the user. For example a VolumeID in a container reservation.

This is correct this is not in the PR yet. keep in mind this is still a draft and more things needs to be added.

> The node can't use the ID to point to a reservation, hence it can't figure out what u pointing to (volumeID to actual volume path) unless the node start tracking all active reservation locally in some sort of dict.

This is exactly the idea, see https://github.com/threefoldtech/zosv2/issues/83#issuecomment-515235594 for more detail

zaibon:
Ok I took time to read you again @muhamadazmy and all your point are valid.
The way I wanted to generated new workload ID based on the reservation won't work, cause I then create an indirection between the reservation ID and the workload ID that I cannot resolve without storing the mapping somewhere (we don't want state in the node)

So really the only solution I have is to directly use the reservation ID as workload ID, so the mapping is direct and doesn't requires to store any stage. 
I just need to double check how are the reservation ID will be generated (I think BCDB will be the done doing that)

zaibon:
In my last batch of commits I updated the tnodb_mock so he's the one generating reservation ID. This is going to map better with own bcdb works. The generated id now looks like `r-$id` where `$id` is an incremental int. We can change it later on. 

But one thing to keep in mind is this id needs to be relatively short, this is because since we use the reservation ID as container name, the ID cannot be too long (not sure how much) cause containerd create a unixsocket with the tenant namespace+container name
the tenant namespace being the user ID, it is already 44 chars long
Most probably the same kind of limitation applies for the volumes name in btrfs, which is now also the reservation ID.

The rest of the commits are some clean up and update of tfuser to map the new flow. 

This should be already big enough like this. I'll open this PR for review, let's see if we can merge, then let's play a bit with that built to gather info on what to do next.

muhamadazmy:
> The generated id now looks like r-$id where $id is an incremental int. We can change it later on.

Incremental IDs in a distributed ecosystem, this gonna be really hard or almost impossible to implement

zaibon:
> Incremental IDs in a distributed ecosystem, this gonna be really hard or almost impossible to implement

As far as I know this is how BCDB works. But like I said, this is implement like this to allow us to make progress now. We'll have to adapt once BCDB is live.

cmds/upgraded/main.go
```diff
@@ -4,6 +4,8 @@ import (
 	"os"
 	"time"
 
+	"github.com/threefoldtech/zosv2/modules/zinit"
```
muhamadazmy: I don't believe this entire file should be part of this Pull Request.

muhamadazmy (approved):

cmds/upgraded/main.go
zaibon: yep indeed, I moved it to https://github.com/threefoldtech/zosv2/pull/155

PR #146, opened 2019-07-28
Fix size conversion

### LINKED ISSUE #118: Volume provisioning give btrfs subvolume of size 0
I did a provisioning of volume of 5GiB, then when I wanted to write some file into my volume I immediately got an `quota exceeded error`

muhamadazmy:
Fixes #118


zaibon (approved):
Nice catch

PR #148, opened 2019-07-29
Development identity daemonize

muhamadazmy:


cmds/identityd/main.go
```diff
@@ -1,24 +1,33 @@
...
 	"github.com/cenkalti/backoff"
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
 	"github.com/threefoldtech/zosv2/modules/identity"
 	"github.com/threefoldtech/zosv2/modules/version"
 )
 
-const seedPath = "/var/cache/seed.txt"
+const (
+	seedPath = "/var/cache/seed.txt"
```
zaibon: Let's create a dedicated directory on the cache for identityd and store all its files in there instead

cmds/identityd/main.go
```diff
@@ -29,63 +38,48 @@ func main() {
...
-		}
-		return nil
+		log.Fatal().Err(err).Msg("failed to create identity manager")
 	}
 
-	err = backoff.Retry(f, backoff.NewExponentialBackOff())
+	farmID, err := manager.FarmID()
 	if err != nil {
-		return
+		log.Fatal().Err(err).Msg("fail to read farmer id from kernel parameters")
```
zaibon, line 60: Question, should we allow a node to be used when it doesn't have a farm ID ?

cmds/identityd/main.go
```diff
@@ -29,63 +38,48 @@ func main() {
...
 	}
 
-	log.Info().Msg("node registered successfully")
-}
+	nodeID := manager.NodeID()
+	log.Info().
+		Str("identify", nodeID.Identity()).
```
zaibon:
```suggestion
		Str("identity", nodeID.Identity()).
```

modules/identity.go
```diff
@@ -0,0 +1,37 @@
+package modules
+
+//go:generate mkdir -p stubs
+//go:generate zbusc -module identity -version 0.0.1 -name manager -package stubs github.com/threefoldtech/zosv2/modules+IdentityManager stubs/identity_stub.go
+
+// Identifier is the interface that defines
+// how an object can be used an identity
```
zaibon:
```suggestion
// how an object can be used as an identity
```

modules/identity.go
```diff
@@ -0,0 +1,37 @@
...
+
+// Identity implements the Identifier interface
+func (s StrIdentifier) Identity() string {
+	return string(s)
+}
+
+// IdentityManager interface.
+type IdentityManager interface {
+	// NodeID returns the node id (public key)
+	NodeID() StrIdentifier
```
zaibon, line 24:
```suggestion
	NodeID() Identifier
```
would be better, but does zbus supports interface as return value ?

modules/identity.go
```diff
@@ -0,0 +1,37 @@
...
+
+	// FarmID return the farm id this node is part of. this is usually a configuration
+	// that the node is booted with. An error is returned if the farmer id is not configured
+	FarmID() (StrIdentifier, error)
```
zaibon, line 28:
```suggestion
	FarmID() (Identifier, error)
```
same remark

modules/identity.go
```diff
@@ -0,0 +1,37 @@
...
+
+	// Sign
```
zaibon:
```suggestion
	// // Sign signs message with the private key of the node and returns a signature.
```

modules/identity.go
```diff
@@ -0,0 +1,37 @@
...
+	Sign(data []byte) ([]byte, error)
```
zaibon:
```suggestion
	Sign(message []byte) ([]byte, error)
```

modules/identity.go
```diff
@@ -0,0 +1,37 @@
...
+
+	// Verify
```
zaibon:
```suggestion
	// // Verify reports whether sig is a valid signature of message by the public key of the node
```

modules/identity.go
```diff
@@ -0,0 +1,37 @@
...
+	Verify(data, sig []byte) error
```
zaibon:
```suggestion
	Verify(message, sig []byte) error
```

modules/identity/identityd.go
```diff
@@ -0,0 +1,84 @@
...
+		pair, err = FromSeed(seed)
+		if err != nil {
+			return nil, errors.Wrap(err, "invlaid seed file")
+		}
+	}
+
+	return &identityManager{pair}, nil
+}
+
+// LocalNodeID loads the seed use to identify the node itself
```
zaibon:
```suggestion
// NodeID returns the node identity
```

modules/identity/identityd.go
```diff
@@ -0,0 +1,84 @@
...
+func (d *identityManager) NodeID() modules.StrIdentifier {
+	return modules.StrIdentifier(d.key.Identity())
+}
+
```
zaibon, line 54:
```suggestion
// FarmID returns the farm ID of the node or an error if no farm ID is configured
```

modules/identity/identityd.go
```diff
@@ -0,0 +1,84 @@
...
+	if !found {
+		return "", fmt.Errorf("farmer id not found in kernel parameters")
+	}
+
+	return modules.StrIdentifier(farmerID[0]), nil
+}
+
+// Sign signs the message with privateKey and returns a signature.
+func (d *identityManager) Sign(data []byte) ([]byte, error) {
+	if len(d.key.PrivateKey) != ed25519.PrivateKeySize {
```
zaibon: Instead of re-implementing the cryptographic logic here, just wrap the crypto package instead.

modules/identity/identityd.go
```diff
@@ -0,0 +1,84 @@
...
+		return nil, fmt.Errorf("private key has the wrong size")
+	}
+	return ed25519.Sign(d.key.PrivateKey, data), nil
+}
+
+// Verify reports whether sig is a valid signature of message by publicKey.
+func (d *identityManager) Verify(data, sig []byte) error {
```
zaibon: Instead of re-implementing the cryptographic logic, here just wrap the crypto package instead.

zaibon (changes requested):
Mostly LGTM, some minor remarks about documentation and crypto

cmds/identityd/main.go
muhamadazmy: It seems a node can't register if it doesn't have a farmer id set. if the node is not registered, there is no way u can provision anything on it, and u don't have any other means to access it. so it will be useless. 
So I would say no.

modules/identity.go
muhamadazmy: No, because it will not know what underlying type to use for the deserialization of the wire data. Hence u need to use a final type. Plus identity should never expose the private key, hence StrIdentifier is always the only possible type to return.

modules/identity/identityd.go
muhamadazmy: The crypto module on master does not have the Sign, Verify pair yet. I can move this to crypto and then just call it

zaibon: ha yep, I probably didn't push it yet, Let me update the crypto pacakge on master then.

cmds/identityd/main.go
zaibon: Fair enough. I created an issue https://github.com/threefoldtech/zosv2/issues/153
So we don't forget to document and enforce this everywhere and be as explicit about in our error message to avoid to much confusion from farmers

modules/identity.go
zaibon: ok got it.

modules/network/net_resource.go
```diff
@@ -383,7 +383,7 @@ func configWG(localResource *modules.NetResource, network *modules.Network, wgPe
 // localResource return the net resource of the local node from a list of net resources
 func (n *networker) localResource(resources []*modules.NetResource) *modules.NetResource {
 	for _, resource := range resources {
-		if resource.NodeID.ID == n.nodeID.Identity() {
+		if resource.NodeID.ID == n.identity.NodeID().Identity() {
```
muhamadazmy, line 386: Why u have to do this dynamically? the node identity is unlikely to change while the node is running !

zaibon: well I have to get the ID from somewhere

zaibon: Although you are right that the node ID should never change. I think now that we have the identity module we should try to rely on it as much as possible for these kind of things.

Imagine the following scenario:
One day we have to change the format of the node ID or any other case where the node ID change.
If you have in memory copy of the node ID in every module, it is imposible to make sure this new ID will be used properly everywhere. While if you always rely on identityd, then we know this is the only module that needs to reload the ID.

zaibon (approved):

PR #149, opened 2019-07-29
upgraded: allow upgraded to update itself

### LINKED ISSUE #145: upgraded is not able to upgrade itself
Since upgraded is running while an upgraded is happening (obviously) we get the `text file busy` error when trying to copy the new version of upgraded.

zaibon:
fixes #145


delandtj (approved):
let's test ;-)

PR #160, opened 2019-07-31
limit the amount of storage user can write on root fs of a container

### LINKED ISSUE #159: containerd: limit the amount of storage user can write on root
Currently, if a user creates a container without any volume mounted. He can abuse the root of is container and store pretty much unlimited amount of data in there.
This data ends up in the cache of the node itself.

We need to limit this as much as possible. I don't think we can make root fully read-only, but the usage of volume in container should be really encouraged.

zaibon:
fixes #159

### DIFF
--- a/modules/flist/flist.go
+++ b/modules/flist/flist.go
@@ -16,14 +16,17 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/pkg/errors"
 	"github.com/rs/zerolog/log"
 	"github.com/threefoldtech/zosv2/modules"
 )
 
 const (
 	defaultStorage = "zdb://hub.grid.tf:9900"
-	defaultRoot    = "/var/modules/flist"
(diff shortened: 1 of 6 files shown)


delandtj (approved):

modules/flist/flist.go
```diff
@@ -94,7 +107,7 @@ func (f *flistModule) Mount(url, storage string) (string, error) {
 	logPath := filepath.Join(f.log, rnd) + ".log"
 
 	args := []string{
-		"-backend", filepath.Join(f.root, "backend", rnd),
+		"-backend", filepath.Join(path, "backend"),
```
muhamadazmy: I think this can be simplified to this.
```suggestion
		"-backend", path,
```

zaibon: indeed

PR #167, opened 2019-08-05
container namespace

muhamadazmy:
When starting a container that should be part of a tenant network. The provision module will ask the network modules to `join` a specific module. the network module in response will create a new namespace and create a veth pair that has one end inside this namespace, while the other will be connected to the tenant bridge. 

The nic inside the container will be assigned a new IP from the NR object associated with the ndoe, and is granted to be unique.


modules/network/networker.go
```diff
@@ -164,13 +255,43 @@ func (n *networker) ApplyNetResource(network modules.Network) (string, error) {
...
 	path := filepath.Join(n.storageDir, string(network.NetID))
-	if err := ioutil.WriteFile(path, []byte(nibble.NetworkName()), 0660); err != nil {
+	file, err := os.Create(path)
+	if err != nil {
 		return "", err
 	}
+	defer file.Close()
+
+	enc := json.NewEncoder(file)
+	if err := enc.Encode(network); err != nil {
```
muhamadazmy: I know store the full network object in the network file. I thought may be we should only store the local NR for this network, but it seems i need more info to create a nibble for that network, that is associated with the full network object.

modules/network/networker.go
```diff
@@ -65,12 +70,98 @@ func validateNetwork(n *modules.Network) error {
...
+	}
+
+	br, err := bridge.Get(brName)
+	if err != nil {
+		return name, err
+	}
+
+	// TODO: the number in the name below must be changed to a deterministic
+	// value, may be driven from the IP assigned to this namespace
+	name = fmt.Sprintf("%s-%d", "container", 1) //TODO: CHANGE ME
```
muhamadazmy: The name here for the veth pair i have no idea yet what it should be, but may be part of the ip or the container name, may be thename should be passed to the join?

modules/network/join.go
```diff
@@ -0,0 +1,59 @@
...
+	"path"
+
+	"github.com/containernetworking/cni/pkg/types"
+	"github.com/containernetworking/cni/pkg/types/current"
+	"github.com/containernetworking/plugins/plugins/ipam/host-local/backend/allocator"
+	"github.com/containernetworking/plugins/plugins/ipam/host-local/backend/disk"
+	"github.com/threefoldtech/zosv2/modules"
+)
+
+// allocateIP allocaes a unique IP for the entity defines by the given id (for example container id, or a vm).
```
zaibon:
```suggestion
// allocateIP allocates a unique IP for the entity defines by the given id (for example container id, or a vm).
```

modules/network/bridge/bridge.go
```diff
@@ -36,6 +36,20 @@ func New(name string) (*netlink.Bridge, error) {
 	return newBr, nil
 }
 
+// Get a bridge by name
+func Get(name string) (*netlink.Bridge, error) {
+	link, err := netlink.LinkByName(name)
+	if err != nil {
+		return nil, err
```
zaibon:
```suggestion
		return nil, errors.Wrapf(err,"bridge %s not found", name)
```
Cause the error is just a file not found with not much information

modules/network/namespace/namespace.go
```diff
@@ -121,6 +121,9 @@ func getCurrentThreadNetNSPath() string {
 
 // Delete deletes a network namespace
 func Delete(ns ns.NetNS) error {
+	if ns == nil {
```
zaibon, line 124: :+1:

modules/network/network_test.go
```diff
@@ -115,7 +115,7 @@ func TestCreateNetwork(t *testing.T) {
 
 	storage := filepath.Join(dir, netName)
 	networker := &networker{
-		nodeID:     node1.NodeID,
+		//nodeID:     node1.NodeID,
```
zaibon: Left over ?

zaibon (changes requested):

modules/network/network_test.go
muhamadazmy: Yeah, This caused some tests to fail. and I was testing something, I will spend sometimes on getting these to work properly.

modules/provision/container.go
```diff
@@ -61,7 +61,7 @@ func ContainerProvision(ctx context.Context, reservation Reservation) (interface
 		Msg("deploying network")
 
 	networkMgr := stubs.NewNetworkerStub(GetZBus(ctx))
-	ns, err := networkMgr.Namespace(modules.NetID(config.Network.NetwokID))
+	ns, err := networkMgr.Join(reservation.ID, modules.NetID(config.Network.NetwokID))
```
zaibon: We should also return the IP allocated in the namespace to the caller, cause provisiond will have to send this information back to the BCDB at some point.

zaibon (changes requested):

zaibon:
Also need to resolve these conflicts before we can merge

modules/provision/container.go
muhamadazmy: True. Will do.

zaibon (approved):

zaibon:
Let see why travis is not happy before we merge

PR #180, opened 2019-08-09
network: create portm package

### LINKED ISSUE #176: network module port management
We have now 2 use case where we need to be able to ask the network module for a unused port in a specific network namespace.

- when wireguard interface needs to listen on a port for public nodes
- when creating 0db container in the same public namespace

To achieve this the network module should prove an interface to reserve and release port. A bit like the same concept with have for ip management,  but for port.

My idea is to implement the same kind of logic ip management uses. where we keep a list of used port on the local filesystem during the lifetime of the node. This is to allow networkd to be restarted without losing that information.
And since this information is only valid for the lifetime of a node and is not required to be kept safe between reboots, we don't introduce state on the node.

zaibon:
portm package implement a port manager that allows to keep
track of which port is being used in a network namespace

fixes #176

### DIFF
--- a/modules/network/portm/allocator.go
+++ b/modules/network/portm/allocator.go
@@ -6,28 +6,37 @@ import (
 	"github.com/threefoldtech/zosv2/modules/network/portm/backend"
 )
 
+// ErrNoFreePort is returned when trying to reserve a port but all the
+// the port of the range have been already reserved
 var ErrNoFreePort = errors.New("no free port find")
 
+// PortRange hold the beginging and end of a range of port
+// a PortAllocator can reserve
 type PortRange struct {
 	Start int
 	End   int
 }
 
-type allocator struct {
+// Allocator implements the PortAllocator interface
+type Allocator struct {
 	pRange PortRange
 	store  backend.Store
 }
 
-var _ PortAllocator = (*allocator)(nil)
+var _ PortAllocator = (*Allocator)(nil)
 
-func NewAllocator(pRange PortRange, store backend.Store) PortAllocator {
-	return &allocator{
+// NewAllocator return a PortAllocator
+func NewAllocator(pRange PortRange, store backend.Store) *Allocator {
+	return &Allocator{
 		pRange: pRange,
 		store:  store,
 	}
 }
 
-func (a *allocator) Reserve(ns string) (int, error) {
+// Reserve implements PortAllocator interface
+func (a *Allocator) Reserve(ns string) (int, error) {
+	a.store.Lock()
+	defer a.store.Unlock()
 
 	allocatedPorts, err := a.store.GetByNS(ns)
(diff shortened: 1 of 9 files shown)


modules/network/portm/allocator.go
```diff
@@ -0,0 +1,77 @@
...
+
+func NewAllocator(pRange PortRange, store backend.Store) PortAllocator {
+	return &allocator{
+		pRange: pRange,
+		store:  store,
+	}
+}
+
+func (a *allocator) Reserve(ns string) (int, error) {
+
```
muhamadazmy, line 40: I assume u need to do 
```suggestion
	 a.store.Lock()
	 defer a.store.UnLock()
```
otherwise u have serious race conditions.

modules/network/portm/allocator.go
```diff
@@ -0,0 +1,77 @@
...
+		if reserved {
+			return port, nil
+		}
+	}
+
+	return 0, ErrNoFreePort
+}
+
+func (a *allocator) Release(ns string, port int) error {
+	return a.store.Release(ns, port)
```
muhamadazmy, line 80: Same, u need to lock the store before u do a release.

modules/network/portm/backend/fs.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	return strconv.Atoi(string(data))
+}
+
+func (s *fsStore) GetByNS(ns string) ([]int, error) {
+	var ports []int
+	dir := filepath.Join(s.root, ns)
+
+	// walk through all ips in this network to get the ones which belong to a specific ID
+	_ = filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
```
muhamadazmy: Why do u have to do a walk (which is recursive) instead of `ioutil.ReadDir` which will give u the list of all files in that `ns` directory.

modules/network/portm/backend/lock.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/alexflint/go-filemutex"
+)
+
+// FileLock wraps os.File to be used as a lock using flock
+type FileLock struct {
+	f *filemutex.FileMutex
+}
+
+// NewFileLock opens file/dir at path and returns unlocked FileLock object
+func NewFileLock(lockPath string) (*FileLock, error) {
```
muhamadazmy, line 16: I think it will be cleaner if this method will return `FileMutex` directly (drop the wrapper). I think the only reason u created this file is the code in the `NewFileLock` method, but u don't need the wrapper struct.

Worst case u define the wrapper as
```go
type FileLock struct {
*filemutex.FileMutex
}
```
this way, u don't need the proxy methods.

muhamadazmy (approved):

PR #182, opened 2019-08-11
provision: implement reservation expiration

### LINKED ISSUE #121: Provisioning module has no notion of reservation expiration
Currently in the provisioning module, the reservation have no notion of time. We lack the information about for how long does the reservation needs to be live.

We need to add a duration to the Reservation struct, so we can know if a reservation is still valid or if it should be deleted from the system

zaibon:
the provision modules now keeps track of the reservation currently
deployed on the node.
It then check periodically which reservation have expired and
decomission them from the system.

fixes #121


zaibon:
I open this as a draft cause there are couple of things I'm too happy about.
While this PR is fully working I would like to discuss some point:

- In this PR I now keep the full reservation object in the local_store of provisiond. I do that cause to decommission networks I need to have the full network object
- the local store has only a memory backend at the moment, we need to move that on disk
- I didn't handle cases when provisiond process restarts
- The concurrent logic for the provision engine is a bit rough, I think we can improve that part
- The value for expiration check is hard-coded and arbitrary
- Most of the new code have no tests

One thing I notice while writing this is that both `iterExpiration` and `iterReservation` function have the same signature `func ctx context.Context) <-chan Reservation` which is the same as the only method in the `ReservationSource` interface
```go
type ReservationSource interface {
	Reservations(ctx context.Context) <-chan Reservation
}
```

so actually we could maybe improve the `combinedResource` struct:
```go
type compinedSource struct {
	s1 ReservationSource
	s2 ReservationSource
}
```
to support any number of source by implementing a proper fade-in and then just make `iterExpiration` and `iterReservation` a source for the engine.

modules/provision/container.go
```diff
@@ -150,3 +150,10 @@ func ContainerProvision(ctx context.Context, reservation Reservation) (interface
 	log.Info().Msgf("container created with id: '%s'", id)
 	return id, nil
 }
+
+func containerDecomission(ctx context.Context, reservation Reservation) error {
```
muhamadazmy: `Decommission` not `Decomission`

modules/provision/container.go
```diff
@@ -150,3 +150,10 @@ func ContainerProvision(ctx context.Context, reservation Reservation) (interface
...
+	client := GetZBus(ctx)
+
+	containerClient := stubs.NewContainerModuleStub(client)
+	return containerClient.Delete(reservation.User, modules.ContainerID(reservation.ID))
```
muhamadazmy: U also need to unmount the rootfs of the container, so I guess u will need to do inspect first, store the rootfs, delete the container then unmount the filesystem.

modules/provision/engine.go
```diff
@@ -18,40 +20,141 @@ type defaultEngine struct {
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			log.Info().Msg("provision engine context done, exiting")
+			return nil
+
+		case reservation, ok := <-cReservation:
+			if !ok {
+				log.Info().Msg("reservation source is emptied. stopping engine")
```
muhamadazmy, line 53: Why do we need this. I assume a reservation source should never considered "terminated", it can just have no new stuff for sometime.

modules/provision/engine.go
```diff
@@ -18,40 +20,141 @@ type defaultEngine struct {
...
+				log.Error().Err(err).Msg("failed to remove reservation %s from expiration store")
+			}
 		}
-		e.reply(reservation.ID, result, err)
 	}
 
-	return nil
+}
+
+func (e *defaultEngine) iterReservation(ctx context.Context) <-chan Reservation {
```
muhamadazmy: I don't get why u need this method. I can see that you only copy the objects from the source channel to another channel, why not use the e.source.Reservations(ctx) channel directly ?

> If u need to do call to the `validate()` i think this can be done easier in the source itself.

modules/provision/engine.go
```diff
@@ -18,40 +20,141 @@ type defaultEngine struct {
...
+				return
+			case c <- reservation:
+			}
+		}
+	}()
+
+	return c
+}
+
+func (e *defaultEngine) iterExpiration(ctx context.Context) <-chan *Reservation {
```
muhamadazmy: I think here u can also create a `ReservationSource` to wrap this method. It will be cleaner to have this logic in a different source than directly on the engine. Because that's a lot of logic that doesn't directly relate to the engine.

modules/provision/engine.go
zaibon: If we don't handle that case and for any reason the source provider close its channel the loop on the select will be an infinite loop receiving nil reservations. Which will eventually crash the daemon.
I had the case when working on this. When provisiond boots, it first provide all the reservations for this node once, then starts waiting for new one. I now changed the logic so I don't use a different source for this, but the underlying problem remains.

modules/provision/engine.go
zaibon: yep I can move the validation in the source itself, and get rid of this function.

modules/provision/engine.go
```diff
@@ -17,41 +23,87 @@ type defaultEngine struct {
...
+
+	cReservation := e.source.Reservations(ctx)
+
+	for {
+		select {
+		case <-ctx.Done():
+			log.Info().Msg("provision engine context done, exiting")
+			return nil
+
+		case reservation, ok := <-cReservation:
```
muhamadazmy, line 51: Just some cosmetics here, this switch case is so long, can we move the logic to a separate function(s) instead of one big select case.

modules/provision/local_store.go
```diff
@@ -0,0 +1,81 @@
...
+	for _, r := range s.m {
+		if !isExpired(r) {
+			continue
+		}
+		output = append(output, r)
+	}
+	return output, nil
+}
+
+func isExpired(r *Reservation) bool {
```
muhamadazmy: Why this is not defined as a member method on the Reservation itself? 
like
```suggestion
func (r *Reservation) Expired() bool {
```

PR #196, opened 2019-08-21
0-db namespace provisioning

zaibon:


modules/provision/zdb.go
```diff
@@ -4,107 +4,318 @@ import (
...
+		}
+		return nil, errors.Wrap(err, "failed to create container")
 	}
-	for _, ns := range nss {
-		if ns == nsID {
-			// namespace already exists
-			return
+
+	var containerIP net.IP
+	for timeout := time.Now().Add(time.Minute * 2); !time.Now().After(timeout); time.Sleep(time.Second * 10) {
```
delandtj: wut ? ;-)

modules/network/networker.go
```diff
@@ -188,6 +189,72 @@ func (n *networker) Join(member string, id modules.NetID) (join modules.Member,
 	return join, bridge.AttachNic(hostVeth, br)
 }
 
+// ZDBPrepare sends a macvlan interface into the
+// network namespace of a ZDB container
+func (n networker) ZDBPrepare() (string, error) {
```
muhamadazmy, line 194: Can we have this with a more generic name. Networker should not care if this is ZDB or not, the provisiond instead (which knows the available modules) can instead use the generic functionality of both networker and storeaged to do the setup.

modules/network/networker.go
```diff
@@ -188,6 +189,72 @@ func (n *networker) Join(member string, id modules.NetID) (join modules.Member,
...
+
+	// we don't set any route or ip
+	if err := macvlan.Install(macVlan, []*net.IPNet{}, []*netlink.Route{}, netNs); err != nil {
+		return "", err
+	}
+
+	return netNSName, nil
+}
+
+func (n networker) ZDBIp(netNSName string) (net.IP, error) {
```
muhamadazmy, line 230: Same here, i don't see why this can't be a generic (not zdb oriented call)

muhamadazmy (approved):

modules/network/networker.go
zaibon: This one is a bit more tricky to make generic cause it does really specific things for how 0-db networking is done at the moment.
Maybe I could split it into 3 or 4 other methods... I'll see what I can do. If that's too much work I'll just open a issue to follow up on that so we don't block this PR

PR #199, opened 2019-08-22
Document storage paths under cache

### LINKED ISSUE #79: Document file location of each module
I would like to have clear view on all the file that each module need to be able to write/create.

Then from there we can have a reflection if this is the best layout or if we can improve things a bit.

The main idea is I want to avoid that any module can write anywhere on the filesystem. This leads to less security and make things harder do debug.

muhamadazmy:
Also fix upgraded cache path and make sure
it's set in the zinit config

Fixes #79

### DIFF
--- a/zinit/upgraded.yaml
+++ b/zinit/upgraded.yaml
@@ -1,4 +1,4 @@
-exec: upgraded
+exec: upgraded -root /var/cache/modules/upgrade
 after:
   - boot
   - flistd
--- a/cmds/contd/main.go
+++ b/cmds/contd/main.go
@@ -26,7 +26,7 @@ func main() {
 		ver           bool
 	)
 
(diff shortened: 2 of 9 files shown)


zinit/upgraded.yaml
```diff
@@ -1,4 +1,4 @@
-exec: upgraded
+exec: upgraded -root /var/cache/modules/upgrade
```
muhamadazmy, line 1: It seems upgraded was using a location outside the cache disk which means it used the tmpfs to download the files. Don't think this was intended

zaibon (approved):

zinit/upgraded.yaml
zaibon: upgraded doesn't download anything, it just uses the flist module. and does the copy from there into the rootfs. But anyhow your change are good

Issue #170, opened 2019-08-06
Upgrade and lifetime procedures for zosv2 binaries


delandtj:
We need to come up with a proper lifetime definition and accordance with the web-served image for booting the nodes.
Basically, even when live-upgrading binaries in zos, there needs to be :
  - a global state for all binaries, with one version (or a list of compatible versions)
  - the served 0-OS (uefi) image needs to follow that versioning, checking necessary migrations along the way until all migrations are done before getting to a 'ready' state.
  - also figure out __where__ we 'install' (or overwrite) binaries, with their migration scripts? tools?

- [x] Specifications
- [x] Data migration support in daemons
- [x] Bootstrap

zaibon:
Current status is:
- We have a single 0-OS image that is updated manually by triggering a new build whenever we feel like it
- The live-upgrading is done using upgraded ([see doc](https://github.com/threefoldtech/zosv2/tree/master/docs/upgrade)) and we have a development version of the "upgrade server" at https://versions.dev.grid.tf. This server list all the available version and detail of the flist to use to apply the upgrade

Now with what we have at the moment there is already multiple issues:
1. there is no synchronization between the version publish for upgraded and the 0-OS images
2. If there are some migration steps to be done, only the live-upgrade currently support it. There is nothing in place in the 0-OS images that allow to do any migration.
3. As new version are published and if we don't creates more recent images, the time to apply all update at boot will become more and more long. 
4. If we always starts from an old 0-OS image and redo the full incremental update, we will try to redo the same data migration multiple time, which will eventually create a disaster...

muhamadazmy:
## How about we do this

- 0-OS official image should has no binary, only the absolute necessary plus the `upgraded` daemon (or some sort of other bootstrap daemon), may be also `networkd` so machine can connect to the internet.
- On boot, the upgraded will always have to download latest binaries from the official releases (in sort of a manifest file) it has to always get the latest. binaries, and only store them in the tmpfs (in memory) no persistence.
- The image (and upgraded) should not care about the data migration, it's up to the  daemons themselves to be support older versions of their own data. if an older version format is found on disk, it can update the data to newer version. It means that daemons must keep track of an object schema version in the data itself.
- The upgrade server `versions.dev.grid.tf` should have a manifest of the latest binaries to download, hashes, links to binaries, etc... so `zos` on boot can download the manifest, followed by the needed binaries (all binaries), zinit config files and then validate them before starting the services.
[...]

PR #208, opened 2019-08-25
Master upgrade procedure

muhamadazmy:
This PR concerns only with the following:
- Specifications of the upgrade/fresh boot procedure
- Data migration of the daemons

Another PR will be created for `bootstraping` the node

Part Of #170


modules/identity/keys.go
```diff
@@ -37,17 +44,36 @@ func GenerateKeyPair() (k KeyPair, err error) {
...
 // KeyPair using the seed
-func LoadSeed(path string) (k KeyPair, err error) {
-	seed, err := ioutil.ReadFile(path)
+func LoadKeyPair(path string) (k KeyPair, err error) {
+	seed, err := LoadSeed(path)
 	if err != nil {
 		return k, err
 	}
-
+	// version update can go here.
```
zaibon: How do we know the version of the seed we just loaded at this point ? I don't see how we can decide to do an ugprade or not here.

zaibon (commented):
LGTM, just had one question

modules/identity/keys.go
muhamadazmy: Ah, I moved the code but didn't move the comment :/

modules/versioned/data_test.go
```diff
@@ -0,0 +1,182 @@
...
+	dec := json.NewDecoder(reader)
+
+	if reader.Version().Compare(latest) <= 0 {
+		//data version is older than or equal latest
+		var data string
+		if err := dec.Decode(&data); err != nil {
+			panic(err)
+		}
+		fmt.Println("data is:", "my data goes here")
+	}
```
zaibon, line 178: can we add an example how we would migrate the data if the version is not the one expected ?

muhamadazmy: sure

zaibon (approved):

maxux:
This breaks `tools/tfuser` build.

PR #214, opened 2019-08-27
Capacity module

zaibon:


cmds/capacityd/main.go
```diff
@@ -0,0 +1,76 @@
...
+
+	log.Info().Msg("sends capacity detail to BCDB")
+	if err := store.Register(identity.NodeID(), resources, dmi); err != nil {
+		log.Fatal().Err(err).Msgf("failed to write resources capacity on BCDB")
+	}
+
+	for {
+		<-time.After(time.Minute * 10)
+
+		log.Info().Msg("send heat-beat to BCDB")
```
muhamadazmy:
```suggestion
		log.Info().Msg("send heart-beat to BCDB")
```

modules/capacity/dmi/dmi.go
```diff
@@ -0,0 +1,341 @@
+package dmi
```
muhamadazmy, line 1: You should have also moved the dmi_test.go file here to make sure we have this covered for potential bugs.

modules/storage/filesystem/btrf_test.go
```diff
@@ -133,145 +134,105 @@ func TestMain(m *testing.M) {
 func basePoolTest(t *testing.T, pool Pool) {
 	t.Run("test mounted", func(t *testing.T) {
 		_, mounted := pool.Mounted()
-		if ok := assert.False(t, mounted); !ok {
-			t.Error()
-		}
+		assert.False(t, mounted)
```
muhamadazmy, line 137: is `assert ` alone will set the test to `error`, or u need to change this to `require`. I think `require ` equals calling `Fatal()`

muhamadazmy (commented):
So the farmer will be rewarded for the provided capacity only? not taking into account the actual used capacity ?

Actually this makes sense, since cpu and memory is already paid for (one time purchase) but I am not sure about networking

zaibon:
> So the farmer will be rewarded for the provided capacity only? not taking into account the actual used capacity ?

He is paid when people reserve capacity. But this can be easily computed from BCDB itself. I don't think there is need for 0-OS to report that.

modules/storage/filesystem/btrf_test.go
zaibon: assert make the test fails is the predicate is false. So this is fine since this is the last line of the sub-test and that what we want to verify.
I usually use requires only to early exit tests when condition that are needed for the test to be able to run are not met

PR #216, opened 2019-08-27
provisiond: implement debug mode

### LINKED ISSUE #205: Debug mode for the node
I had an idea regarding debugging for the node once they will be in the field.

When node will be in production we won't have any way to reads the logs. My idea is to create a new type of reservation. This reservation would need to be signed by the farmer and contains an endpoint where the logs needs to be forwarded.
So when we want to debug a node we send this reservation schema to the farmer he signs it then send it to BCDB. The nodes will pick it up and start forwarding the logs to the endpoint. This could also enable debug level for the logs.

This is not really a backdoor since we don't open an access on the node itself and it is still fully up to the farmer to agree

zaibon:
fixes #205
The debug mode is implemented as a new reservation type
The mode received a debug reservation that contains a redis endpoint
Once it receives it, it will start forwarding its log to the redis
endpoint using pubsub.


modules/provision/debug.go
```diff
@@ -0,0 +1,116 @@
...
+	var cfg Debug
+	if err := json.Unmarshal(reservation.Data, &cfg); err != nil {
+		return nil, err
+	}
+
+	return startZLF(ctx, reservation.ID, cfg)
+}
+
+func debugDecommission(ctx context.Context, reservation *Reservation) error {
+	return stopZLS(ctx, reservation.ID)
```
maxux: ZLF

modules/provision/debug.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func startZLF(ctx context.Context, ID string, cfg Debug) (string, error) {
```
maxux, line 37: ZLF

modules/provision/debug.go
```diff
@@ -0,0 +1,116 @@
...
+	name := fmt.Sprintf("zlf-debug-%s", ID)
+	if err := zinit.AddService(name, s); err != nil {
+		return "", errors.Wrap(err, "fail to add init service to zinit")
+	}
+
+	if err := z.Monitor(name); err != nil {
+		return "", errors.Wrap(err, "failed to start monitoring zlf service")
+	}
+
+	if err := z.Start(name); err != nil {
```
muhamadazmy: You don't need to call start, monitor will auto start it.
you only need to call start if you previously did a `stop`

zaibon: got it

modules/provision/debug.go
```diff
@@ -0,0 +1,112 @@
...
+
+	channel := fmt.Sprintf("%s-logs", identity.NodeID().Identity())
+	if cfg.Channel != "" {
+		channel = cfg.Channel
+	}
+
+	s := zinit.InitService{
+		Exec:    fmt.Sprintf("%s --host %s --port %d --channel %s", path, cfg.Host, cfg.Port, channel),
+		Oneshot: false,
+		After:   []string{"networkd"},
```
muhamadazmy, line 60: You also need to set `log: stdout` otherwise the logs of the zlf will go to the ringbuffer, which will cause a positive feedback loop

muhamadazmy: I can see that `zlf` only print max 2 lines of log messages on start, so it will not cause a positive feedback loop. this can happen though if zlf start logging on each 'forward' operation.

zaibon: better safe then sorry

muhamadazmy (approved):

modules/zinit/service.go
```diff
@@ -3,16 +3,40 @@ package zinit
...
+	if err := unmarshal(&buf); err != nil {
+		return err
+	}
+	*s = LogType(strings.ToLower(buf))
+	return nil
+}
+
+// All the type of logging supported by zinit
+const (
+	NullLogType   LogType = "null"
```
muhamadazmy: We don't have the null type anymore after your request to drop it :D

zaibon: ha ok, Then we need to udpate the documentation at: https://github.com/threefoldtech/zinit/tree/master/docs cause the null mode is still present there

PR #219, opened 2019-08-28
Add autobuild script

muhamadazmy:


zaibon (commented):
Can we make this flow depend on the success of the build flow ?
So we don't have to redo the all test/building phase ?

.github/workflows/flist.yml
```diff
@@ -0,0 +1,56 @@
...
+      env:
+        CODECOV_TOKEN: 83d31816-f8ad-4d22-9bf5-21f42e3c58f7
+        
+    - name: Build binaries
+      run: |
+        cd cmds
+        make
+      env:
+        GO111MODULE: on
+    - name: Build FList
```
zaibon: add an extra space here for readability.

zaibon: shouldn't you have a `if: success()` here ?

zaibon (changes requested):

zaibon (dismissed):
This LGTM.
But we also need a way to notify upgraded of the current version before it starts otherwise it will try to re-apply some update.
Check at https://github.com/threefoldtech/zosv2/blob/master/modules/upgrade/upgrade.go#L65 to see how upgraded check the current version

zaibon:
And the documentation around live upgrade also needs to be updated to reflect the new way of working.

muhamadazmy:
> But we also need a way to notify upgraded of the current version before it starts otherwise it will try to re-apply some update.

This PR is not related to `upgraded` or the update life cycle. This only configure the actions to auto-build the flist with all the binaries.

It also fixes a small issue in networkd that makes the network not work if the interfaces already have IP. (if the internet service from the 0-initramfs is working)

zaibon:
Then only the build a scripts need to be part of this PR. And we need to wait to get have the rest of the life cycle changed before we change the initramfs and introduce the internet service.

muhamadazmy:
> Then only the build a scripts need to be part of this PR. And we need to wait to get have the rest of the life cycle changed before we change the initramfs and introduce the internet service.

I believe I am not following on what u mean. Merging this PR here will not break the current build or change the image by anyhow (there is another PR on 0-initramfs) that **SHOULD NOT** be merged until we have the full cycle in place.

The next action we need to add here is to update version server with the correct flist name to use. then we can work on the upgraded daemon to use this new flist instead.

zaibon:
Yeah I went over the changes again an indeed. Nothing beaks here. I thought the addresses deletion happened after the bootstrap. That's not the case.
Does zinit stop on a non existing service doesn't return an error though?

zaibon (approved):

PR #220, opened 2019-08-28
Provisiond: send status back to BCDB after deployement of a workload

### LINKED ISSUE #218: provisiond: send status back to BCDB after deployement of a workload
At the moment, there is no way for a user to know what was the status of a reservation is or to receive some connection information for a deployed workload.

The node needs to write these detail back to the BCDB in all cases. All being or it succeeded to deploy the workload or it didn't. In case it worked it needs to sends some connection information.

Response information for each workload:
- volume: volume ID
- container: container ID and IP address in the private network
- 0-db namespace: namespace ID, IP address and port 
- network: nothing, everything is know before deployment

Next to these connection information, a reservation needs to be marked as deployed or failed. If it failed an error message explaining the reason of the failure is attached to the response.

zaibon:
fixes #218

Send response for :
- [x] container
- [x] volume
- [x] 0-db
- [x] network (nothing to send back)

Have deterministic IP for :
- [x] container
- [ ] 0-db


zaibon:
@muhamadazmy Still have to do some clean up, but can you already have a look at this draft please.

modules/provision/engine.go
```diff
@@ -15,9 +18,14 @@ type ReservationReadWriter interface {
 	Remove(id string) error
 }
 
+type Resulter interface {
```
zaibon: This name is terrible, could use another one

muhamadazmy: Of course. change it to whatever feels appropriate.

modules/provision/engine.go
```diff
@@ -116,16 +129,39 @@ func (e *defaultEngine) decommission(ctx context.Context, r *Reservation) error
 	return nil
 }
 
-func (e *defaultEngine) reply(id string, result interface{}, err error) {
-	//TODO: actually push the reply to the endpoint defined by `to`
-	if err != nil {
+func (e *defaultEngine) reply(ctx context.Context, r *Reservation, rErr error, info interface{}) error {
+	zbus := GetZBus(ctx)
+	identity := stubs.NewIdentityManagerStub(zbus)
+	result := &Result{}
```
muhamadazmy, line 137: why u do this syntax instead of 
```suggestion
	var result Result
```

modules/provision/engine.go
muhamadazmy: What about 
```suggestion
type  ResultCallback func(id string, result *Result) error
```

modules/provision/provision.go
```diff
@@ -55,18 +52,21 @@ type Reservation struct {
...
 	// Date of creation
 	Created time.Time `json:"created"`
 	// Duration of the reservation
 	Duration time.Duration `json:"duration"`
+	// Signature is the signature to the reservation
+	// it contains all the field of this struct except the signature itself and the Result field
+	Signature []byte `json:"signature,omitempty"`
+
+	// Result is filled by the node after the reservation has been processed
+	Result *Result
```
muhamadazmy, line 64: Why result is part of reservation?

modules/provision/volume.go
```diff
@@ -49,7 +49,9 @@ func volumeProvision(ctx context.Context, reservation *Reservation) (interface{}
 		return path, nil
 	}
 
-	return storageClient.CreateFilesystem(reservation.ID, config.Size*gigabyte, modules.DeviceType(config.Type))
+	_, err = storageClient.CreateFilesystem(reservation.ID, config.Size*gigabyte, modules.DeviceType(config.Type))
+	// nothing to return to BCDB
```
muhamadazmy: Why, shouldn't this return the ID of the volume ?

modules/zinit/service.go
```diff
@@ -31,11 +31,11 @@ const (
...
-	Exec    string
-	Oneshot bool
-	Test    string
-	After   []string
-	Log     LogType
+	Exec    string   `yaml:"exec,omitempty"`
+	Oneshot bool     `yaml:"oneshot,omitempty"`
+	Test    string   `yaml:"test,omitempty"`
+	After   []string `yaml:"after,omitempty"`
+	Log     LogType  `yaml:"log,omitempty"`
```
muhamadazmy: I don't believe this relate to the PR but okay

tools/tnodb_mock/provision.go
```diff
@@ -92,11 +92,10 @@ func getRes(nodeID string, all bool) []*provision.Reservation {
 			continue
 		}
 		// if we are long polling, only return the new reservation
-		if !all && r.Sent {
+		if !all && r.Reservation.Result != nil || r.Reservation.Expired() {
```
muhamadazmy: So i see why u store the result on the reservation. But I feel this is not the optimum way to mark reservations as (processes).

Also don't you need to actually re-create all old provisions on reboot of the node anyway?

modules/provision/engine.go
zaibon: What you suggest doesn't produce the same result btw.
I'm creating a pointer to a Result object,
You are creating an Result object

regarding the syntax this is just by habit. I always user literal whenever I can.

modules/provision/engine.go
zaibon: Make this a type func would force me to pass a function to the constructor. 
I don't really like that idea.

modules/zinit/service.go
zaibon: indeed, I already pushed it on master so I can remove it from here. https://github.com/threefoldtech/zosv2/commit/a36852a93f4e9b68db1be020c1602e19bb876df8

tools/tnodb_mock/provision.go
zaibon: It doesn't hurt regarding the reboot flow, cause at boot I ask BCDB to send me all the reservation.
But you are right, a single reservation can potentially be deployed multiple time of the node reboots. So maybe we should keep a list of result per reservation. And having it inside the Reservation object is probably also wrong. We can definitely  move it out.

modules/provision/volume.go
zaibon: I didn't send it cause the ID is actually the reservation ID. But this is an implementation detail and that could change in the future. so yep I need to send the volume ID back.

PR #226, opened 2019-08-31
WIP: implement gedis client to BCDB

### LINKED ISSUE #224: implement gedis client to BCDB for identityd
The implementation of the BCDB actor has started : https://github.com/threefoldtech/tf_cockpit/issues/7

As this effort is going we need to implement the gedis client in our side so we can gradually replace all the client we use to talk to our bcdb_mock with the real one.

We have multiple location in our code base where we have implemented a part of the BCDB interface:
- https://github.com/threefoldtech/zosv2/blob/master/modules/identity/store.go
- https://github.com/threefoldtech/zosv2/blob/master/modules/network/tnodb/store.go 

Could be I missed some place where we have other code talking to BCDB, so an audit of the full code is probably required.

Part of this task should be implemented some interoperability tests between our go client and the python BCDB actors

zaibon:
Fixes #224

### DIFF
--- a/cmds/identityd/main.go
+++ b/cmds/identityd/main.go
@@ -70,7 +70,7 @@ func main() {
 		store := identity.NewHTTPIDStore(tnodbURL)
 		f := func() error {
 			log.Info().Msg("start registration of the node")
-			if err := store.RegisterNode(nodeID, farmID); err != nil {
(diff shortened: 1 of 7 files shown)


zaibon:
will be merged as part of https://github.com/threefoldtech/zosv2/pull/245

PR #236, opened 2019-09-09
Split networkd

delandtj:
Azmy, help! ;-)


modules/network/bootstrap.go
```diff
@@ -31,9 +28,9 @@ func Bootstrap() error {
 		return err
 	}
 
-	if _, err := sysctl.Sysctl(fmt.Sprintf("net.ipv6.conf.%s.disable_ipv6", DefaultBridge), "1"); err != nil {
-		return errors.Wrapf(err, "failed to disable ip6 on bridge %s", DefaultBridge)
-	}
+	// if _, err := sysctl.Sysctl(fmt.Sprintf("net.ipv6.conf.%s.disable_ipv6", DefaultBridge), "1"); err != nil {
```
muhamadazmy: When code is not needed, we delete it instead of commenting it out. otherwise it will cause confusion.

modules/network_encoding.go
```diff
@@ -339,10 +339,10 @@ func (d *DNAT) MarshalJSON() ([]byte, error) {
...
 func (e *ExitPoint) UnmarshalJSON(b []byte) error {
 	tmp := struct {
-		Ipv4Conf  *Ipv4Conf `json:"ipv4_conf"`
-		Ipv4Dnat  []*DNAT   `json:"ipv4_dnat"`
-		Ipv6Conf  *Ipv6Conf `json:"ipv6_conf"`
-		Ipv6Allow []net.IP  `json:"ipv6_allow"`
+		Ipv4Conf  *Ipv4Conf    `json:"ipv4_conf"`
+		Ipv4Dnat  []*DNAT      `json:"ipv4_dnat"`
+		Ipv6Conf  *Ipv6Conf    `json:"ipv6_conf"`
+		Ipv6Allow []*Ipv6Allow `json:"ipv6_allow"`
```
muhamadazmy, line 345: Any reason this was defined as list of pointers instead of list of struct?
```suggestion
		Ipv6Allow []Ipv6Allow `json:"ipv6_allow"`
```

modules/network_encoding.go
```diff
@@ -339,10 +339,10 @@ func (d *DNAT) MarshalJSON() ([]byte, error) {
 // UnmarshalJSON implements encoding/json.Unmarshaler
 func (e *ExitPoint) UnmarshalJSON(b []byte) error {
```
muhamadazmy, line 340: I don't get the point of the custom json marshaling, i will review this part to see if it's needed

assets/boot_sequence.plantuml
```diff
@@ -11,6 +11,7 @@ package "node-ready"{
 
 package "boot" {
     [storaged]
+    [initnet]
```
zaibon: need to re-generate the .png of this diagram

cmds/initnet/go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/threefoldtech/zosv2/cmds/initnet
+
+go 1.12
+
+require (
+	github.com/cenkalti/backoff v2.2.1+incompatible
+	github.com/rs/zerolog v1.15.0
+	github.com/threefoldtech/zosv2/modules v0.0.0-20190902164829-025b3c42efbc
```
zaibon: Since we didn't release any version of this module yet, we need to use a replace statement.

`replace github.com/threefoldtech/zosv2/modules => ../../modules/`

cmds/initnet/main.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/threefoldtech/zosv2/modules/version"
+	"github.com/threefoldtech/zosv2/modules/zinit"
+)
+
+const redisSocket = "unix:///var/run/redis.sock"
+const module = "initnet"
+
+func main() {
+	var (
+		root   string
```
zaibon: I don't see this code using any disk, so we don't need this root variable

cmds/initnet/main.go
```diff
@@ -0,0 +1,95 @@
...
+		broker string
+		ver    bool
+	)
+
+	flag.StringVar(&root, "root", "/var/cache/modules/initnet", "root path of the module")
```
zaibon: same here, root flag can be removed

cmds/initnet/main.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/threefoldtech/zosv2/modules/zinit"
+)
+
+const redisSocket = "unix:///var/run/redis.sock"
+const module = "initnet"
+
+func main() {
+	var (
+		root   string
+		broker string
```
zaibon: this module also doesn't talk to any other module so no need to redis broker address

cmds/initnet/main.go
```diff
@@ -0,0 +1,95 @@
...
+		ver    bool
+	)
+
+	flag.StringVar(&root, "root", "/var/cache/modules/initnet", "root path of the module")
+	flag.StringVar(&broker, "broker", redisSocket, "connection string to broker")
```
zaibon: broker flag can be removed

cmds/initnet/main.go
```diff
@@ -0,0 +1,95 @@
...
+	flag.BoolVar(&ver, "v", false, "show version and exit")
+	flag.Parse()
+	if ver {
+		version.ShowAndExit(false)
+	}
+
+	log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr})
+
+	if err := os.MkdirAll(root, 0750); err != nil {
```
zaibon: no need to create root, it's not used

cmds/networkd/main.go
```diff
@@ -56,22 +53,7 @@ func main() {
...
-		log.Error().Err(err).Msg("failed to bootstrap network")
-		os.Exit(1)
-	}
-
-	log.Info().Msg("network bootstraped successfully")
-
-	if err := ready(); err != nil {
-		log.Fatal().Err(err).Msg("failed to mark networkd as ready")
-	}
+	db := tnodb.NewHTTPTNoDB(tnodbURL)
```
zaibon, line 60: shouldn't we at least do some basic checks that the current state of the node networking is what we expect it to be?
Like zos bridge present with an IP address on it etc ? Just to prevent we start networkd while initnet somehow failed to do his job ?

zaibon (changes requested):

muhamadazmy (commented):
Can we also rename `initnet` to something nicer. May be `internetd` is a nicer name

zaibon:
>  May be internetd is a nicer name

actually this is not a deamon, so the `d` suffix doesn't apply here.

modules/network_encoding.go
zaibon: it was needed because it used to contain a `net.IP`, but since it is now gone, maybe we can avoid the custom marshaling for this struct.

modules/network/bootstrap.go
delandtj: hmm I was in limbo about that, as now I see we need it, but instead of setting to `1`, we need to set to `0` expicitely

cmds/networkd/main.go
delandtj: with the new boot diagram, it would be like that, no ? Azmy ?

muhamadazmy: @delandtj since this is a oneshot service. if it fail it will not be "retried" but other services depend on it (like networkd) won't start. The code is already implementing a "retry" mechanism so when it succeed it's granted (fingercrossed) that the bridge setup is correct.

zaibon: I get that, I just want to add another layer of verification here. Better safe then sorry. This will also ensure we don't forget to update networkd if we ever change the bootstrap process.

muhamadazmy: networkd does not depend on the `zos` bridge for it's operation as far as I understand. But I added a check in `interent` to make sure that after network bootstraping is complete that we actually have intenet connection before handing over network management to networkd

zaibon: It actually does for the zdb networking

muhamadazmy (approved):

cmds/internet/main.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/rs/zerolog/log"
+
+	"github.com/threefoldtech/zosv2/modules/network"
+	"github.com/threefoldtech/zosv2/modules/network/ifaceutil"
+
+	"github.com/threefoldtech/zosv2/modules/version"
+	"github.com/threefoldtech/zosv2/modules/zinit"
+)
+
+const redisSocket = "unix:///var/run/redis.sock"
```
zaibon: we can get rid of this one too. I don't think it's used

cmds/internet/main.go
```diff
@@ -0,0 +1,116 @@
...
+const module = "initnet"
```
zaibon: this one too

zaibon (changes requested):
ok LGTM, just 2 lines to remove and still the boot diagram to re-generate and we're good to go

muhamadazmy (approved):

zaibon (approved):

PR #241, opened 2019-09-12
provisiond: allow to remove workload before expiration

### LINKED ISSUE #234: Workload deletion
The jumpscale workload actor/model defines that a reservation can be deleted, which would then delete the workloads, prior to the expiration (https://github.com/threefoldtech/jumpscaleX_threebot/blob/development/ThreeBotPackages/threefold/tfgrid_workloads/models/reservation.toml#L15)

zaibon:
fixes #234


modules/provision/remote_store.go
```diff
@@ -45,17 +45,22 @@ func (s *HTTPStore) Reserve(r *Reservation, nodeID modules.Identifier) error {
 }
 
 // Poll retrieves reservations from BCDB. If all is true, it returns all the reservations
-// for this node, otherwise is return only the reservation never sent yet and do long polling
-func (s *HTTPStore) Poll(nodeID modules.Identifier, all bool) ([]*Reservation, error) {
+// for this node.
+// otherwise it returns only the reservation never sent yet or the reservation that needs to be deleted
+// and do long polling
+func (s *HTTPStore) Poll(nodeID modules.Identifier, all bool, since int64) ([]*Reservation, error) {
```
muhamadazmy: You should use a `Time` object or even a `Duration` instead of an int64. It's not clear what the value of since should be.

muhamadazmy (commented):
You pushed a binary file again 
image

zaibon:
> You pushed a binary file again

I removed it you mean ?

PR #245, opened 2019-09-13
WIP: environment detection

### LINKED ISSUE #190: creates constant file for different flavor of 0-OS
First read about the different flavor of 0-OS in the update documentation: https://github.com/threefoldtech/zosv2/tree/master/docs/upgrade#philosophy

I would like we add a constant file to the 0-OS build that define what value to use for different components of the different flavor.

So the different flavor would be :
- dev: purpose for local development
- test: purpose is for new feature testing and staging
- main: the actual production

This constant file would holds values for:
- URL for the BCDB
- timeouts and interval for provisioning
- ...
- factory to get the different clients that depends on the URL and timeouts ?

maxux:
This is a preparation to fix #190


modules/environment/environmentd.go
```diff
@@ -0,0 +1,60 @@
...
+func NewManager() (modules.EnvironmentManager, error) {
+	fmt.Println("New Environment Manager")
+
+	// TODO: detect running environment
+
+	env := &environmentManager{
+		runningMode: runningDev,
+	}
+
+	if env.runningMode == runningDev {
```
muhamadazmy: why not use `switch` here ?

muhamadazmy (commented):
I would instead do something like 

```go
var (
   development = Environment {
       URL: "...",
       provisionTimeout: 60,
     ...
     }
     testing = ...
)
```

Then return the proper one on NewManager.

maxux:
You can override all the settings using environment variable. This is IMO the easiest way to support new options, new features etc. without touching binaries arguments. You set everything via the environment (or you keep default value).

What do you think ?

PR #253, opened 2019-09-16
WIP: New upgrade procedure

muhamadazmy:


modules/zinit/commands.go
```diff
@@ -3,38 +3,95 @@ package zinit
...
+}
+
+// Is checks if service state is equal to a possible state
+func (s *ServiceState) Is(state PossibleState) bool {
+	return strings.EqualFold(string(s.state), string(state))
+}
+
+// Exited is true if the service state in a (stopped) state
+func (s *ServiceState) Exited() bool {
+	return s.Is(ServiceStateSuccess) || s.Is(ServiceStateError)
```
zaibon: `ServiceStateFailure` is not considered exited ?

bootstrap/etc/zinit/init/bootstrap.sh
```diff
@@ -16,9 +18,13 @@ function bootflist() {
...
 }
 
-BOOTFLIST=https://hub.grid.tf/$(bootflist)
+FLIST=$(bootflist)
+# record the value of the boot flist we used
+echo ${BOOTFLIST} > ${BOOTFILE}
+chmod 0400 ${BOOTFILE}
 
-echo "Bootstraping: ${BOOTFLIST}"
+BOOTFLIST=https://hub.grid.tf/${FLIST}
```
zaibon, line 39: maybe we should avoid hardcoding `https://hub.grid.tf/` here. Or do you see this has a way to ensure we cannot boot from any other source ?

modules/zinit/commands.go
```diff
@@ -115,6 +174,87 @@ func (c *Client) Stop(service string) error {
 	return err
 }
 
+// StartWait stops a service and wait until it exits, or until the timeout
```
zaibon:
```suggestion
// StartWait starts a service and wait until it exits, or until the timeout
```

modules/zinit/commands.go
```diff
@@ -115,6 +174,87 @@ func (c *Client) Stop(service string) error {
...
+			return fmt.Errorf("service '%s' did not start in time", service)
+		default:
+			status, err := c.Status(service)
+			if err != nil {
+				return err
+			}
+
+			if status.Target != ServiceTargetUp {
+				// it means some other entity (another client or command line)
+				// has set the service back to up. I think we should immediately return
```
zaibon:
```suggestion
				// has set the service back down. I think we should immediately return
```

cmds/upgraded/main.go
```diff
@@ -75,11 +117,13 @@ func main() {
 	})
 
 	for {
+		if err := SafeUpgrade(&upgrader); err != nil {
+			//TODO: crash or continue!
```
zaibon, line 122: indeed...

modules/upgrade/hub.go
```diff
@@ -0,0 +1,48 @@
...
+	// HubStorage default hub db
+	HubStorage = "zdb://hub.grid.tf:9900"
+)
+
+// Hub API for f-list
+type Hub struct{}
+
+// URL returns the full url of given flist.
+func (h *Hub) URL(flist string) string {
+	return HubBaseURL + flist
```
zaibon: better to a do a real URL building with `net.url` to detect any error in the flist string passed as argument.

zaibon: Also, does this needs to be a public method ?

modules/upgrade/upgrade.go
```diff
@@ -8,429 +8,288 @@ import (
...
+	return ioutil.WriteFile(verFile, []byte(version), 0400)
 }
 
-func (u *Upgrader) isInSelfUpgrade(version semver.Version) (bool, error) {
-	path := filepath.Join(u.root, "selfupgraded")
-
-	defer os.Remove(path)
+// DefaultBootInfo get boot info set by bootstrap process
+func DefaultBootInfo() BootInfo {
+	flist, _ := ioutil.ReadFile(bootFile)
```
zaibon: why no error checking ?

modules/upgrade/upgrade.go
```diff
@@ -8,429 +8,288 @@ import (
...
-	if !isExecutable(info.Mode().Perm()) {
-		return fmt.Errorf("%s exists but is not executable", name)
+	if err := u.BootInfo.Commit(version); err != nil {
+		return err
 	}
 
-	cmd := exec.Command(path)
-	err = cmd.Run()
+	// one last thing. the self upgrade.
+	// 1- Find where upgraded is running from
```
zaibon: copyFile function already allow you to copy a binary file in place even if a process of this binary is already running.
This `exec mv` trick is not necessary and more complex so I would avoid that.

zaibon (changes requested):
Did a first review of the code, added comments.

Documentation still need to be adapted to reflect new upgrade flow.

I have 2 remarks.
1. I don't like that now the version is an md5 hash of the flist. We need to use semver.
2. We don't have any way to upgrade file shipped from the images like redis or coreX. We need to solve that. I'll open an issue so we can discuss it there. :https://github.com/threefoldtech/zosv2/issues/258

bootstrap/etc/zinit/init/bootstrap.sh
muhamadazmy: It's exactly that actually,l. I think even in production this should not be even configurable and should be hardcoded to a fixed release flist

modules/upgrade/upgrade.go
muhamadazmy: Because there is a case where this file is not set at all (booting with overlay) so the flist string is just empty. The daemon in that case will not perform any update.

modules/upgrade/upgrade.go
muhamadazmy: I thought that u couldn't solve that, since i found a lot of code regarding updating upgrade itself.

modules/zinit/commands.go
muhamadazmy: Good catch

modules/upgrade/upgrade.go
zaibon: Hum could be for the current process you cannot. Let me have another look at the old code see what is what.

muhamadazmy: Well, i remember you had an issue that u couldn't copy over `upgraded` with "file is busy" error or something, hence u had all the logic for the self update.

bootstrap/etc/zinit/init/bootstrap.sh
zaibon: indeed.

modules/upgrade/upgrade.go
zaibon: I just check the old code. copyFile indeed already handle the problem of busy file, so no point doing your trick here.

But the logic around upgraded was to allow to restart the daemon itself when there were a new version of it. So the logic was:
- copy files from flist
- if upgraded has a new version => write some flag file => exit(0)
- upgraded then is restarted. I know runs the new version
- if flag file is there, remove it and avoid stopping itself to continue upgrade of the rest.

I think we need to keep this kind of logic. Where the first things we do is to upgrade `upgraded` daemon. Reason is if we need to introduce some change in the upgrade flow. We want upgraded to get this upgrade first before trying to touch any other file.
Early in the development I had the case where a bug in upgraded prevented me to do a full upgraded. And since this bug happened before upgraded was updated, then I ended in a dead lock situation

muhamadazmy:
@zaibon 

> I don't like that now the version is an md5 hash of the flist. We need to use semver.

I totally understand the reason, and I agree with u. But unfortunately hub and flists does not support tags atm. If we can add tags to flists, the upgraded can check on new tags instead of change in checksum. @maxux is that possible ?

.gitignore
```diff
@@ -8,7 +8,7 @@ tools/bcdb_mock/bcdb_mock
 tools/tfuser/tfuser
 tools/tffarmer/tffarmer
 tools/schemac/schemac
-tools/upgradectl/upgradectl
+tools/updatectl/updatectl
```
zaibon, line 11: I don't think this needs to be renamed

cmds/Makefile
```diff
@@ -1,15 +1,11 @@
-OUT = ../../bin
+OUT = $(shell realpath -m ../bin)
```
zaibon, line 1: :+1:

modules/upgrade/hub.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+// Hub API for f-list
+type Hub struct{}
+
+// MountURL returns the full url of given flist.
+func (h *Hub) MountURL(flist string) string {
+	url, err := url.Parse(HubBaseURL)
+	if err != nil {
+		panic("invalid base url")
```
zaibon, line 31: is this valid to panic here ? why not returning an error ?

modules/upgrade/hub.go
```diff
@@ -0,0 +1,122 @@
...
+// StorageURL return hub storage url
+func (h *Hub) StorageURL() string {
+	return HubStorage
+}
+
+// Info gets flist info from hub
+func (h *Hub) Info(flist string) (info FListInfo, err error) {
+	u, err := url.Parse(HubBaseURL)
+	if err != nil {
+		panic("invalid base url")
```
zaibon, line 46: is this valid to panic here ? why not returning an error ?

modules/upgrade/hub.go
```diff
@@ -0,0 +1,122 @@
...
+	Target  string `json:"target"`
+	Type    string `json:"type"`
+	Updated uint64 `json:"updated"`
+	Hash    string `json:"hash"`
+	Size    uint64 `json:"size"`
+}
+
+// Commit write version to version file
+func (b *FListInfo) Commit(path string) error {
+	f, err := os.OpenFile(path, os.O_CREATE|os.O_WRONLY, 0400)
```
zaibon:
```suggestion
	f, err := os.OpenFile(path, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0400)
```

modules/utils/context.go
```diff
@@ -0,0 +1,50 @@
+package utils
```
zaibon, line 1: Let's not use `utils` as package name, this is a bad practice.
This package deals with zbus signal management, let's find a name that relates to that.

modules/version/version.go
```diff
@@ -20,6 +22,10 @@ var (
 	Dirty = "{dirty}"
 )
 
+var (
+	re = regexp.MustCompile(`^Version:([^@]*)@Revision:([^\(]+)`)
```
zaibon, line 26: semver library doesn't handle case like that ?
So I guess this notation is not a valid semantic version ?

modules/zinit/zinit.go
```diff
@@ -7,38 +7,32 @@ import (
...
 type Client struct {
-	socket string //path to the unix socket
-	conn   net.Conn
+	conn net.Conn
+	scan *bufio.Scanner
 }
 
 // New create a new Zinit client
-func New(socket string) *Client {
+func New(socket string) (*Client, error) {
```
zaibon, line 24: I'm trying to have all the constructor to never return an error.
That's why I had a `New` that just builds the object and then a `Connect` that dials and can return an error.

qemu/overlay/etc/zinit/init/bootstrap.sh
```diff
@@ -0,0 +1 @@
+../../../../../bootstrap/etc/zinit/init/bootstrap.sh
```
zaibon, line 1: this path is out of the repo no ?

qemu/overlay/etc/zinit/init/node-ready.sh
```diff
@@ -0,0 +1 @@
+../../../../../zinit/init/node-ready.sh
```
zaibon, line 1: this path is out of the repo no ?

tools/updatectl/hub.go
```diff
@@ -0,0 +1,105 @@
...
+	baseHubURL = "https://hub.grid.tf/api/flist"
+)
+
+// Hub API
+type Hub struct {
+	base   *url.URL
+	client http.Client
+}
+
+func newTestHub(baseURL, user, token string) (*Hub, error) {
```
zaibon: can you most this into the test file so we don't pollute actual code with testing code.

tools/updatectl/README.md
```diff
@@ -0,0 +1,26 @@
...
+```bash
+# FLIST is the flist to release
+export FLIST=flist-to-release.flist
+# RELEASE is the release name
+export RELEASE=zos:production
+# VERSION is the version tag
+export VERSION=2.0.1
+
+# NOTE: token.jwt is a file that has your valid jwt token for itsyou.online
+updatectl release -t $(cat token.jwt) -f ${FLIST} -r ${RELEASE} ${VERSION}
```
zaibon, line 37: should explain how to get a jwt token

zaibon (changes requested):

.gitignore
muhamadazmy: I didn't rename it, i just added the correct name to .gitignore

modules/upgrade/hub.go
muhamadazmy: The HubBaseURL is a constant in the binary, if you fail to parse this it's a programming error (not runtime error) hence i panic instead.

modules/upgrade/hub.go
muhamadazmy: Same answer as above.

qemu/overlay/etc/zinit/init/bootstrap.sh
muhamadazmy: No, the `bootstrap` directory is at the root of the repo

tools/updatectl/hub.go
muhamadazmy: yup

qemu/overlay/etc/zinit/init/node-ready.sh
muhamadazmy: No, `zinit` is at the root of the repo

modules/version/version.go
muhamadazmy: This is the build version, set by the make file it simply tags the binary with a branch and a git revision number. It does not relate to the version of the release.

I use the git revision hash to check if both binaries are identical or not.

modules/zinit/zinit.go
muhamadazmy: I am not sure about this, it feels like a weird pattern that i have never encounter before. Also it just doubles the work you have to do to get a working client. IMHO if you always have to do New then Connect every time u need a client, you can always have a util method `New` or `Connect` that do that for u. No ?

modules/utils/context.go
muhamadazmy: Actually it's not related to zbus at all. it's just a generic context cancellation mechanism on signal receiving. I use it to handle signals to gracefully shutdown services.

may be rename this to `signal` ? I couldn't come up with a better name.

zaibon (approved):

PR #264, opened 2019-09-19
provisiond: use unix socket to talk to 0-db servers

zaibon:


modules/provision/zdb.go
```diff
@@ -186,29 +191,21 @@ func zdbProvision(ctx context.Context, reservation *Reservation) (interface{}, e
...
-		return nil, errors.Wrapf(err, "failed to set size on namespace %s in 0-db at %s", nsID, addr)
+	defer netNS.Close()
+
+	if err := netNS.Do(func(_ ns.NetNS) error {
+		slog.Info().
+			Str("name", nsID).
+			Str("netns", netNSName).
+			Str("addr", addr).
+			Msg("create 0-db namespace")
+		return createZDBNamespace(addr, nsID, config)
```
muhamadazmy: If you already are inside the zdb namespace, why do u need the address? the port should be deterministic, no ? unless all zdbs are running inside the same namespace. is that the case?

muhamadazmy (approved):

modules/provision/zdb.go
zaibon: Well I need to know to which IP I need to connect. 
And no, each 0-db container runs in their own network namespace

maxux (approved):

muhamadazmy (approved):

PR #276, opened 2019-09-25
Improve stresstest with support of feedback

maxux:
https://asciinema.org/a/HCww9jhh3hwKIarpjcR2vdL1i


maxux:
Closed, there is one merge which should not be there

PR #278, opened 2019-09-25
allow user to choose ipv4 range per network resource

### LINKED ISSUE #268: networkd: allow user to choose ipv4 range per network resource
After extensive discussion about how `networkd` works right now. It has been decided that going all the way deterministic is not acceptable. We need a way to let a user choose the ipv4 subnet he wants to use for his network as well as the subnet per network resource.

This new requirements has a pretty big impact on the all architecture. I'll try to summarize all the changes required to make this work.

follow progress at: https://github.com/threefoldtech/zosv2/pull/278

zaibon:
- [x] remove the concept of prefix allocation given from a farm per network resource
- [x] exit point are not required to create a network anymore
  - [x] all the node must allow traffic to exit directly from them
- [x] Introduction of a DMZ network namespace to allow masquerading of exit traffic and keep isolation between network resource
- [x] wireguard mesh must be 100% under control of the user, not more determinism for listen port or address.
  - [x] this means we need to re-introduce port management again per node.

fixes https://github.com/threefoldtech/zosv2/issues/268


modules/network.go
```diff
@@ -39,212 +41,46 @@ type Networker interface {
...
+	// IP range of the network, must be an IPv4 /16
+	IPRange *net.IPNet `json:"ip_range"`
 
-const (
-	// ReachabilityV4Hidden The Node lives in an RFC1918 space, can't listen publically
-	ReachabilityV4Hidden ReachabilityV4 = iota
-	// ReachabilityV4Public The Node's Wireguard interfaces listen address is reachable publicly
-	ReachabilityV4Public
-)
+	NetResources []*NetResource `json:"net_resources"`
```
muhamadazmy: Why u create a List of pointers instead of a list of struct? this beside it looks ugly, it's not really useful since it allows a nil element in the list.

modules/network/networker.go
```diff
@@ -54,71 +57,82 @@ func NewNetworker(identity modules.IdentityManager, tnodb TNoDB, storageDir stri
...
-	}
-
-	if n.Exit == nil {
-		return fmt.Errorf("Exit point cannot be empty")
-	}
-
-	if n.AllocationNR < 0 {
-		return fmt.Errorf("AllocationNR cannot be negative")
-	}
+	// TODO
```
muhamadazmy: delete commented code?

tools/tfuser/cmds_network.go
```diff
@@ -1,167 +1,167 @@
...
-	if err := json.NewDecoder(f).Decode(r); err != nil {
-		return nil, errors.Wrapf(err, "failed to decode json encoded reservation at %s", name)
-	}
-
-	if err := json.Unmarshal(r.Data, network); err != nil {
-		return nil, errors.Wrapf(err, "failed to decode json encoded network at %s", name)
-	}
-	return network, nil
-}
+// import (
```
muhamadazmy, line 3: Either delete or adapt this code to current changes

zaibon: I opened another issue to tackle this: https://github.com/threefoldtech/zosv2/issues/287
So I left the code there so we can base the work on it, but still wanted to have our CI to pass for this PR

PR #281, opened 2019-09-26
Implement gedis client

zaibon:


modules/identity.go
```diff
@@ -18,6 +22,11 @@ func (s StrIdentifier) Identity() string {
 	return string(s)
 }
 
+// Hex implements the Identifier interface
+func (s StrIdentifier) Hex() string {
```
muhamadazmy: Should this method be added to the `Identifier` interface ? and how it's different from `Identity()` method. 

I mean the KeyPair could return the hex representation as a result to `Identity()` call, no ?

modules/network/tnodb/store.go
```diff
@@ -205,28 +205,11 @@ func (s *httpTNoDB) PublishInterfaces(local modules.Identifier) error {
...
-		GWs:   make([]string, len(gws)),
-		Type:  types.MacVlanIface, //TODO: allow to chose type of connection
-	}
-
-	for i := range ips {
-		output.IPs[i] = ips[i].String()
-		output.GWs[i] = gws[i].String()
-	}
-
+func (s *httpTNoDB) SetPublicIface(node modules.Identifier, pub *types.PubIface) error {
```
muhamadazmy, line 158: This method will probably break against the current mock.

muhamadazmy (approved):

modules/gedis/commands_identity.go
```diff
@@ -1,281 +1,303 @@
...
+}
 
-	b, err := json.Marshal(req)
-	if err != nil {
-		return err
+func nodeFromSchema(node directory.TfgridNode2) types.Node {
+	return types.Node{
+		NodeID: node.NodeID,
+		FarmID: node.FarmID,
+		Ifaces: func() []*types.IfaceInfo {
```
zaibon, line 159: Why is this a function ?

modules/gedis/commands_identity.go
```diff
@@ -1,281 +1,303 @@
...
+			var r []*types.IfaceInfo
+			for _, iface := range node.Ifaces {
+				v := infFromSchema(iface)
+				r = append(r, &v)
+			}
+			return r
+		}(),
+		PublicConfig: func() *types.PubIface {
```
zaibon, line 167: Why is this a function ?

modules/gedis/commands_identity.go
```diff
@@ -1,281 +1,303 @@
...
+	if err := json.Unmarshal(resp, &out); err != nil {
+		return nil, err
 	}
-
-	return nil
+	f := farmFromSchema(out)
+	return &f, nil
 }
 
+// //UpdateFarm implements modules.IdentityManager interface
```
zaibon: Let's remove if not needed anymore.

modules/gedis/commands_network.go
```diff
@@ -1,278 +1,316 @@
...
-	}
-
-	versions := make(map[modules.NetID]uint32)
-	if err := json.Unmarshal(resp, &versions); err != nil {
-		return nil, err
-	}
-
-	return versions, nil
-}
+// //SelectExitNode implements network.TNoDB interface
```
zaibon, line 217: Let remove if not needed anymore

modules/gedis/commands_provision.go
```diff
@@ -0,0 +1,131 @@
...
+// Get implements provision.ReservationGetter
+func (g *Gedis) Get(id string) (*provision.Reservation, error) {
+	result, err := Bytes(g.Send("workload_manager", "workload_get", Args{
+		"gwid": id,
+	}))
+
+	if err != nil {
+		return nil, err
+	}
+	fmt.Println(string(result))
```
zaibon:
```suggestion
```

modules/gedis/gedis_test.go
```diff
@@ -25,65 +22,65 @@ func TestPing(t *testing.T) {
 	assert.Equal(t, "PONG", resp)
 }
 
-func TestNodesCRUD(t *testing.T) {
-	gedis := testClient(t)
+// func TestNodesCRUD(t *testing.T) {
```
zaibon: Let's remove this since this is never going to work in CI. Instead we should think about some compatibility test we can enable in CI. Something that would deploy a gedis server and then run some test using our client.

modules/identity.go
zaibon: This is actually not used anywhere, we should remove.

modules/identity/keys.go
```diff
@@ -27,6 +28,11 @@ func (k KeyPair) Identity() string {
 	return base58.Encode(k.PublicKey)
 }
 
+// Hex implements the Identifier interface
```
zaibon: don't think it's used anywhere we should remove it

modules/gedis/commands_identity.go
muhamadazmy: IMHO, this looks as `inline` conversion of a type. and avoid having so many variables declared in the same context. 

Note: this is **NOT** a function, it's  just an inline function call to do the conversion of the type

modules/gedis/commands_identity.go
muhamadazmy: Same, just inline conversion of the type. avoiding having to do all the processing (and variable declaration) in the function body.

PR #284, opened 2019-09-30
Update QEMU doc to add IPv6 notes

maxux:


qemu/README.md
```diff
@@ -35,9 +35,16 @@ To prepare your environment call `make prepare`. This will:
...
 
+You can specify your farm id (if you have one) in the command line with:
+```
+make FARMERID=A3y5F8CoHVZiq3Sxxxx start
+```
+
 ### Reach internet inside the VM
 
-You need to be able to reach internet inside the VM, 0-OS needs internet.
+You need to be able to reach internet inside the VM, 0-OS needs internet. The system will starts
```
zaibon, line 45: The IPv6 requirement is not really true. At least not when https://github.com/threefoldtech/zosv2/pulls will be merged.

maxux: And what about zdb deployment ?

PR #296, opened 2019-10-01
networkd: identify uniquely the networks

### LINKED ISSUE #286: networkd: find a way to identity uniquely the networks
Since https://github.com/threefoldtech/zosv2/pull/278
the network is not deterministic anymore. So we cannot use the ipv6 allocation from source of uniqueness. So all the component that are defined in the host network namespace can potentially conflicts. We need to find a way to generate unique ID for them:

- wireguard interface
- network resource network namespace
- network resource bridge
- veth interface for the bridge of the network resource

zaibon:
fixes #286


delandtj (approved):
cool cool

PR #297, opened 2019-10-01
First implementation of Orphanage farm

maxux:


tools/tffarmer/main.go
```diff
@@ -35,7 +35,7 @@ func main() {
 		cli.StringFlag{
 			Name:   "tnodb, u",
 			Usage:  "URL of the TNODB",
-			Value:  "https://tnodb.dev.grid.tf",
+			Value:  "https://bcdb.dev.grid.tf",
```
zaibon, line 38: you can also rename the flag to `bcdb`

zaibon:
We should also prevent provisiond to deploy when there is no farmer_id defined.

maxux:
Maybe not prevent everything, it's cool to keep `debug` for example ?

zaibon:
> Maybe not prevent everything, it's cool to keep debug for example ?

I don't really see the point. If there is no farmer_id, then the node should not be used for anything. So there wouldn't be anything to debug anyhow.

cmds/provisiond/main.go
```diff
@@ -56,6 +56,14 @@ func main() {
 		log.Fatal().Msgf("fail to connect to message broker server: %v", err)
 	}
 
+	env := environment.Get()
```
zaibon: it looks like you can do this even before creating the redis client at line 54

zaibon, line 54: it looks like you can do this even before creating the redis client at line 54

zaibon (changes requested):

zaibon (approved):

PR #298, opened 2019-10-01
Bootstrap default flist to use based on runmode

muhamadazmy:


bootstrap/etc/zinit/init/bootstrap.sh
```diff
@@ -17,21 +15,48 @@ function retry() {
     done
 }
 
-# bootflist reads the boot flist name from kernel cmd
-# TODO: this should probably be not allowed at somepoint to
-function bootflist() {
+function param() {
```
maxux: `function` is deprecated

bootstrap/etc/zinit/init/bootstrap.sh
```diff
@@ -17,21 +15,48 @@ function retry() {
...
             return 0
         fi
     done
 
-    echo ${DEFAULT_FLIST}
+    return 1
 }
 
-FLIST=$(bootflist)
+function default_param() {
```
maxux: `function` is deprecated

bootstrap/etc/zinit/init/bootstrap.sh
```diff
@@ -17,21 +15,48 @@ function retry() {
     done
 }
 
-# bootflist reads the boot flist name from kernel cmd
-# TODO: this should probably be not allowed at somepoint to
-function bootflist() {
+function param() {
+    local KEY=$1
+    local KEY=$(echo $KEY=)
```
maxux: What about `local KEY="${1}="` ?

bootstrap/etc/zinit/init/bootstrap.sh
muhamadazmy: Seriously ? I didn't know that

bootstrap/etc/zinit/init/bootstrap.sh
muhamadazmy: Haha, i must have been low on coffee !

maxux (approved):

PR #302, opened 2019-10-02
update tfuser to work with revised network and BCDB

### LINKED ISSUE #287: update tfuser to work with revised network
Since https://github.com/threefoldtech/zosv2/pull/278/commits/8a4b100375139ef7f0d8b13573c9d6389e014385 the network definition has changed and tfuser doesn't build anymore.

We should update tfuser to make it compatible with the new network design.

zaibon:
fixes #287


muhamadazmy (commented):
I just have 2 comments:
- It seems the watcher.go under the network pkg is no longer needed, it's better that you also delete that code because it's confusing.
- tffarmer still using the temporary TNoDBUtils interface https://github.com/threefoldtech/zos/blob/tfuser/pkg/network/tnodb.go#L26) which was done to separate the parts of the api that need to be implemented for zos operations and the other part for the tooling (tfuser, and tffarmer)

PR #308, opened 2019-10-03
enable CI

muhamadazmy:
Early draft to show the changes in tests


zaibon:
Had a first quick overview, this looks realy cool.

One thing I forgot to mention. In the github action workload, to run the test we call `make coverage`:
https://github.com/threefoldtech/zos/blob/da3d88e815007cac78134a8b89758092708d6a3e/.github/workflows/test.yml#L26-L29
which in turn execute this script: https://github.com/threefoldtech/zos/blob/master/pkg/buildscripts/go-coverage.sh

So you have to also remove the module from the grep to enable the test on them. This is a bit confusing mess. maybe we could improve the makefile/workflow a bit

go.mod
```diff
@@ -9,51 +9,37 @@ require (
...
-	github.com/opencontainers/runc v0.1.1 // indirect
 	github.com/opencontainers/runtime-spec v1.0.1
 	github.com/pkg/errors v0.8.1
 	github.com/rs/zerolog v1.15.0
 	github.com/shirou/gopsutil v2.18.12+incompatible
 	github.com/stretchr/testify v1.3.0
-	github.com/syndtr/gocapability v0.0.0-20180916011248-d98352740cb2 // indirect
 	github.com/tcnksm/go-input v0.0.0-20180404061846-548a7d7a8ee8
 	github.com/threefoldtech/zbus v0.0.0-20190711124326-09379d5f12e0
+	github.com/threefoldtech/zosv2/modules v0.0.0-20191001184911-b6d8bc81a7f0
```
zaibon, line 36: that doesn't look right. This import path doesn't exist anymore.

pkg/flist/test.sh
```diff
@@ -0,0 +1,2 @@
+(sleep 3s && echo "exited") &
```
zaibon: is this file needed ?

pkg/gedis/gedis_test.go
```diff
@@ -0,0 +1,122 @@
...
+	return &pool, &conn
+}
+
+func mustMarshal(t *testing.T, v interface{}) []byte {
+	r, err := json.Marshal(v)
+	require.NoError(t, err)
+	return r
+}
+
+func EqualJSON(t *testing.T, expected, actual []byte) {
```
zaibon, line 82: is this function needs to be exported ?

pkg/storage/filesystem/btrfs_test.go
```diff
@@ -0,0 +1,105 @@
+/*
+NOTE:
+  This test file tries to create loop devices to work against on some
+  sparse files, to avoid causing any actual changes to permanent storage
+  on the test machine. This comes with a price that this test file
+  need to run as root to be able to run the `losetup` commands.
+
+  this can be easily done as
+  `sudo GOPATH=$GOPATH go test -v ./...`
```
zaibon: this comment can be removed. the test now mock calls to btrfs

zaibon (approved):
LGTM

pkg/gedis/gedis_test.go
muhamadazmy: No it does not, but it's in a test file so it won't be available on the package anyway.

pkg/flist/test.sh
muhamadazmy: nope

Issue #19, opened 2019-03-21
Define container profiles


zaibon:
After the discussion from https://github.com/threefoldtech/0-OS_research/issues/5#issuecomment-475149757
We agreed we will provide different container "profiles". 
This issue is about choosing what kind of profile we want and write the config.json file for them.

muhamadazmy:
Link to specs (which is still WIP by the time we added this link)
https://github.com/threefoldtech/zosv2/blob/flist-meta/specs/container/readme.md#oci-bundle

zaibon:
closing this, nobody seems to care about that anyhow.

PR #310, opened 2019-10-07
Provisioning fixes

zaibon:


pkg/container/container.go
```diff
@@ -121,6 +121,18 @@ func (c *containerModule) ensureNamespace(ctx context.Context, client *container
 	return service.Create(ctx, namespace, nil)
 }
 
+func removeRunMount() oci.SpecOpts {
```
muhamadazmy, line 124: Can we change the "seed" spec we use instead of removing the run mount ? just wondering. 
Otherwise, i think this is a good solution.

muhamadazmy (approved):

pkg/container/container.go
zaibon: I would love to and there is an open issue about that for quite some time now: https://github.com/threefoldtech/zos/issues/19
But for some reason no progress has ever been made on that topic

PR #317, opened 2019-10-08
Networkd fixes

zaibon:


delandtj:
looks ok to me, but the routes to all netresources could be subject to change into more fine-grained routes instead of one catch-all ... future

delandtj (approved):
go for it ;-)

muhamadazmy (approved):
I would like to have @delandtj opinion on this PR. to validated "wiring" changes. From coding point of view changes LGTM

PR #318, opened 2019-10-09
contd: read environment variable from .stratup.toml if it exists and ensure /etc/resolv.conf

zaibon:


pkg/container/container.go
```diff
@@ -157,6 +162,19 @@ func (c *containerModule) Run(ns string, data pkg.Container) (id pkg.ContainerID
 		return id, ErrEmptyRootFS
 	}
 
+	startupFile, err := os.Open(filepath.Join(data.RootFS, ".startup.toml"))
```
muhamadazmy: Can this be moved to another method, it looks like a lot of code in the Run method . Something like `updateEnv(&data, path.Join(root, ".startup.toml"))`

pkg/container/container.go
```diff
@@ -157,6 +162,19 @@ func (c *containerModule) Run(ns string, data pkg.Container) (id pkg.ContainerID
...
+
+		log.Info().Msg("startup file found, extracting environment variable")
+		envs, err := readEnvs(startupFile)
+		if err != nil {
+			return id, errors.Wrap(err, "failed to extract environment variable from startup file")
+		}
+
+		data.Env = mergeEnvs(envs, data.Env)
+	}
+
```
muhamadazmy, line 87: I see that u still do not use the `exec` from the `startup.toml` as an entrypoint for the container.
The procedure should be the following
- check if .startup.toml exists (opens with no error)
- load the `entry` point.
- if an `entrypoint` is already provided by the reservation, use it.
- if no `entrypoint` is provided, use the one from .startup.tom
- update the env

muhamadazmy (changes requested):

pkg/container/container.go
```diff
@@ -298,3 +309,72 @@ func (c *containerModule) Delete(ns string, id pkg.ContainerID) error {
...
+	for _, ns := range namespaces {
+		if ns == namespace {
+			return nil
+		}
+	}
+
+	return service.Create(ctx, namespace, nil)
+}
+
+func setResolvConf(root string) error {
```
muhamadazmy: I don't get what this method is trying to do (i know it is trying to write the etc/resolv.conf with static content) but the check and the stats, i don't get. I mean why exactly didn't u just use `ioutil.WriteFile` method. ?

pkg/container/container.go
```diff
@@ -298,3 +309,72 @@ func (c *containerModule) Delete(ns string, id pkg.ContainerID) error {
...
+	}
+	if info.Size() == 0 {
+		_, err = f.WriteString(tmp)
+	}
+	return err
+}
+
+func applyStartup(data *pkg.Container, path string) error {
+	f, err := os.Open(path)
+	if err == nil {
```
muhamadazmy, line 328: instead of indenting all the method implementation, it would better if u do 
```go
if err != nil {
    return err
}
defer f.Close()
// everything else goes here.
```
no the method just look weird can confusing.

pkg/container/startup.go
```diff
@@ -0,0 +1,68 @@
...
+		envs = append(envs, fmt.Sprintf("%s=%s", k, v))
+	}
+	return envs
+}
+
+// mergeEnvs merge a into b
+// all the key from a will endup in b
+// if a key is present in both, key from a are kept
+func mergeEnvs(a, b []string) []string {
+	ma := make(map[string]string, len(a))
```
muhamadazmy: Why this has to be 2 dicts instead of one, the merge can done in the 2 loops directly but now u need 3 loops instead. Also did u test what will happen if u have `["KEY=V1", "KEY=V2"]` passed directly to process Env, because i think libc will just do the dedubing for u. this can be easily tested with one exec.Command line of `env`

muhamadazmy (changes requested):
More few comments

pkg/container/container.go
zaibon: Cause I don't want to overwrite the content if there is any. So here I'm writing only of the file doesn't exist or if it's empty

pkg/container/container.go
zaibon: Well this is on purpose cause I really just want to do something when the file exists.
If it doesn't the this function does nothing. I though having it in the `err == nil` block would make this more obvious.

I can change it to a `os.IsNotExists` check if you prefer.

pkg/container/startup.go
zaibon: > Why this has to be 2 dicts instead of one, the merge can done in the 2 loops directly but now u need 3 loops instead. 

Right... :-)

>  Also did u test what will happen if u have ["KEY=V1", "KEY=V2"] passed directly to process Env, because i think libc will just do the dedubing for u.

I don't want to rely on a side effect when I can be sure to do it myself here.

pkg/container/container.go
muhamadazmy: Does docker do that? because to me this sound weird. A container should always have a valid dns related to it's runtime network, not build time network. So this sounds a bit weird to me.

pkg/container/container.go
muhamadazmy: Oh, i see. in that case i think it's fine.

pkg/container/container.go
maxux: Docker mount host resolv.conf to container resolv.conf using volume mount (AFAIK)

zaibon: yes as far as I've seen. Docker does write a file on the host filesystem and mounts it inside the container when it starts.
Docker does the same for hosts hostname and hosts file.

I can also do it the same way if you prefer. It just that since we already have a copy on write layer with flist I though maybe this is easier.

But the containerd has indeed a set of methods to deal with such a things:
```go
// WithHostHostsFile bind-mounts the host's /etc/hosts into the container as readonly
func WithHostHostsFile(_ context.Context, _ Client, _ *containers.Container, s *Spec) error {
	s.Mounts = append(s.Mounts, specs.Mount{
		Destination: "/etc/hosts",
		Type:        "bind",
		Source:      "/etc/hosts",
		Options:     []string{"rbind", "ro"},
	})
	return nil
}

// WithHostResolvconf bind-mounts the host's /etc/resolv.conf into the container as readonly
func WithHostResolvconf(_ context.Context, _ Client, _ *containers.Container, s *Spec) error {
	s.Mounts = append(s.Mounts, specs.Mount{
		Destination: "/etc/resolv.conf",
		Type:        "bind",
		Source:      "/etc/resolv.conf",
		Options:     []string{"rbind", "ro"},
	})
	return nil
}

// WithHostLocaltime bind-mounts the host's /etc/localtime into the container as readonly
func WithHostLocaltime(_ context.Context, _ Client, _ *containers.Container, s *Spec) error {
	s.Mounts = append(s.Mounts, specs.Mount{
		Destination: "/etc/localtime",
		Type:        "bind",
		Source:      "/etc/localtime",
		Options:     []string{"rbind", "ro"},
	})
	return nil
}
```

PR #321, opened 2019-10-09
provision testing

muhamadazmy:


pkg/network.go
```diff
@@ -47,7 +48,7 @@ type Network struct {
 	//unique id inside the reservation is an autoincrement (USE AS NET_ID)
 	NetID NetID `json:"net_id"`
 	// IP range of the network, must be an IPv4 /16
-	IPRange *net.IPNet `json:"ip_range"`
+	IPRange schema.IPRange `json:"ip_range"`
```
zaibon: This is not right. Now we leaking gedis/jsx types system into zos code base. We specifically decided not to do that and to keep all the JSX schema to zos conversion into the gedis package. So we need to revert this and all the code that has been changed because of this.

pkg/network.go
```diff
@@ -56,7 +57,7 @@ type Network struct {
 type NetResource struct {
 	NodeID string `json:"node_id"`
 	// IPV4 subnet from network IPRange
-	Subnet *net.IPNet `json:"subnet"`
+	Subnet schema.IPRange `json:"subnet"`
```
zaibon: Same here, don't leak jsx schema into zos codebase

pkg/provision/provision_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"context"
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+	"github.com/threefoldtech/zbus"
+)
+
+func MustMarshal(t *testing.T, v interface{}) []byte {
```
zaibon, line 13: should this be exported ?

pkg/provision/provision_test.go
```diff
@@ -0,0 +1,82 @@
...
+	bytes, err := json.Marshal(v)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return bytes
+}
+
+type TestClient struct {
```
zaibon, line 21: should this be exported ?

pkg/provision/provision_test.go
```diff
@@ -0,0 +1,82 @@
...
+	var str string
+	var rerr *zbus.RemoteError
+	require.NoError(response.Unmarshal(0, &str))
+	require.NoError(response.Unmarshal(1, &rerr))
+
+	require.Equal("result", str)
+	require.Nil(rerr)
+}
+
+type TestOwnerCache struct {
```
zaibon, line 61: should this be exported ?

pkg/provision/source_test.go
```diff
@@ -5,36 +5,72 @@ import (
...
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
 	"github.com/threefoldtech/zos/pkg"
 )
 
-type testReservationStore struct {
-	fReserve func(r Reservation, nodeID pkg.Identifier) error
-	fPoll    func(nodeID pkg.Identifier, all bool, since time.Time) ([]*Reservation, error)
-	fGet     func(id string) (Reservation, error)
+type TestPollSource struct {
```
zaibon, line 13: should this be exported ?

pkg/provision/source_test.go
```diff
@@ -5,36 +5,72 @@ import (
...
+		}, ErrPollEOS)
+
+	reservations := []*Reservation{}
+	for res := range chn {
+		reservations = append(reservations, res)
+	}
+
+	require.Len(reservations, 2)
+	require.Equal("res-1", reservations[0].ID)
+}
```
zaibon, line 44:
```suggestion
require.Equal("res-2", reservations[1].ID)
```
Let's test them all since they are only 2 results

pkg/provision/volume_test.go
```diff
@@ -0,0 +1,91 @@
...
+	const module = "storage"
+	version := zbus.ObjectID{Name: "storage", Version: "0.0.1"}
+
+	reservation := Reservation{
+		ID:   "reservation-id",
+		User: "user",
+		Type: VolumeReservation,
+	}
+
+	// force creation by returning an error
```
zaibon:
```suggestion
	// force decomission by returning a nil error
```

pkg/provision/zdb.go
```diff
@@ -48,38 +48,41 @@ type ZDBResult struct {
...
+}
+
 // Get returns the container ID where namespace lives
 // if the namespace is not found an empty string and false is returned
-func (z *ZDBMapping) Get(namespace string) (string, bool) {
+func (z *zdbMapping) Get(namespace string) (string, bool) {
 	z.RLock()
 	defer z.RUnlock()
 
-	if z.m == nil {
```
zaibon, line 62: I do like to have struct that have a usable default value. So I would keep this so someone creating a zdbMapping using a literal notation still have a working object.

pkg/provision/zdb.go
```diff
@@ -48,38 +48,41 @@ type ZDBResult struct {
...
 	return id, ok
 }
 
 // Set saves the mapping between the namespace and a container ID
-func (z *ZDBMapping) Set(namespace, container string) {
+func (z *zdbMapping) Set(namespace, container string) {
 	z.Lock()
 	defer z.Unlock()
 
-	if z.m == nil {
```
zaibon, line 75: Same here, we can keep this so allow a usable default value of the zdbMapping struct

zaibon (changes requested):

pkg/network.go
muhamadazmy: schema.IPRange is not a (jsx) schema type and not generated. It's just a wrapper around net.IPNet that can do json serialization/deserialization. Then i use this type in generated schema structures for the same reason. 

Because i switched to this type, i was able to remove lots of custom Marshaling code.

pkg/network.go
muhamadazmy: Same as my answer above. This type is defined here and is not generated.

https://github.com/threefoldtech/zos/blob/master/pkg/schema/types.go#L145

pkg/provision/provision_test.go
muhamadazmy: It doesn't really matter because it's defined in a `_test.go` file so it doesn't get exported.

pkg/provision/provision_test.go
muhamadazmy: It doesn't really matter because it's defined in a _test.go file so it doesn't get exported.

pkg/provision/provision_test.go
muhamadazmy: It doesn't really matter because it's defined in a _test.go file so it doesn't get exported.

pkg/provision/source_test.go
muhamadazmy: It doesn't really matter because it's defined in a _test.go file so it doesn't get exported.

:smile:

pkg/provision/zdb.go
muhamadazmy: The can't create it because `zdbMapping` is not longer exposed, and they have to use the `NewZDBMapping` method to create one.

pkg/provision/zdb.go
muhamadazmy: The type is not exposed anymore (has been renamed to zdbMapping), so no way someone can create an instance without calling the `NewZDBMapping` method.

pkg/network.go
zaibon: Ok but I still don't like we depend on a type that is versioning in another module. I get that it indeed removed quite some code so that's good. But let's then move this type into the network package. Also, is IPRange the proper name for it ?

pkg/provision/zdb.go
zaibon: That is only true for other package, provision package can still create it.

pkg/network.go
muhamadazmy: the name unfortunately (while not the best) is inherited from the jsx `iprange` type. About moving the type out of the schema module is not okay, because a generated type (like the generated types in gedis client) will reference the `schema.IPRange` type for each `iprange` type in jsx schema. 

Copying the code also is not a good solution, but if you really insist, i can create a type alias in the network package, so if we want to separate the types later, we can still do it. What do u think?

pkg/provision/zdb.go
muhamadazmy: in that case, developer of `provision` package should be aware of the implementation. I would rather have one initialization at the `new` method, instead of a check on each method call (even if there is not gonna be many of them)

pkg/network.go
zaibon: > the name unfortunately (while not the best) is inherited from the jsx iprange type.

I know but we said we don't want to leak anything from the JSX types into the main code base. 

>  About moving the type out of the schema module is not okay, because a generated type (like the generated types in gedis client) will reference the schema.IPRange type for each iprange type in jsx schema.

I don't understand what is the problem ? A conversion between types is already happening in gedis package anyhow.  And if anything, types from our code base should spreads into the client that implement interface with outside components, not the other way around where we end up using stuff from outside into the core core of 0-OS.

> Copying the code also is not a good solution, but if you really insist, i can create a type alias in the network package, so if we want to separate the types later, we can still do it. What do u think?

Not happy with that either. If you want to have a type that implement the marshaling around an `net.IPNet`, fine, but then it must be define in our network package. Not an alias from another type and not a type from an external package.

pkg/provision/zdb.go
zaibon: > in that case, developer of provision package should be aware of the implementation.

Yes, but why just removing code that give you this kind of protection for free ?? This type has 2 method, so 100% of the case were covered. Now we are just introducing more change to miss-use the type

pkg/provision/container.go
```diff
@@ -70,16 +74,20 @@ func containerProvision(ctx context.Context, reservation *Reservation) (interfac
 	_, err := containerClient.Inspect(tenantNS, pkg.ContainerID(containerID))
 	if err == nil {
 		log.Info().Str("id", containerID).Msg("container already deployed")
-		return containerID, nil
+		return ContainerResult{
+			ID: containerID,
+			//IMPORTANT
+			//TODO: set IPv4 and IPv6 here
```
zaibon, line 80: forgot to implement the todo ?

zaibon (approved):

PR #322, opened 2019-10-10
cleaning up network namespace after deleting container

### LINKED ISSUE #314: provisiond: container network namespace are not properly cleanup after decomission

zaibon:
fixes #314


muhamadazmy (approved):
LGTM. Only on comment (may be later)
The containerProvision should also take care to `leave` if an error occured in the provision process after the join call was made.

PR #324, opened 2019-10-11
identityd: send proper version during node registration

### LINKED ISSUE #313: identityd: send proper version during node registration
https://github.com/threefoldtech/zos/blob/1f23fb12d6105eb79933b5706656082df2cb15d5/cmds/identityd/main.go#L87

zaibon:
fixes #313


cmds/identityd/main.go
```diff
@@ -52,6 +54,13 @@ func main() {
 		log.Fatal().Err(err).Msg("failed to create module root")
 	}
 
+	redis, err := zbus.NewRedisClient(msgBrokerCon)
+	if err != nil {
+		log.Fatal().Err(err).Msg("failed to create zbus client")
+	}
+
+	upgrade := stubs.NewUpgradeModuleStub(redis)
```
muhamadazmy: I really don't like the idea that `identity` depends on upgrade for its operation. Even in zinit units, upgraded indirectly depends on identityd (which is considered a boot service).

muhamadazmy (changes requested):
I don't really like this cross dependencies between modules. 
- zinit will start upgraded only when idendityd is running
- identityd will not register the node until upgraded is running.

Doing registration before even upgraded is done upgrading to latest will already register the node with "old" version, then may be later will re-register with latest version. 

IMHO: I think it's better (and makes more sense) to have both upgraded and identity merged in one single daemon. (upgraded)

Upgraded should do the following:
- On start, do update to latest version
- Register the node 
- Check new version available and update if necessary 
- On update, re-register the node.
...

zaibon:
I very much like the idea of merging things together, they are indeed very inter-depended.

muhamadazmy (approved):

Issue #330, opened 2019-10-15
flistd: flist mountpoint have 750 permission


zaibon:
The normal behavior should be that the mountpoint have the same permission as `/` in the flist

PR #331, opened 2019-10-16
flistd: use 755 permission when creating flistd directories

zaibon:


muhamadazmy (approved):
LGTM. But i still find it weird, because i tried to use `g8ufs` directly with the same flist, and the root perm seems to be fine.

zaibon:
Yeah I know, this is a workaround until https://github.com/threefoldtech/zos/issues/330 can be figured out.

PR #332, opened 2019-10-16
Upgrade Uninstall

muhamadazmy:
Support removing old files and services. This will make sure only the needed binaries are available on the system.


cmds/upgraded/main.go
```diff
@@ -188,39 +177,43 @@ func main() {
...
+		}
+
+		version, err := upgrader.Version()
+		if err != nil {
+			log.Fatal().Err(err).Msg("failed to read current version")
+		}
+
+		log.Info().Str("version", version.String()).Msg("current")
+
+		if _, err = idStore.RegisterNode(nodeID, farmID, version.String()); err != nil {
```
zaibon: Maybe we should also have a backoff retry here in case there is a network issue that prevent to register the version directly ?

pkg/upgrade/hub.go
```diff
@@ -120,3 +130,48 @@ func (b *FListInfo) Version() (semver.Version, error) {
...
+	}
+
+	u.Path = filepath.Join("api", "flist", flist)
+	response, err := http.Get(u.String())
+	if err != nil {
+		return nil, err
+	}
+
+	defer response.Body.Close()
+	defer ioutil.ReadAll(response.Body)
```
zaibon, line 167: Whis is this needed ? you already consume the body with the json decoder

pkg/upgrade/hub.go
```diff
@@ -81,12 +84,19 @@ func LoadInfo(path string) (info FListInfo, err error) {
...
+	Target     string `json:"target"`
+	Type       string `json:"type"`
+	Updated    uint64 `json:"updated"`
+	Hash       string `json:"hash"`
+	Size       uint64 `json:"size"`
+	Repository string `json:"-"`
+}
+
+// File is the file of an flist
+type File struct {
```
zaibon: Seems this type doesn't need to be exposed

pkg/upgrade/hub.go
```diff
@@ -120,3 +130,48 @@ func (b *FListInfo) Version() (semver.Version, error) {
...
+	name := b.Name
+	if b.Type == "symlink" {
+		name = b.Target
+	}
+
+	return filepath.Join(b.Repository, name)
+}
+
+// Files gets the list of the files of an flist
+func (b *FListInfo) Files() ([]File, error) {
```
zaibon: And this method also doesn't need to be exposed

zaibon (changes requested):

cmds/upgraded/main.go
muhamadazmy: Idd, will add it

pkg/upgrade/hub.go
muhamadazmy: the line below, I check on the status of the response, and then return an error if the status is not OK. in that case, the body will not be consumed, and hence the connection will not be reused (according to the docs)

So I found that it's a good practice to always consume the body in a `defer` to prevent this.

pkg/upgrade/hub.go
muhamadazmy: While idd it's used by the Updater only, i still find it useful to have it exposed, just for the sake of the API completion.

Same reason I also exposed the `File` struct. A user of the pkg, will find a usable Hub client

zaibon: Well I tend to try to keep the public API as small as possible.
If we want to provide a proper Hub lib, then we should extract it into its own package separate from the Upgrade type so it can really be reused by other.

muhamadazmy: Okay, this makes sense. but in that case I will un-export the entire client then.

cmds/upgraded/main.go
```diff
@@ -160,18 +147,24 @@ func main() {
...
+		log.Info().Str("version", current).Msg("node registered successfully")
 	}
 
 	// 3. start zbus server to serve identity interface
 	server, err := zbus.NewRedisServer(module, broker, 1)
 	if err != nil {
 		log.Fatal().Msgf("fail to connect to message broker server: %v\n", err)
 	}
+
 	server.Register(zbus.ObjectID{Name: module, Version: "0.0.1"}, &idMgr)
```
zaibon: I think identityd the Name of the objectID was "manager" and not "identityd"

zaibon: and it should be `idMgr` not `&idMgr`

zaibon (approved):

PR #335, opened 2019-10-17
Update docs for identityd

muhamadazmy:


docs/identity/readmd.md
```diff
@@ -0,0 +1,7 @@
+# Identity Module
```
zaibon: there is a typo in the name of this file. `readmd.md` instead of `readme.md`

zaibon: There is a typo in the name of this file

zaibon (approved):
Only a typo in a file name, other then that LGTM, we can merge

Issue #261, opened 2019-09-19
Prepare release pipeline


zaibon:
To make our release process as smooth as possible we want to create a 3 step release pipeline.
development => testing => production

Development: this is used by developer to develop test new features. It uses the bcdb_mock so the all infrastructure is in our hands if we need to do some change at the bcdb level too. We could have a small farm or we can just keep this development only running with the 0-OS vm from developers

- Testing: once a new version is stable enough it will be promoted to staging. Testing environment needs to map 1 on 1 with production one:
- real BCDB
- farm running in office lab DC
- ...
The testing environment should never be reset cause that would mean we have a release that cannot be pushed to production without breaking it.

On the testing we will run the full integration test suite (still to be created). This test suite will test ALL the feature 0-OS offers. At the time of writing this is very few:
- 4 workloads (network, container, zdb, volume)
- Test the upgraded flow: ensure all the daemon are restarted properly and no workload is affected.

- production: Once all the test are green on staging. we will promote the version to production.
If all goes well, all the node in the grid should progressively get the new version.
At this point we need to have our instrumentation right and monitor the spreading of the new version as well as any abnormal error reports.

This is it for the what, here comes the how:

- [x] prepare a staging farm in office lab (4 nodes enough ?)
- [ ] prepare 3 bcdb
  - [x] bcdb.dev.grid.tf
  - [x] bcdb.test.grid.tf
  - [ ] bcdb.prod.grid.tf
- [ ] develop staging test suite
- [x] write release/promote scripts and automation that goes with it.
  - [x] each commit on master build a new development flist
  - [x] each pre-release creation promote the flist of this version to testing
  - [x] actual release promote the flist to production

zaibon:
release pipeline as been implemented in #346
I"ll close this issue and open a new one just for the staging test suite

PR #346, opened 2019-10-28
Setup deployment pipeline and building

maxux:
Implement #261


.github/workflows/publish-pre-release.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+    - name: Promote flist (testing)
+      if: success()
+      uses: threefoldtech/publish-flist@master
+      with:
+        action: promote
+        user: tf-zos
+        token: ${{ secrets.HUB_JWT }}
+        name: tf-autobuilder/zos-${{ github.sha }}.flist
+        target: zos-${{ github.sha }}.flist
```
zaibon: This is not correct. The target should be `zos:{tag}.flist` so if the release is for the version 0.1.1 then the target should be `zos:0.1.1.flist`

maxux: Yeah, this is a copy/paste from default one, not yet configured, gonna change after forked and tested

.github/workflows/publish-pre-release.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+    - name: Symlink flist (testing)
+      if: success()
+      uses: threefoldtech/publish-flist@master
+      with:
+        action: symlink
+        user: tf-zos
+        token: ${{ secrets.HUB_JWT }}
+        name: zos-${{ github.sha }}.flist
```
zaibon: need to adapt to match the `zos:{tag}.flist` from the previous step

.github/workflows/publish-pre-release.yaml
zaibon: This is not correct. The target should be `zos:{tag}.flist` so if the release is for the version 0.1.1 then the target should be `zos:0.1.1.flist`

zaibon (changes requested):
Flow looks good, but the name of the generated flist is not correct. I added a comment about it. It needs to be fixed in all 3 worklfow

.github/workflows/publish-release.yaml
```diff
@@ -0,0 +1,63 @@
+name: Publishing Files
+on:
+  release:
+    types: [created]
```
zaibon: are we sure this doesn't trigger on creating prerelease too ?

zaibon (changes requested):

.github/workflows/publish-release.yaml
maxux: Probably, I don't know yet, since this is not yet tested.
https://help.github.com/en/github/automating-your-workflow-with-github-actions/events-that-trigger-workflows#release-event-release

muhamadazmy (commented):
This looks awesome, and very clean. But I also agree with @zaibon comments.

zaibon (approved):

PR #348, opened 2019-10-28
Tested pipeline using private repository

maxux:
In order to not adding lot of mess with release into `threefoldtech/zos` repository, all tests were made here. Everything seems good now.


.github/workflows/publish-development.yaml
```diff
@@ -39,22 +39,12 @@ jobs:
         token: ${{ secrets.HUB_JWT }}
         name: zos-${{ github.sha }}.flist
 
-    - name: Promote flist (development)
```
zaibon, line 42: You should keep this block. The symlink needs to happen in the `tf-zos` folder not `tf-autobuilder`

.github/workflows/publish-development.yaml
```diff
@@ -39,22 +39,12 @@ jobs:
...
-        name: tf-autobuilder/zos-${{ github.sha }}.flist
-        target: zos-${{ github.sha }}.flist
-
     - name: Symlink flist (development)
       if: success()
       uses: threefoldtech/publish-flist@master
       with:
         action: symlink
-        user: tf-zos
+        user: tf-autobuilder
```
zaibon:
```suggestion
        user: tf-zos
```

docs/release/README.md
```diff
@@ -0,0 +1,28 @@
+# Releases of Zero-OS
```
zaibon: Add a link to this page from the documentation index of the `/docs` folder.

zaibon (changes requested):

.github/workflows/publish-development.yaml
maxux: That become tricky. You don't want tf-zos folder contains 5000 builds...

zaibon: good point. But we need to do something, cause at the moment our code doesn't check anything else the tf-zos folder.
https://github.com/threefoldtech/zos/blob/ee944be1d79e0b021aa24f90bbc5748befa2aac2/bootstrap/etc/zinit/init/bootstrap.sh#L43-L57

maxux: We can promote directly to `zos:development:latest` which won't be a symlink then, but we don't keep track of the commit id then. We could still do it on tf-autobuilder (so we know the symlink) but not within tf-zos.

zaibon: Maybe development should check into `tf-autobuilder` instead then.

@muhamadazmy any comment on this ?

zaibon: > We can promote directly to `zos:development:latest` which won't be a symlink then, but we don't keep track of the commit id then. We could still do it on tf-autobuilder (so we know the symlink) but not within tf-zos.

The problem is we need the symlink to detect which version we are running. The upgraded code does check what the flist points to and parse the name to extract the version.

maxux: Checking inside `ŧf-autobuilder` seems better IMO then.

muhamadazmy: The name flist `zos:*:latest.flist` must be a symlink, otherwise there is no way to detect the version from  the name and auto-update will never work (basically no way to detect a change in the version). The machine will still boot normally with that flist but basically a new build will never trigger an update.

> Actually even upgraded might misbehave because it won't be able to extract a version number from the name nor the target.

We can change upgraded to just not attempt any updates if no version was detected, but it means development nodes need to restart in case new builds are available. Actually this might be a good idea so your "development/debugging" does not get interrupted when someone else pushes a new change.

zaibon: > We can change upgraded to just not attempt any updates if no version was detected, but it means development nodes need to restart in case new builds are available. Actually this might be a good idea so your "development/debugging" does not get interrupted when someone else pushes a new change.

I do not want to go there. We already have a case where there is not upgrade when a node is not booted from an flist, AKA when this is a local VM from a developer. But we also have some nodes that are long running on real hardware for our devnet. And for these I want to keep the auto update working. This is the first stage of the pipeline that will ensure the upgrade works properly.

docs/release/readme.md
```diff
@@ -0,0 +1,28 @@
+# Releases of Zero-OS
+We use a simple but powerful pipeline release workflow. Building and file distribution are made using GitHub Actions.
+Usable files are available on the [Zero-OS Hub](https://playground.hub.grid.tf/tf-zos).
+
+## Development build
```
zaibon, line 5: Need to update this section to match final implementation

zaibon (changes requested):
workflows LGTM. just need to perfect the documentation and we're good to do

Issue #334, opened 2019-10-17
provisiond: integration test against real BCDB actor


zaibon:
We need to have a full test suite of test against the [provision actor](https://github.com/threefoldtech/jumpscaleX_threebot/blob/development/ThreeBotPackages/threefold/tfgrid_workloads/actors/workload_manager.py)

~~Blocked on https://github.com/threefoldtech/jumpscaleX_threebot/issues/90~~

serboctor:
https://github.com/threefoldtech/jumpscaleX_threebot/issues/90 is fixed.

muhamadazmy:
@zaibon I wanted to start on this, but currently there are some major changes happening on the development_bcdb branch. Should we put this on hold, or start with what we have?

zaibon:
@muhamadazmy no point working on this if the interface of gedis/bcdb will change soon. We need a stable version to test against.

PR #349, opened 2019-10-29
provision: add code to convert workload descript from jsx schema to zos types

zaibon:
part of https://github.com/threefoldtech/zos/issues/334

### DIFF
--- a/docs/provision/provision.md
+++ b/docs/provision/provision.md
@@ -45,7 +45,7 @@ Container provision schema
 ```go
 // Network struct
 type Network struct {
-	NetwokID string `json:"network_id"`
+	NetworkID string `json:"network_id"`
 }
 ```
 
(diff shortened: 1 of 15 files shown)


muhamadazmy (approved):
LGTM. I only don't like that we have too much changes in a supposedly auto generated files. I would say we either extend the schema to support more types and directives (to control how an attribute behave, like `allownull, rawdata, etc...` (all out of scope, i know)

zaibon:
>  I only don't like that we have too much changes in a supposedly auto generated files.

I don't see these as generated in the sense we always generated them during build. As for now schemas is just a tool to help generate a bunch of types from a jsx notation format. Then once they are generated, we push them into the code base and managed them as normal types. So for now I don't mind. If we reach the point where we always generate these types for each build, then we can think about improving stuff.
I just don't think this is worth it to inverst time in this at the moment. jsx schema is still very young and who knows what can change there. But I hear you, more stable schema format would be nice.

Issue #357, opened 2019-10-31
Ensure minting can happen for node running v2


zaibon:
Now that we have started the migration to v2. 
We need to ensure everything is in place so minting can happens also for the node running v2.

- [ ] all required information is sent from the node to BCDB
  - [ ] farmer ID
  - [ ] total resource units
  - [ ] DMI decode dump
- [ ] BCDB directory actors work as expected
  - [ ] resource units are saved
  - [ ] DMI decode dump is saved properly
  - [ ] uptime of the node is computed properly

andhartl:
do we need to re register the nodes? if so we need  to make sure nodes which are already registered have the right amount of periods left for minting attached to them.

zaibon:
> do we need to re register the nodes? if so we need to make sure nodes which are already registered have the right amount of periods left for minting attached to them.

This is all going to be a fresh starts. Since we don't even migrate farms I don't see how we can migrate nodes. 
I can try to see if there is a way to match v1 nodeID to v2. I'll let you know.

zaibon:
DMI decode disks dump is saved properly: https://github.com/threefoldtech/jumpscaleX_threebot/pull/178
all required information is sent from the node to BCDB : https://github.com/threefoldtech/zos/pull/359

zaibon:
@LeeSmet FYI, we have now some node reporting the hardware dump on testnet. You can see a full node detail at: https://explorer.testnet.grid.tf/web/gedis/http/nodes/get?node_id=5rabu25PAnxv5SHgMSwsuqSWQgjhwrr82HJUPQ7LPSHd&proofs=true

That should get you started to update the minting scripts.

zaibon:
Last open item is `uptime of the node is computed properly`
So what needs to happen is, any time this method is called: https://github.com/threefoldtechhttps://github.com/threefoldtech/jumpscaleX_threebot/blob/9ce6118df6c1ca7ff0f39dd6e8ea2110368df5d0/ThreeBotPackages/tfgrid/directory/actors/nodes.py#L237-L251

The actor also write the uptime into influxdb. So during minting we can use influx to compute the total uptime of the month.

The old system code is located there: https://github.com/threefoldtech/jumpscale9_lib/blob/03c1451133d777e5af106fcc6f75c1138bb997f2/JumpscaleLib/servers/grid_capacity/server/handlers/RegisterCapacityHandler.py#L83-L86

we could just copy the bits we need.

PR #359, opened 2019-10-31
capacityd: send the hardware proof to BCDB at boot

zaibon:
when capacityd starts it does a scan fo all the hardware of the node.
this scan is sent to BCDB and this information is used to mint tokens.

part of #357


tools/bcdb_mock/nodes_store.go
```diff
@@ -184,3 +230,30 @@ func (s *nodeStore) Requires(key string, handler http.HandlerFunc) http.HandlerF
 		handler(w, r)
 	}
 }
+
+// hashProof return the hex encoded md5 hash of the json encoded version of p
+func hashProof(p map[string]interface{}) (string, error) {
+
+	// we are trying to have always produce same hash for same content of p
```
muhamadazmy, line 237: It's a good idea, but shouldn't this done recursively on all `map` types ?

muhamadazmy (commented):
I don't get exactly what the hardware proof is used for. I understand you create a hash of the disk information and then store it. Then what?

muhamadazmy (approved):

PR #361, opened 2019-11-02
Fix github actions workflow filesname

maxux:
This changes small things on actions workflow:
- Revert coma on tf-autobuilder (not needed, let's keep dash like all other build)
- Fix the promote filename for development build


muhamadazmy (approved):

zaibon (changes requested):
@maxux I think you forgot to also modify the publish-pre-release.yaml file

zaibon (approved):

PR #366, opened 2019-11-05
Migration routine to reuse disks from v1

### LINKED ISSUE #339: Test migration from node v1 to a v2
sub task of https://github.com/threefoldtech/zos/issues/328

In order to prepare the migration of node from the v1 to v2. We need to identify the potential data migration that will be required.

I think the only things we want to migrate is the data on disks. 
Networking, containers and VM are not something we will migrate.

To do so lets try the following test plan:
- boot a node with a v1 kernel.
- create some storage pool / filesystem
- write some data to it.
- shutdown the node
- update the kernel to v2 and boot
- check that storaged can mount the storage pool properly
- try to mount the storage pool into a container.

muhamadazmy:
This detects if a disk is was used by zos v1
and make sure the disk is cleaned up for v2 usage

Fixes #339

### DIFF
--- a/pkg/storage/filesystem/migrate.go
+++ b/pkg/storage/filesystem/migrate.go
@@ -9,6 +9,10 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
+const (
+	zosV1PoolPrefix = "sp_"
+)
+
 var (
 	zeros [512]byte
 )
@@ -43,13 +47,13 @@ func migrate(ctx context.Context, m DeviceManager, exe executer) (DeviceManager,
 	var destroyMe []Device
 loop:
 	for _, device := range devices {
-		if strings.HasPrefix(device.Label, "sp_") {
+		if strings.HasPrefix(device.Label, zosV1PoolPrefix) {
 			destroyMe = append(destroyMe, device)
 			continue
 		}
 
 		for _, child := range device.Children {
-			if strings.HasPrefix(child.Label, "sp_") {
+			if strings.HasPrefix(child.Label, zosV1PoolPrefix) {
 				destroyMe = append(destroyMe, device)
 				continue loop
 			}
--- a/pkg/storage/filesystem/migrate.go
+++ b/pkg/storage/filesystem/migrate.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"io/ioutil"
 	"strings"
+	"syscall"
 
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog/log"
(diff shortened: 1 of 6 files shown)


pkg/storage/filesystem/migrate.go
```diff
@@ -0,0 +1,76 @@
...
+func migrate(ctx context.Context, m DeviceManager, exe executer) (DeviceManager, error) {
+	devices, err := m.Raw(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "migration failed, couldn't list disks")
+	}
+
+	var destroyMe []Device
+loop:
+	for _, device := range devices {
+		if strings.HasPrefix(device.Label, "sp_") {
```
zaibon: shouldn't "sp_" be a constant ?

pkg/storage/filesystem/migrate.go
```diff
@@ -0,0 +1,76 @@
...
+
+var (
+	zeros [512]byte
+)
+
+func wipe(ctx context.Context, device *Device, exe executer) error {
+	//we first make an msdos parition, because we know it's 512 bytes
+	//creating the msdos partition table also makes us validate that this
+	//is indeed a valid block device
+	if _, err := exe.run(ctx, "parted", "-s", device.Path, "mktable", "msdos"); err != nil {
```
zaibon: I don't get this. Why do we re-created a partition table already  and not just write zeros on the head of the disk to remove any existing filesystems?
Storaged initialization does all this automatically when he detects a disk that has no filesystem

zaibon (changes requested):

pkg/storage/filesystem/migrate.go
muhamadazmy: I am avoiding issues that can get triggered by having an unknown partition table on the disk. May be it's not gpt, so i can't just write gpt size bytes to disk. What if it's a different partition table that is more dynamic, and may be writing to the first 512 bytes won't affect it. 

So instead, I make sure we do msdos partition which is always 512 bytes on the fist sector. And then wipe it out with zeros. 

Also note that, using parted also works as a sort of validation that this is indeed a block device.

I don't see which part of the code you pointing at that storaged does that automatically!

pkg/storage/filesystem/migrate.go
muhamadazmy: Yup

pkg/storage/filesystem/migrate.go
zaibon: alright I get what you do here. So you first create a partition of known size, so you know how much you need to zeroed.
But if this is the size of the partition table you're afraid of. then why don't we just zeroed a big portion of the disk ? few MiB ?

muhamadazmy: I did a quick test to erase a gpt partition by only writing zeros to the 512 bytes header, and it seems it works fine. I will just make sure by writing zeros of 1MB

I will also try to find a way to call both `sync` and `partprope` via systemcalls instead of calling binaries.

zaibon: can't we avoid the need to call sync if we open the file in direct IO mode ?

zaibon (approved):

PR #380, opened 2019-11-12
implement json.Marshaler and json.Unmarshaler for TfgridNodePublicIface1 type

zaibon:


pkg/gedis/types/directory/tfgrid_node_2.go
```diff
@@ -86,3 +88,27 @@ func (e TfgridNodePublicIface1TypeEnum) String() string {
 	}
 	return "UNKNOWN"
 }
+
+// MarshalJSON implements json.Marshaler interface
+func (e TfgridNodePublicIface1TypeEnum) MarshalJSON() ([]byte, error) {
```
muhamadazmy: I don't get why this is done manually, if this type in the schema file is already an Enum, the marshalers should be generated for u automatically!

muhamadazmy (approved):

PR #385, opened 2019-11-15
contd: remove coreX chroot hack

### LINKED ISSUE #381: provisiond: failed to mount volume in container when using coreX
Problem seems to come from the fact that coreX does a chroot after it stars which somehow mask the bind mount  in the container.

This chroot approach was taken cause we needed to have all the libraries and binaries of coreX in the container, so when using coreX, the container is started with `/var/lib/corex` as root filesystem, the flist mounted at `/sandbox` and the a chroot was done into `/sandbox` to have the root of the flist as root of the container.

This all hack seems to have reached his limit. 
The proposed solution for this is to have a static build if coreX that we could easily insert into the container when needed. That would remove the need of starting the container with `/user/lib/corex` as root and avoid the chroot too.

CoreX status build issue: https://github.com/threefoldtech/corex/issues/13

zaibon:
fixes #381
Now that we have coreX build staticly, we can just mount the binary into
the container. We don't need to do the chroot hack

Let's wait we have a new base image with static coreX binary before merging this.

### DIFF
--- a/pkg/container.go
+++ b/pkg/container.go
@@ -20,10 +20,8 @@ type NetworkInfo struct {
 
 // MountInfo defines a mount point
 type MountInfo struct {
-	Source  string   // source of the mount point on the host
-	Target  string   // target of mount inside the container
-	Type    string   // mount type
-	Options []string // mount options
+	Source string // source of the mount point on the host
+	Target string // target of mount inside the container
 }
 
 //Container creation info
--- a/pkg/container/container.go
+++ b/pkg/container/container.go
@@ -81,22 +81,13 @@ func (c *containerModule) Run(ns string, data pkg.Container) (id pkg.ContainerID
 		return id, ErrEmptyRootFS
 	}
 
-	if err := applyStartup(&data, filepath.Join(data.RootFS, ".startup.toml")); err != nil {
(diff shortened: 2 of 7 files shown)


pkg/container.go
```diff
@@ -20,10 +20,8 @@ type NetworkInfo struct {
 
 // MountInfo defines a mount point
 type MountInfo struct {
-	Source  string   // source of the mount point on the host
-	Target  string   // target of mount inside the container
-	Type    string   // mount type
```
muhamadazmy, line 25: Good

pkg/container/container.go
```diff
@@ -178,6 +138,7 @@ func (c *containerModule) Run(ns string, data pkg.Container) (id pkg.ContainerID
 	for _, linxNS := range spec.Linux.Namespaces {
 		log.Info().Msgf("namespace %+v", linxNS.Type)
 	}
+	log.Info().Msgf("mounts %+v", spec.Mounts)
```
muhamadazmy, line 141: I think this should be `Debug()` not info

muhamadazmy (approved):
LGTM, only 1 minor comment

PR #386, opened 2019-11-15
Also send node id v1 to BCDB when registring the node

zaibon:
this is required to be able to match nodes between v1 and v2.

### DIFF
--- a/pkg/gedis/types/directory/tfgrid_node_2.go
+++ b/pkg/gedis/types/directory/tfgrid_node_2.go
@@ -9,6 +9,7 @@ import (
 //TfgridNode2 jsx schema
 type TfgridNode2 struct {
 	NodeID            string                    `json:"node_id"`
+	NodeIDv1          string                    `json:"node_id_v1"`
 	FarmID            uint64                    `json:"farm_id"`
 	OsVersion         string                    `json:"os_version"`
 	Created           schema.Date               `json:"created"`
--- a/pkg/network/v1.go
+++ b/pkg/network/v1.go
@@ -1,0 +1,33 @@
+package network
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/vishvananda/netlink"
+)
+
+// NodeIDv1 returns the node ID as it was calculated in 0-OS v1
+func NodeIDv1() (string, error) {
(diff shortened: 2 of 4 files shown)


pkg/network/v1.go
```diff
@@ -0,0 +1,19 @@
+package network
+
+import (
+	"github.com/vishvananda/netlink"
+	"strings"
+)
+
+// NodeIDv1 returns the node ID as it was calculated in 0-OS v1
+func NodeIDv1() (string, error) {
+	link, err := netlink.LinkByName("zos")
```
muhamadazmy: We never created a zos bridge on v1. and since this is a linux bridge. You have no idea what mac it will get (I don't believe it's even persisted if u deleted and re-added the bridge (on boot for example) . But I think, V1 used the physical network card. (I think this was in the templates)

pkg/gedis/types/directory/tfgrid_node_2.go
```diff
@@ -9,6 +9,7 @@ import (
 //TfgridNode2 jsx schema
 type TfgridNode2 struct {
 	NodeID            string                    `json:"node_id"`
+	NodeIDv1          string                    `json:"node_id_v1"`
```
muhamadazmy, line 12: Once we have this, we will have it forever. Which means, even nodes that never run v1 software, will still have this set. with no real use.

May be, the V1 version should be sent as a separate argument to the register-node method. And bcdb can use it to find a v1 node (and do whatever linking needed), otherwise ignore it. But then the node data will be clean, with no useless garbage.

What do u think?

pkg/network/v1.go
zaibon: indeed very good, thx

pkg/gedis/types/directory/tfgrid_node_2.go
zaibon: Well since there is no way to know from the node if it was running v1 before or not, we will anyhow always send it. So that seems fine for now. Also this change match the one in the model in BCDB.

muhamadazmy: I know that we will always send it, but we don't have to keep it on the node object, is what I am trying to say

PR #388, opened 2019-11-18
Log infrastructure

muhamadazmy:
- Make sure promtail is started with "correct" configurations
- We need to figure out how to inject the node-id in the log labels (still working on that)
- Some file organization for flist building


etc/zinit/promtail.yaml
```diff
@@ -0,0 +1,7 @@
+exec: |
+  promtail 
+  -config.file /etc/promtail/promtail.yaml
+  -client.url http://192.168.122.1:3100/loki/api/v1/push
```
muhamadazmy: This should change to a testnet loki instance

zaibon (approved):

PR #391, opened 2019-11-21
provisiond: sort the reservation when the are received from BCDB 

zaibon:


pkg/gedis/commands_provision.go
```diff
@@ -112,6 +113,19 @@ func (g *Gedis) Poll(nodeID pkg.Identifier, from uint64) ([]*provision.Reservati
 		reservations[i] = r
 	}
 
+	// sorts the primitive in the oder they need to be processed by provisiond
+	// network, zdb, volumes, container
+	sort.Slice(reservations, func(i int, j int) bool {
+		order := map[provision.ReservationType]int{
```
muhamadazmy: The order static map should be taken to a global scope to avoid re-allocation every time the method is called. This method will be called many many times during a single list call. So reallocation of the map is gonna be a problem.

muhamadazmy (changes requested):

muhamadazmy (approved):

PR #393, opened 2019-11-21
rewrite bootstrap in rust

muhamadazmy:
This is a first step to support booting from a minimal image, and have all core-utils and services installed from latest version.

Currently the bootstrap binary does basically the same as the bootstrap.sh but now we have ability to implement a more complex logic for core-utils updates


zaibon (approved):

zaibon:
LGTM. We should also not forget to document this multistage bootstrap in the docs.

PR #395, opened 2019-11-22
don't mount the root zdb flist as read only

### LINKED ISSUE #394: provisiond: ZDB provision is broken
```
{
    "level": "error",
    "error": "failed to create 0-db container: failed to create container: OCI runtime create failed: container_linux.go:345: starting container process caused \"process_linux.go:430: container init caused \\\"rootfs_linux.go:58: mounting \\\\\\\"proc\\\\\\\" to rootfs \\\\\\\"/var/cache/modules/flistd/mountpoint/278a0299f935ebc5a1ed219e45e67274bf84c50283bb53e439e6629dd5bc775b\\\\\\\" at \\\\\\\"/proc\\\\\\\" caused \\\\\\\"mkdir /var/cache/modules/flistd/mountpoint/278a0299f935ebc5a1ed219e45e67274bf84c50283bb53e439e6629dd5bc775b/proc: read-only file system\\\\\\\"\\\"\": unknown",
    "time": "2019-11-22T08:00:03Z",
    "message": "failed to provision reservation 97-1"
}
```
I'm suspecting this is because we do a read-only mount of the flist now for ZDB container.

zaibon:
fixes #394

### DIFF
--- a/pkg/provision/zdb.go
+++ b/pkg/provision/zdb.go
@@ -220,7 +220,10 @@ func createZdbContainer(ctx context.Context, name string, mode pkg.ZDBMode, volu
 	)
 
 	slog.Debug().Str("flist", zdbFlistURL).Msg("mounting flist")
-	rootFS, err := flist.Mount(zdbFlistURL, "", pkg.ReadOnlyMountOptions)
+	rootFS, err := flist.Mount(zdbFlistURL, "", pkg.MountOptions{
+		Limit:    10,
+		ReadOnly: false,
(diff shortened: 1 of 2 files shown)


zaibon:
@muhamadazmy it actually seems we would never be able to start a container using a read only flist. Because we always try to mount some stuff into the container (/proc, /sys, any volume).
Having the root flist prevent these things.

Should I just completly remove the read only option from the pkg.MountOptions struct ?

PR #402, opened 2019-11-25
Work on landing docs

muhamadazmy:


muhamadazmy:
I don't believe that this is any where near enough, but I am honestly don't know exactly the level of depth we need to cover for people to start working zos.

I know in production, people will no have to worry about bcdb infra-structure. but for development and testing, we still missing a lot of documentations to people to start to play around with the OS.

README.md
```diff
@@ -25,12 +25,14 @@ If you want to contribute read the [contribution guideline](CONTRIBUTING.md) and
...
 
 ### Provisioning of workloads
-ZOS does not expose an interface, instead of wait for reservation to happen on a trusted
-source, and once this reservation is available, the node will actually apply it to reality. You can start reading about [provisioning](docs/provision) in this document.
+ZOS does not expose an interface, instead it waits for reservation to happen on a trusted source, and once this reservation is available, the node will actually apply it to reality. You can start reading about [provisioning](docs/provision) in this document.
 
 Currently zos supports two different source:
-- BCDB
-- BCDB mock
+#### BCDB
```
zaibon, line 31: I feel this section is redundant with the `Grid Networks` section just above. It also talk about pretty low level things link kernel parameters. I don't think this belong on home page readme.

zaibon (changes requested):

zaibon:
I think zos repo should try to remain the source of documentation for 0-OS itself. 
When writting this I realize we already talk about `grid networks` on the home page. 
We should probably rephrase all this and point that the first use case of 0-OS is the TF grid and then only explain the different network in place and add some link to the grid documentation and how to provision stuff.

Next to that we could indeed improve the getting started for developers themself. Make sure that the development environment is understood and documented and that anyone can go from code to running 0-OS VM locally and provision stuff on it.

zaibon:
We actually already have some doc for newcomers:
- https://github.com/threefoldtech/zos/blob/master/docs/tfuser/readme.md
- https://github.com/threefoldtech/zos/tree/master/qemu

tfuser documentation is actually a bit outdated, network creation doesn't work the way it is documented in there anymore.

PR #407, opened 2019-11-26
contd: allow to limit the amount of CPU and memory a container can use

### LINKED ISSUE #406: contd: allow to limit the CPU and memory of a container

zaibon:
fixes #406

### DIFF
--- a/pkg/container/opts_test.go
+++ b/pkg/container/opts_test.go
@@ -1,0 +1,64 @@
+package container
+
+import "testing"
+
+func Test_cruToLimit(t *testing.T) {
+	type args struct {
+		cru      uint
+		totalCPU int
+	}
+	tests := []struct {
+		name       string
+		args       args
+		wantQuota  int64
+		wantPeriod uint64
+	}{
+		{
+			name: "1-1",
+			args: args{
+				cru:      1,
+				totalCPU: 1,
+			},
+			wantPeriod: 1000000,
+			wantQuota:  1000000,
+		},
+		{
+			name: "1-4",
+			args: args{
+				cru:      1,
+				totalCPU: 4,
+			},
+			wantPeriod: 250000,
+			wantQuota:  1000000,
+		},
+		{
+			name: "1-6",
+			args: args{
+				cru:      1,
+				totalCPU: 6,
+			},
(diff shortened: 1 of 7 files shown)


zaibon:
The CPU limit has not yet been implemented. the OCI spec allow to control the CPU cgroup of a container. We should agree on the approach we take to map the cgroup notion to the CPU resource unit.

Here is the structure the OCI spec have to controle CPU:
```go
// LinuxCPU for Linux cgroup 'cpu' resource management
type LinuxCPU struct {
	// CPU shares (relative weight (ratio) vs. other cgroups with cpu shares).
	Shares *uint64 `json:"shares,omitempty"`
	// CPU hardcap limit (in usecs). Allowed cpu time in a given period.
	Quota *int64 `json:"quota,omitempty"`
	// CPU period to be used for hardcapping (in usecs).
	Period *uint64 `json:"period,omitempty"`
	// How much time realtime scheduling may use (in usecs).
	RealtimeRuntime *int64 `json:"realtimeRuntime,omitempty"`
	// CPU period to be used for realtime scheduling (in usecs).
	RealtimePeriod *uint64 `json:"realtimePeriod,omitempty"`
	// CPUs to use within the cpuset. Default is to use any CPU available.
	Cpus string `json:"cpus,omitempty"`
	// List of memory nodes in the cpuset. Default is to use any available memory node.
	Mems string `json:"mems,omitempty"`
}
```

muhamadazmy (approved):

pkg/container/opts_test.go
```diff
@@ -0,0 +1,64 @@
...
+			wantPeriod: 1000000,
+			wantQuota:  1000000,
+		},
+		{
+			name: "1-4",
+			args: args{
+				cru:      1,
+				totalCPU: 4,
+			},
+			wantPeriod: 250000,
```
delandtj, line 31: afaict , Period is the reference and the quota is how much of the reference you want to take
hence, numbers need to be switched

delandtj: so period = 1000000, quota = 250000

zaibon: indeed nice catch

PR #419, opened 2019-11-27
deletes qgroups when deleting sub volumes

### LINKED ISSUE #397: subvolume qgroup limit are never destroyed when subvolume is deleted

### LINKED ISSUE #409: storaged: unable to limit requested quota group: No such file or directory
https://explorer.devnet.grid.tf/reservations/122-1
```
"failed to allocate storage: failed to grow limit of volume zdb2db9538c-9d4a-4a9e-8d04-f239c3beb4b0: ERROR: unable to limit requested quota group: No such file or directory\n"
```

zaibon:
fixes #397
fixes #409

The sub-volume delete method has been updated to also delete the
associated qgroup at the same time.

A new method "Maintenance" has been added to the Volume interface
so any left over qgroups from before this commit will be also cleaned
as soon as storaged is updated and restarted

### DIFF
--- a/pkg/storage/filesystem/btrfs.go
+++ b/pkg/storage/filesystem/btrfs.go
@@ -174,6 +174,10 @@ func (p *btrfsPool) mounted(fs *Btrfs) (string, bool) {
 	return "", false
 }
 
+func (p *btrfsPool) ID() int {
+	return 0
+}
+
 func (p *btrfsPool) Name() string {
 	return p.name
 }
@@ -287,6 +291,7 @@ func (p *btrfsPool) Volumes() ([]Volume, error) {
 
 	for _, sub := range subs {
 		volumes = append(volumes, newBtrfsVolume(
+			sub.ID,
 			filepath.Join(mnt, sub.Path),
 			p.utils,
 		))
@@ -296,11 +301,17 @@ func (p *btrfsPool) Volumes() ([]Volume, error) {
 }
 
 func (p *btrfsPool) addVolume(root string) (*btrfsVolume, error) {
-	if err := p.utils.SubvolumeAdd(context.Background(), root); err != nil {
+	ctx := context.Background()
+	if err := p.utils.SubvolumeAdd(ctx, root); err != nil {
 		return nil, err
 	}
 
-	return newBtrfsVolume(root, p.utils), nil
+	volume, err := p.utils.SubvolumeInfo(ctx, root)
+	if err != nil {
+		return nil, err
+	}
+
+	return newBtrfsVolume(volume.ID, root, p.utils), nil
 }
 
 func (p *btrfsPool) AddVolume(name string) (Volume, error) {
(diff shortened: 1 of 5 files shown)


pkg/storage/filesystem/btrfs.go
```diff
@@ -314,7 +325,19 @@ func (p *btrfsPool) AddVolume(name string) (Volume, error) {
 }
 
 func (p *btrfsPool) removeVolume(root string) error {
-	return p.utils.SubvolumeRemove(context.Background(), root)
+	ctx := context.Background()
```
muhamadazmy, line 339: I would actually remove the subvol first before the qgroup. Because if the subvolume deletion failed for any reason, we don't leave a subvolume with no size limit.

pkg/storage/filesystem/btrfs.go
```diff
@@ -393,18 +416,59 @@ func (p *btrfsPool) Reserved() (uint64, error) {
 	return total, nil
 }
 
+func (p *btrfsPool) Maintenance() error {
+	// this method cleans up all the unused
+	// qgroups that could exists on a filesystem
+
+	volumes, err := p.Volumes()
```
muhamadazmy, line 428: Volumes() for both Pool and Subvol actually only lists the first immediate sub volumens. in other words it uses the `btrfs subvol list -o` (note the -o flag) (print only subvolumes below specified path).

Which means if you have sub-volumes of other sub-volumes they won't be cleaned up. Note that while the library supports this, we don't actually expose this via the storage module so there is no "way" that this situation gonna happen.

The question is, should we just do cleaning on the first level of sub-volumes, or make the library code complete by making sure the maintenance do the proper cleaning regardless the depth of the subvol

A quick solution is to implement a SubvolListAll() util and call it directly at this line

zaibon: My full solution assumes we only create a single level of sub volume anyhow. And indeed that's the only thing we support at the moment.
I would be more tempted to remove any code that allow to create sub volume of subvolume and harden the fact that a qgroup is always linked to a single  sub volume, that makes management much easier. What do you say ?

muhamadazmy: Sure we can do that too. I can't think of a scenario where a user would like to create sub-sub-volumes on his volumes.

zaibon: ok, Done in my last commit

pkg/storage/filesystem/btrfs.go
```diff
@@ -314,7 +335,24 @@ func (p *btrfsPool) AddVolume(name string) (Volume, error) {
...
+	if err := p.utils.SubvolumeRemove(ctx, root); err != nil {
+		return err
+	}
+
+	qgroups, err := p.utils.QGroupList(ctx, root)
+	if err != nil {
+		return err
+	}
+
+	for _, qgroup := range qgroups {
```
muhamadazmy: Isn't this gonna delete ALL qgroups for that btrfs filesystem ?

pkg/storage/filesystem/btrfs_ci_test.go
```diff
@@ -301,3 +294,56 @@ func TestBtrfsListCI(t *testing.T) {
...
+	pools, err := fs.List(context.Background(), func(p Pool) bool {
+		return strings.HasPrefix(p.Name(), "test-")
+	})
+	pool := pools[0]
+
+	_, err = pool.Mount()
+	require.NoError(t, err)
+	defer pool.UnMount()
+
+	volume, err := pool.AddVolume("vol1")
```
muhamadazmy, line 327: I think the test here should add 2 (or more) volumes and then delete 1 and then make sure only qgroup for that volume is deleted

zaibon: yep you are right

muhamadazmy (approved):

PR #420, opened 2019-11-27
inverted usage of veth pair

delandtj:
That made reapplying a reservation for a network create new veth pairs
with different names silently but incorrectly work, as the randomnames
always verify existence before creating to make sure its different

### DIFF
--- a/pkg/network/nr/net_resource.go
+++ b/pkg/network/nr/net_resource.go
@@ -364,7 +364,7 @@ func (nr *NetResource) createVethPair() error {
 			Str("veth", vethName).
 			Msg("Create veth pair in net namespace")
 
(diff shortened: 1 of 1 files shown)


pkg/network/nr/net_resource.go
```diff
@@ -364,7 +364,7 @@ func (nr *NetResource) createVethPair() error {
 			Str("veth", vethName).
 			Msg("Create veth pair in net namespace")
 
-		hostVeth, containerVeth, err := ip.SetupVeth(vethName, 1500, hostNS)
+		containerVeth, hostVeth, err := ip.SetupVeth(vethName, 1500, hostNS)
```
zaibon: This is incorrect, `On success, SetupVeth returns (hostVeth, containerVeth, nil)`

zaibon (changes requested):

PR #423, opened 2019-11-28
include nodeID into the Reservation type

zaibon:


muhamadazmy (commented):
I have a general comment on this PR. The node ID just need to be inserted in the Result of the reservation so why it's not just added int he `engine.reply` method ?

zaibon:
Why only in the result ?
Does it make more sense to have it in the reservation itself? A result is anyhow linked to a reservation.

muhamadazmy (approved):

PR #425, opened 2019-11-29
Install/update binaries from tf-zos-bins

muhamadazmy:
This PR in draft for the following reasons
- bootstrap need to also make sure binaries are installed from the tf-zos-bins repository 
- we need to agree on a script hook mechanism to make sure actions are taken when new binaries are installed.

Suggestions:
- An action files in the flist (yaml may be) with a limit set of operations that need to be taken during upgrade (like restart service for example)

Concerns:
i think we will (at some point) will need to have a better tracking of installed packages (files) in a package manager similar fashion. The current code operation might cause some package deletions to remove files that belong to other packages. This is why package managers actually track each package files, to make sure no files are deleted by accident from different packages (for example if both packages install the same file)


pkg/flist/flist.go
```diff
@@ -349,36 +350,30 @@ func random() (string, error) {
...
-					}
-					cErr <- nil
-				} else {
-					if err == nil {
-						time.Sleep(delay)
-						continue
-					}
-					cErr <- nil
-				}
+loop:
```
zaibon: I don't get this change. now, this function will always wait at minimum `delay`. While before it wouldn't, no ? And does the time.After in a loop is called every time or just once ?

zaibon (approved):
LGTM, only one question regarding a change in a function

pkg/flist/flist.go
muhamadazmy: You are right regarding this point, I can re-fix that. But the reason I refactored the method is that It was giving a lot of weird behavior (sometimes just blocks) and it didn't require to run in a go-routine at all. This is why i added a unit-test and refactored the implementation, it's much simpler this way.

For the wait, I can change fix this to not wait on first try.

PR #429, opened 2019-12-03
zdb: correct the way storage allocation is done for 0-db sub volumes

zaibon:


muhamadazmy (commented):
I don't really see any change here except the code organization

zaibon:
@azertyalex 
the main change is this line: https://github.com/threefoldtech/zos/pull/429/files#diff-9a7c5ba1492d9a09f436c35fc6a52705R67
I now use the size of the selected "volume" instead of the pool.
So before this change we would set the size limit to (pool size + reservations size) which is a number bigger then the disk actually have
And now we do set the limit with (volume current size + reservation size) which is always in the boundaries of the disk size and also correct in term of provisioning.

azertyalex:
> @azertyalex 

I think this was a mistake? @zaibon

zaibon:
@azertyalex indeed my bad. Wrong auto completion. I meant to tag @muhamadazmy

PR #431, opened 2019-12-03
provisiond: container decomission handle already deleted container

### LINKED ISSUE #403: failed to decommission reservation
In the logs I see a lot of error like
```
[-] provisiond: {"level":"error","error":"decommissioning of reservation failed: failed to inspect container 112-1: container \"112-1\" in namespace \"nsFuy6CatZGmmtWby13MESN2dYZUJi1ThjPR493w9TyWz1\": not found","time":"2019-11-25T10:53:46Z","message":"failed to decommission reservation 112-1"}
```
I'm not sure of why exactly this happens, but it seems this already shows a problem. The decommission should not fail if the container is already not existing.

zaibon:
Handle the case where a container decomission failed halfway though.
The code now will still try to cleanup network even if the container is
already gone.

fixes #403

### DIFF
--- a/pkg/provision/container.go
+++ b/pkg/provision/container.go
@@ -220,24 +220,25 @@ func containerDecommission(ctx context.Context, reservation *Reservation) error 
 	}
 
 	info, err := container.Inspect(tenantNS, containerID)
-	if err != nil {
-		return errors.Wrapf(err, "failed to inspect container %s", containerID)
-	}
-
-	if err := container.Delete(tenantNS, containerID); err != nil {
-		return errors.Wrapf(err, "failed to delete container %s", containerID)
-	}
-
-	rootFS := info.RootFS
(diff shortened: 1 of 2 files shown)


pkg/provision/container.go
```diff
@@ -217,24 +217,25 @@ func containerDecommission(ctx context.Context, reservation *Reservation) error
 	}
 
 	info, err := container.Inspect(tenantNS, containerID)
-	if err != nil {
-		return errors.Wrapf(err, "failed to inspect container %s", containerID)
-	}
+	if err == nil {
```
muhamadazmy, line 223: Why u didn't keep the `if err != nil` and just change the return to a log, and return nil instead of change the code indentation.

Other that that it's a LGTM

muhamadazmy (approved):

PR #439, opened 2019-12-06
Derive static mac for ndmz and public ns public if

### LINKED ISSUE #316: networkd: DMZ cannot start if farm dhcp doesn't give IPV4 to public interface of DMZ
```
[-] networkd: 10:32AM FTL failed to create DMZ error="public interface in ndmz didn't not received an IP. make sure dhcp is working"
```

LeeSmet:
Derive a static mac address for the public interace in both the ndmz and public namespace (if available), based on the node ID and a static prefix. This ensures these interfaces always have the same MAC.

Fixes #316

### DIFF
--- a/pkg/network/ndmz/ndmz.go
+++ b/pkg/network/ndmz/ndmz.go
@@ -107,7 +107,7 @@ func Create(nodeID pkg.Identifier) error {
 			return err
 		}
 		if !received {
-			return errors.Errorf("public interface in ndmz didn't received an IP. make sure dhcp is working")
+			return errors.Errorf("public interface in ndmz did not received an IP. make sure dhcp is working")
 		}
 		return nil
 	})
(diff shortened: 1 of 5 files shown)


pkg/network/ndmz/ndmz.go
```diff
@@ -104,7 +107,7 @@ func Create() error {
 			return err
 		}
 		if !received {
-			return errors.Errorf("public interface in ndmz didn't not received an IP. make sure dhcp is working")
+			return errors.Errorf("public interface in ndmz didn't received an IP. make sure dhcp is working")
```
zaibon: I prefer to not use abbreviation in error message.
```suggestion
			return errors.Errorf("public interface in ndmz did not received an IP. make sure dhcp is working")
```

zaibon (approved):
LGTM, just one comment regarding an error message. Other then that we're good to merge.

PR #442, opened 2019-12-11
Master zdb allocation

muhamadazmy:
## Changes
- ZDB subvolume has no limit set on it
- Each namespace information is read from the zdb info file on disk 
- On subvolume creation, the system is scanned for the namespace on disk, if found, no suvol allocation is done
- Ensure that the container is running
- Ensure that the namespace is created in zdb


pkg/provision/zdb.go
```diff
@@ -299,13 +207,12 @@ func createZdbContainer(ctx context.Context, name string, mode pkg.ZDBMode, volu
 	bo := backoff.NewExponentialBackOff()
 	bo.MaxInterval = time.Minute * 2
 	if err := backoff.RetryNotify(getIP, bo, nil); err != nil {
```
zaibon:
```suggestion
	if err := backoff.Retry(getIP, bo); err != nil {
```

pkg/provision/zdb.go
```diff
@@ -358,11 +268,14 @@ func zdbDecommission(ctx context.Context, reservation *Reservation) error {
 		return errors.Wrap(err, "failed to decode reservation schema")
 	}
 
-	containerID, ok := zdbMap.Get(nsID)
-	if !ok {
+	allocation, err := storage.Find(reservation.ID)
```
zaibon, line 301: if Find return not found here, does that mean the 0-db namespace is also gone ? Cause do an early return then.

pkg/storage/zdb.go
```diff
@@ -36,116 +70,120 @@ func (s *storageModule) Allocate(diskType pkg.DeviceType, size uint64, mode pkg.
...
-				v = volume
-				break
+			// skip all non-zdb volume
+			if !isZdbVolume(volume) {
+				continue
+			}
+
+			zdb := zdbpool.New(volume.Path())
+
+			reserved, err := zdb.Reserved()
```
zaibon: Here again the way we compute the reserved storage of a disk is not correct. In this case we only check the amount of storage used by 0-db namespaces but we omit the storage reserved by container volumes.

zaibon (changes requested):

pkg/provision/zdb.go
muhamadazmy: It means the namespace directory in 0-db data is gone. I am pretty sure it also means the namespace is gone from zdb itself.

Hence we do an early return once it's 'not found'.

pkg/storage/zdb.go
muhamadazmy: The 0-db subvolume is created with "unlimited" space (size == 0). Hence the only way to compute this is by using the zdbpool reserved.

Yesterday I was considering actually make the btrfs a bit smarter, and that they use the zdbpool to compute the "reserved" space of a subvolume if it is a zdb subvolume. In that case even subvolume creations will take namespaces reservations into account. What do u think? should we do that?

zaibon: yes we should

pkg/storage/storage.go
```diff
@@ -238,6 +239,9 @@ func (s *storageModule) Maintenance() error {
 // CreateFilesystem with the given size in a storage pool.
 func (s *storageModule) CreateFilesystem(name string, size uint64, poolType pkg.DeviceType) (string, error) {
 	log.Info().Msgf("Creating new volume with size %d", size)
+	if strings.HasPrefix(name, "zdb") {
```
zaibon, line 242: :+1:

zaibon (approved):
LGTM

pkg/provision/zdb.go
```diff
@@ -193,6 +193,25 @@ func createZdbContainer(ctx context.Context, allocation pkg.Allocation, mode pkg
 		return errors.Wrap(err, "failed to create container")
 	}
 
+	socket := fmt.Sprintf("unix://%s/zdb.sock", socketDir)
```
zaibon: let's extract this into a function now that you use it in different places.

zaibon: let's extract this into a function now that you use it in different places.

pkg/provision/zdb.go
```diff
@@ -193,6 +193,25 @@ func createZdbContainer(ctx context.Context, allocation pkg.Allocation, mode pkg
...
+	cl := zdb.New(socket)
+	defer cl.Close()
+
+	check := func() error {
```
zaibon: you don't need to wrap it here. just use `cl.Connect` in the `RetryNotify`

zaibon (changes requested):

pkg/provision/zdb.go
muhamadazmy: Yup, true. :stuck_out_tongue_winking_eye:

zaibon (approved):

PR #444, opened 2019-12-13
Fix #232: Ensure storaged does not crash on boot

### LINKED ISSUE #232: storaged : if impossible to mount crashed (ctree failed) btrfs storaged tries indefinitely
If there are problems with the SSD, nothing works, so there is not even a possibility to try to fix the node without going into some form of rescue mode

LeeSmet:
Rather than returning an error while initializing the storage module,
which ultimately crashes the module, maintain separate lists of
(presumably) faulty devices and storagepools. This allows storaged to
finish intializing with all the known working devices and storagpools in
the system. Also expose these lists over the zbus interface, so other
modules have the ability to check if there is faulty hardware and if so,
take action (e.g. notify farmer).

### DIFF
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/google/uuid"
@@ -33,6 +34,8 @@ type storageModule struct {
 	brokenPools   []pkg.BrokenPool
 	devices       filesystem.DeviceManager
 	brokenDevices []pkg.BrokenDevice
+
+	mu sync.RWMutex
 }
 
 // New create a new storage module service
@@ -74,6 +77,8 @@ func New() (pkg.StorageModule, error) {
 
 // Total gives the total amount of storage available for a device type
 func (s *storageModule) Total(kind pkg.DeviceType) (uint64, error) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
 	var total uint64
 
 	for idx := range s.volumes {
@@ -95,11 +100,15 @@ func (s *storageModule) Total(kind pkg.DeviceType) (uint64, error) {
 
 // BrokenPools lists the broken storage pools that have been detected
 func (s *storageModule) BrokenPools() []pkg.BrokenPool {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
 	return s.brokenPools
 }
 
 // BrokenDevices lists the broken devices that have been detected
 func (s *storageModule) BrokenDevices() []pkg.BrokenDevice {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
 	return s.brokenDevices
 }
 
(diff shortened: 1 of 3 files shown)


muhamadazmy (commented):
LGTM. But I am wondering about the behavior of storaged when a disk fails during the normal operation. Not during boot.

For example, if you fail to create a subvolume, if u failed to compute the disk usage (the Total method), etc .. should also storaged move the pool to the broken list, should it try again ?

pkg/storage/storage.go
```diff
@@ -120,11 +134,13 @@ func (s *storageModule) initialize(policy pkg.StoragePolicy) error {
...
+			log.Warn().Msgf("Failed to mount volume %v", volume.Name())
+			continue
 		}
 		log.Debug().Msgf("Mounted volume %s", volume.Name())
+		s.volumes = append(s.volumes, volume)
 	}
-	s.volumes = append(s.volumes, existingPools...)
 
 	// list disks
 	log.Info().Msgf("Finding free disks")
```
zaibon, line 160: Can't you also have some error when listing the devices ?

pkg/storage/storage.go
```diff
@@ -88,6 +92,16 @@ func (s *storageModule) Total(kind pkg.DeviceType) (uint64, error) {
 	return total, nil
 }
 
+// BrokenPools lists the broken storage pools that have been detected
+func (s *storageModule) BrokenPools() []pkg.BrokenPool {
+	return s.brokenPools
```
zaibon, line 105: shouldn't these be protected by a mutex  ?

pkg/storage/storage.go
```diff
@@ -88,6 +92,16 @@ func (s *storageModule) Total(kind pkg.DeviceType) (uint64, error) {
...
+}
+
+// BrokenDevices lists the broken devices that have been detected
+func (s *storageModule) BrokenDevices() []pkg.BrokenDevice {
+	return s.brokenDevices
```
zaibon, line 112: shouldn't these be protected by a mutex  ?

zaibon (changes requested):
The total method needs to be updated to not take in account the amount of storage of the broken devices/pools.

LeeSmet:
> But I am wondering about the behavior of storaged when a disk fails during the normal operation

This should probably be investigated further, though it is not an issue related specifically to storaged imo. After all, once a FS is mounted, storaged is no longer in control, and is not notified if a user gets an error or something. Regardless, I feel that a folow up issue should be created for this since error detection and reporting feels somewhat out of the scope of the original issue.

> The total method needs to be updated to not take in account the amount of storage of the broken devices/pools.

The `Total` method iterates over the elements of the `volumes` field in the storageModule. Since the storagepools which are not mounted are not tracked in this list anymore but in a separate list, this works out of the box.

pkg/storage/storage.go
LeeSmet: I am not intimately familiar with the way lsblk returns it outputs, however I think that it either detects the disk, appending it to the returned disks, or does not detect the disk, and ignores it. Errors returned here are execution errors for the lsblk invocation, and should not be related to any malfunctioning disks afaik. Please let me know if I got this wrong

pkg/storage/storage.go
LeeSmet: I don't think so since the storageModule fields are only written during initialization atm, and access only happens through zbus which should be after initialization. But I'll add it since that can't be defered from just the storage code anyway

zaibon (approved):

PR #447, opened 2019-12-16
Make sure zdb macvlan always have the same mac

muhamadazmy:
This will make SLAAC assign the same IP always
to this interface.

### DIFF
--- a/pkg/network/macvlan/macvlan.go
+++ b/pkg/network/macvlan/macvlan.go
@@ -87,8 +87,14 @@ func Create(name string, master string, netns ns.NetNS) (*netlink.Macvlan, error
 }
 
 // Install configures a macvlan interfaces created with Create method
-func Install(link *netlink.Macvlan, ips []*net.IPNet, routes []*netlink.Route, netns ns.NetNS) error {
+func Install(link *netlink.Macvlan, hw net.HardwareAddr, ips []*net.IPNet, routes []*netlink.Route, netns ns.NetNS) error {
 	return netns.Do(func(_ ns.NetNS) error {
+		if hw != nil && len(hw) != 0 {
+			if err := netlink.LinkSetHardwareAddr(link, hw); err != nil {
+				return err
+			}
+		}
+
 		name := link.Attrs().Name
 
 		for _, ip := range ips {
--- a/pkg/network/public.go
+++ b/pkg/network/public.go
@@ -25,6 +25,7 @@ func CreatePublicNS(iface *types.PubIface, nodeID pkg.Identifier) error {
 		pubNS    ns.NetNS
 		pubIface *netlink.Macvlan
 		err      error
+		mac      net.HardwareAddr
 	)
 
 	if !namespace.Exists(types.PublicNamespace) {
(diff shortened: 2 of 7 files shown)


pkg/network/public.go
```diff
@@ -105,7 +105,7 @@ func CreatePublicNS(iface *types.PubIface, nodeID pkg.Identifier) error {
 		return err
 	}
 
-	if err := macvlan.Install(pubIface, ips, routes, pubNS); err != nil {
+	if err := macvlan.Install(pubIface, nil, ips, routes, pubNS); err != nil {
```
zaibon: now that you can pass the mac address here, I think we can change the line at https://github.com/threefoldtech/zos/blob/0fe0be7b82ffd87af3be00deb8f1d2c72e0261a8/pkg/network/public.go#L47 and instead move set mac here.

pkg/provision/zdb.go
```diff
@@ -133,6 +134,15 @@ func createZdbContainer(ctx context.Context, allocation pkg.Allocation, mode pkg
 		slog = log.With().Str("containerID", string(name)).Logger()
 	)
 
+	// sanity check
+	if len(allocation.VolumeID) < 12 {
+		return fmt.Errorf("invalid volume ID length, expecting zdb<GUID> format, got '%s'", allocation.VolumeID)
+	}
+	hw, err := hex.DecodeString(allocation.VolumeID[len(allocation.VolumeID)-12:])
```
zaibon: I don't understand how the hex of a volume can be a valid mac address ?
What should be use is this function: https://github.com/threefoldtech/zos/blob/5571c15766cf1c81a8c8c7a642fbd77d8b7b36ad/pkg/network/ifaceutil/mac.go#L11

zaibon (changes requested):

muhamadazmy:
I have tested this and works as expected, the zdb now always has an IP associated with it's volume ID which is persisted on the disk.

pkg/provision/zdb.go
muhamadazmy: a mac address is just 6 bytes of data, I use the last 6 bytes of the volume GUID. so I parse the last 12 character of the string representation to 6 bytes, then just use that as MAC address

zaibon: you cannot use any 6 bytes. there are reserved range and a special bit that have to be set. That's why we have a function to do it.

zaibon (approved):

pkg/network/macvlan/macvlan.go
```diff
@@ -87,8 +87,14 @@ func Create(name string, master string, netns ns.NetNS) (*netlink.Macvlan, error
 }
 
 // Install configures a macvlan interfaces created with Create method
-func Install(link *netlink.Macvlan, ips []*net.IPNet, routes []*netlink.Route, netns ns.NetNS) error {
+func Install(link *netlink.Macvlan, hw net.HardwareAddr, ips []*net.IPNet, routes []*netlink.Route, netns ns.NetNS) error {
 	return netns.Do(func(_ ns.NetNS) error {
+		if hw != nil && len(hw) != 0 {
```
delandtj, line 92: `len(hw) != 0` ? what iz dees ? when `hw !=nil` all checks are already done, as hw is already typed.

muhamadazmy: Because HardwareAddr is just an alias to a byte slice. So this check makes sure that not only that the array is allocated, but that it also has elementes (bytes) in it. Idd as u suggest, a len(hw) != 6 is a better check.

PR #458, opened 2019-12-27
Tfuser network

### LINKED ISSUE #453: tfuser: generate correct schema.json
`tfuser gen network` is a dev tool, I admit, but creating the json is woefully inadequate to start testing multiple peering (ipv4-6) and  scenarios. .

`tfuser gen network create --name booo -cidr 172.18.0.0/16` 
That is ok to create the start.

As each node has it's ip addrs registered, we can read that out and fill in the peer addrs with tfuser itself. so:
`tfuser gen --schema booo.json network add-node --node $nodeid --subnet 1 --port 12345 --l3 6` can then have as well the proper peer address as everything filled in. 

`--subnet 1` translates to `172.18.1.0/24` as NRs always have a `/24` subnet
`--l3 [46]` specifies what public peer address to use

`tfuser gen --schema booo.json network add-access --subnet 192.168.123.0/24 --node $nodeid -l3 4` 
adds an access and the routing to that network , plus generates a wireguard config file usable by `wg-quick`.

This FR handles things 2-fold.
- have the building blocks for using `tfuser` to create a network on as well a devnet, as on a BCDB test/prod net
- there are still some things to think about in terms of calls where we haven't yet ventured:
  - ipv4<->ipv6 peering  choice, selecting preference would be ipv6, but if x% nodes are by default only ipv4, we might reconsider
  - I sill have the feeling we haven't considered everything, and generating the json from some 3bot will need to be coded evenly and in sync..

LeeSmet:
Rework the "add-node" subcommand in tfuser, and add an "add-access" subcommand. More specifically, a fully qualified reservation json is now generated after each command invocation. In case of hidden nodes in the network, a public node is selected with ipv4 capability to act as a router for these nodes. For every node, every other directly accessible node is added as a peer, and the subnets which needs to be routed over them is also added.

Add-access allows adding an external subnet, on a machine which is not part of the network (e.g. your laptop). If this command is successful, a wireguard config is printed on stdout. This command automatically generates a set of keys to use, and adds a peer entry on the requested access node (which is part of the network). By default an ipv6 connection is used, unless the `ip4` flag is used.

Closes #453


tools/tfuser/main.go
```diff
@@ -104,34 +104,66 @@ func main() {
...
 									Name:  "subnet",
-									Usage: "subnet to use on this node. The subnet needs to be included in the IP range of the network",
+									Usage: "Subnet to use on this node. The subnet needs to be included in the IP range of the network",
 								},
 								cli.UintFlag{
 									Name:  "port",
+									Usage: "Wireguard port to use. if not specified, tfuser will automatically check BCDB for free fort to use",
+								},
+								cli.UintFlag{
+									Name:  "l3",
```
zaibon: The usage of this flag is wrong. I also do not like `l3` as a flag. It is really cryptic. What is the purpose of this flag ?

tools/tfuser/main.go
```diff
@@ -104,34 +104,66 @@ func main() {
...
 								cli.StringFlag{
 									Name:  "node",
-									Usage: "node ID to remove from the network",
+									Usage: "Node ID to remove from the network",
 								},
 							},
 							Action: cmdsRemoveNode,
 						},
+						{
+							Name:   "graph",
```
zaibon, line 140: That is a nice idea, can we have some example image of what this command produces ?

tools/tfuser/main.go
```diff
@@ -104,34 +104,66 @@ func main() {
...
+							Flags: []cli.Flag{
+								cli.StringFlag{
+									Name:  "node",
+									Usage: "Node ID of the node which will act as an access point",
+								},
+								cli.StringFlag{
+									Name:  "subnet",
+									Usage: "Local subnet which will have access to the network",
+								},
+								cli.BoolFlag{
```
zaibon, line 156: Is this to configure the ip address of the endpoint ? I think the usage value could be more detailed of what this flag does.

pkg/network.go
```diff
@@ -54,11 +54,15 @@ type Network struct {
 	IPRange types.IPNet `json:"ip_range"`
 
 	NetResources []NetResource `json:"net_resources"`
+
+	AccessPoints []AccessPoint `json:"access_points,omitempty"`
```
zaibon: It seems this field if only used by tfuser and has no impact in networkd. So I would prefer this field would not pollute the types used by networkd to actually configure the network. Instead this should be isolated into tfuser code.

pkg/network.go
```diff
@@ -54,11 +54,15 @@ type Network struct {
...
 }
 
 // NetResource is the description of a part of a network local to a specific node
 type NetResource struct {
 	NodeID string `json:"node_id"`
+	// Public endpoints
+	PubEndpoints []net.IP `json:"pub_endpoints"`
```
zaibon: Same remark.

pkg/network.go
```diff
@@ -88,3 +92,14 @@ var (
 	// NetworkSchemaLatestVersion network object latest version
 	NetworkSchemaLatestVersion = NetworkSchemaV1
 )
+
+// AccessPoint info for a network, defining a node which will act as the AP, and
+// the subnet which will be routed through it
+type AccessPoint struct {
```
zaibon: Same remark

tools/tfuser/main.go
zaibon: I also don't see any usage of this flag value anywhere in the code. If it is a left over let's remove it

tools/tfuser/cmds_network.go
```diff
@@ -214,23 +394,311 @@ func isIn(l []uint, i uint) bool {
...
+	for _, pep := range n.PubEndpoints {
+		if pep.To4() != nil {
+			return true
+		}
+	}
+	return false
+}
+
+// This function assumes:
+// - that a hidden node has functioning IPv4
```
zaibon, line 423: Why do we make this assumption ? Also what kind of IPv4 do we expect here ?

tools/tfuser/cmds_network.go
```diff
@@ -214,23 +394,311 @@ func isIn(l []uint, i uint) bool {
...
+				pubNr = nr.NodeID
+				break
+			}
+		}
+		if pubNr == "" {
+			return errors.New("Network has hidden nodes but no public IPv4 node exists")
+		}
+	}
+
+	// We also need to inform nodes how to route the external access subnets.
```
zaibon, line 453: Does this mean the user has no control on which node will be used as "access point" ?

tools/tfuser/cmds_network.go
```diff
@@ -214,23 +394,311 @@ func isIn(l []uint, i uint) bool {
...
+	}
+	return false
+}
+
+// This function assumes:
+// - that a hidden node has functioning IPv4
+// - that a public node ALWAYS has public IPv6, and OPTIONALLY public IPv4
+// - that any public endpoint on any node is actually reachable (i.e. no firewall
+//		blocking incoming traffic)
+func generatePeers(n *pkg.Network) error {
```
zaibon: This function is quite long an complex.
I would love to see tests for it so we can more easily validate it's behavior and lock it down.

zaibon (changes requested):

tools/tfuser/main.go
```diff
@@ -104,34 +104,62 @@ func main() {
...
 						},
+						{
+							Name:   "graph",
+							Usage:  "create a dot graph of the network",
+							Action: cmdGraphNetwork,
+						},
+						{
+							Name:  "add-access",
+							Usage: "Add external access to the network",
+							Flags: []cli.Flag{
```
zaibon, line 147: an extra flag should be added so the use can specify it's own wireguard public key instead of always generating a new one.

zaibon (changes requested):

tools/tfuser/cmds_network.go
```diff
@@ -214,23 +394,311 @@ func isIn(l []uint, i uint) bool {
...
+					for _, subnet := range ipv6OnlySubnets {
+						allowedIPs = append(allowedIPs, subnet)
+						allowedIPs = append(allowedIPs, types.NewIPNet(wgIP(&subnet.IPNet)))
+					}
+				}
+
+				// Endpoint must be IPv4
+				for _, pep := range onr.PubEndpoints {
+					if pep.To4() != nil {
+						endpoint = fmt.Sprintf("%s:%d", pep.String(), nr.WGListenPort)
```
zaibon: I think the wg listen port should be taken from the `onr` object
```suggestion
						endpoint = fmt.Sprintf("%s:%d", pep.String(), onr.WGListenPort)
```

zaibon (changes requested):

tools/tfuser/cmds_network.go
LeeSmet: If the node has IPv6, it should have a public IPv6 endpoint. Since the node has no public endpoints ("hidden") , it can only have a (natted) IPv4. Or it has no internet access, but then it couldn't have registered itself in the first place. Also not sure what you mean by "kind of ipv4"

tools/tfuser/cmds_network.go
LeeSmet: The user has full control over the node (or rather access point) that is used as access point. This is just here to indicate that all other net resources need to be able to reach back to the external node (so the routes need to be configured, like hidden nodes)

tools/tfuser/main.go
LeeSmet: Image has been added in docs/tfuser, and is also linked in the README there

tools/tfuser/main.go
LeeSmet: It sets a subnet for this endpoint, similar to how every net resource also has a subnet. A wireguard IP address is then derived for the endpoint (like it happens for network resources). The subnet is routable from every node in the network. E.G. if you say `--subnet 10.1.0.0/24`, you can set up a bridge on your laptop and attach containers to this bridge, and they will then be reachable from inside the network resources. I'm not really sure how this should be worded, open for suggestions

zaibon: From what I've understood of this flag, what is does is it make the endpoint of the wg connection to be an IPv4 instead of using the public ipv6 address of the access node.
None of this is explained.

tools/tfuser/main.go
```diff
@@ -104,34 +104,66 @@ func main() {
...
+									Name:  "node",
+									Usage: "Node ID of the node which will act as an access point",
+								},
+								cli.StringFlag{
+									Name:  "subnet",
+									Usage: "Local subnet which will have access to the network",
+								},
+								cli.BoolFlag{
+									Name:  "ip4",
+									Usage: "Use an IPv4 connection instead of IPv6",
```
zaibon:
```suggestion
									Usage: "Use an IPv4 connection instead of IPv6 for the wireguard endpoint to the access node",
```

docs/tfuser/readme.md
```diff
@@ -15,113 +15,175 @@ It has 3 main commands:
 Here is a full example how you would use tfuser to create a network schema.
 
 1. create a network from scratch
-`tfuser generate --schema network.json network create --node {nodeID}`
+`tfuser generate network create --name {NETWORK_NAME} --cidr ${CIDR}`
```
zaibon: let's add example with real/working value for each flag.

docs/tfuser/readme.md
```diff
@@ -15,113 +15,175 @@ It has 3 main commands:
...
 
-the `--node` flag point to the node ID of the exit node we have chosen for the network.  
-the value of the `--schema` flag is a path where the provisoin schema will be written by tfuser. If not specified the schema will be printed to stdout.
+`NETWORK_NAME`: Arbitrary name of your network.
+`CIDR`: The prefix which will be used by the network resources in the network.
 
-2. add another node to the network:
-`tfuser generate --schema network.json network add-node --node {nodeID}`
+The schema will be printed on stdout, and should be saved for further actions.
```
zaibon: We should also introduce the --schema flag here since it is then re-used later on in the doc.

docs/tfuser/readme.md
```diff
@@ -15,113 +15,175 @@ It has 3 main commands:
...
 
-the `--node` flag point to the node ID we want to add to the network
+2. add another node to the network:
```
zaibon, line 24: We could add a link to the different explorer so user knows where to look for node IDs

docs/tfuser/readme.md
```diff
@@ -15,113 +15,175 @@ It has 3 main commands:
...
+`WGPORT`: Optional wireguard listening port on the host. If this is not specified, one will be
+	generated automatically. It is the responsibility of the caller to ensure this port
+	is not already in use.
 
 The result of these two command should looks like:
 ```json
 {
   "id": "",
-  "user_id": "E6Zb6cLHYuuXm1KSMeYc1DYNKfknGQqFVKS7tMc3mBvJ",
+  "user_id": "",
```
zaibon, line 68: Is this schema correct ? I count 4 network resource while the documentation only execute `add-node` command once.

docs/tfuser/readme.md
```diff
@@ -15,113 +15,175 @@ It has 3 main commands:
...
   },
-  "signature": "",
-  "created": "",
+  "created": "0001-01-01T00:00:00Z",
   "duration": 0,
+  "to_delete": false
 }
 ```
 
+#### Network provisioning graph
```
zaibon, line 214: There needs to be a paragraph on the `add-access` command too otherwise user are not able to connect to their network.

zaibon (changes requested):

tools/tfuser/main.go
LeeSmet: Oops, I thought this was talking about the subnet flag. My bad.

PR #459, opened 2020-01-02
Monitord

muhamadazmy:


zaibon:
I quickly went over the code and I think we could avoid having a new monitord daemon. 
I think each daemon is already responsible for a part of the UI and the data should come from them.

Everythigng related with system stats (cpu, memory) -> capacityd
Everything related with host (node ID, farm, version) -> identityd
Everything related with network -> networkd
Everything related with disks -> storaged

pkg/provision/engine.go
```diff
@@ -135,6 +167,8 @@ func (e *defaultEngine) decommission(ctx context.Context, r *Reservation) error
 		return errors.Wrap(err, "decommissioning of reservation failed")
 	}
 
+	e.stats[r.Type].Decrement()
```
muhamadazmy: This can/will go below zero (minus value) when we process expired, or to-delete reservations that were never provisioned on the node. There is no way (I know of) to check if the reservation to delete is active, or was never deployed.

zaibon: Well, at line 172, we remove the reservation from the reservation cache.
I think this cache is the most accurate view on what is actually deployed on the node or not. Maybe we could use it to return the number of different workload deployed at any point in time.
This kind of information is something I would also like to send to BCDB at some point.

etc/zinit/networkd.yaml
```diff
@@ -1,3 +1,4 @@
 exec: networkd -broker unix:///var/run/redis.sock -root /var/cache/modules/networkd
+test: ip link show zos
```
zaibon: Internet is the service responsible to create zos,not networkd. So this test will always pass right away and doesn't really bring anything.

pkg/capacity/capacity.go
```diff
@@ -17,11 +17,11 @@ type Capacity struct {
 
 // ResourceOracle is the structure responsible for capacity tracking
 type ResourceOracle struct {
-	storage pkg.StorageModule
+	storage *stubs.StorageModuleStub
```
zaibon, line 20: Why do you replace the interface with the stub type ?

pkg/monitord/host.go
```diff
@@ -0,0 +1,124 @@
...
+	version semver.Version
+	boot    upgrade.Boot
+}
+
+// NewHostMonitor initialize a new host watcher
+func NewHostMonitor(duration time.Duration) (pkg.HostMonitor, error) {
+	if duration == 0 {
+		duration = 2 * time.Second
+	}
+	watcher, err := fsnotify.NewWatcher()
```
zaibon, line 27: Do you still needs this watcher here ? Since the version is not pushed from identityd itself when a new version is available.

pkg/network/namespace/monitor.go
```diff
@@ -0,0 +1,25 @@
...
+func Monitor(ctx context.Context, name string) (chan netlink.AddrUpdate, error) {
+	ns, err := netns.GetFromName(name)
+	if err != nil {
+		return nil, err
+	}
+
+	ch := make(chan netlink.AddrUpdate)
+	if err := netlink.AddrSubscribeAt(ns, ch, ctx.Done()); err != nil {
+		close(ch)
+		ns.Close()
```
zaibon, line 20: are we sure this will always be called ? otherwise we could leak a file descriptor here.

pkg/provision/engine.go
```diff
@@ -40,6 +42,11 @@ type defaultEngine struct {
 // one reservation at a time. On error, the engine will log the error. and
 // continue to next reservation.
 func New(source ReservationSource, rw ReservationReadWriter, fb Feedbacker) Engine {
+	stats := make(map[ReservationType]*Counter)
```
zaibon: This looks like dead code, no ?

qemu/overlay/bin/monitord
```diff
@@ -0,0 +1 @@
+../../../bin/monitord
```
zaibon: this doesn't exist anymore

qemu/overlay/etc/promtail/promtail.yaml
```diff
@@ -1 +0,0 @@
-../../../../etc/promtail/promtail.yaml
```
zaibon, line 1: why is this gone ?

qemu/overlay/etc/zinit/monitord.yaml
```diff
@@ -0,0 +1 @@
+../../../../etc/zinit/monitord.yaml
```
zaibon: this doesn't exist anymore

zaibon (changes requested):

etc/zinit/networkd.yaml
muhamadazmy: Yeah, actually i totally can take this out, i needed it before the refactoring. This just grantee that the zos bridge is created for monitoring, but now network monitoring is done by networkd itself. so I will remove that.

pkg/capacity/capacity.go
muhamadazmy: because zbus events functions unfortunately have different signature between both the module interface, and the generated stubs, unlike the other methods where they generate the same exact interface.

For example
```
// a module that has this method
func (m *Module) Stream(ctx context.Context) <- chan Value{
    
}

// will generate a stub like
func (m *Stub) Stream(ctx context.Context) (<-chan Value, error) {
}
```

Hence it's not possible to use the interface here.

pkg/monitord/host.go
muhamadazmy: yup, this need to go idd.

pkg/network/namespace/monitor.go
muhamadazmy: I checked the AddrSubscribe code, and they close the namespace once the ctx channel is done. Hence it's safe to not close it as long as the subscribe has been successful.

pkg/provision/engine.go
muhamadazmy: Good catch :)

qemu/overlay/bin/monitord
muhamadazmy: Again, very good catch, man you are on fire :fire:

zaibon (approved):

qemu/overlay/etc/promtail/promtail.yaml
muhamadazmy: This is installed from the binaries, it's not part of zos core anymore

qemu/overlay/etc/zinit/monitord.yaml
muhamadazmy: removed

PR #460, opened 2020-01-02
Network changes

### LINKED ISSUE #456: 'sometimes' reprovisioning an NR fails with a nic veth endpoint in ndmz alraedy existing
Wile I suspect a bug in 4.14 and the fact that there have been big changes in 4.18->4.19 to address issues with removing network interfaces that have been sent into namespaces (read: __when__ are we going to have 4.19.or 5.4?) , We still should not rely on the kernel to tear everything down.

Instead, before removing an NS, we should remove the interfaces __in__ the ns before deleting it.

OTOH... if 4.19 (or 5.4) solves that problem, I can live with it.

### LINKED ISSUE #457: networkd: if an error occurs in the creation of an NR, there is still some lingering leftovers in the system
We should make sure we have a fully destructed NR when something fails in the middle

### LINKED ISSUE #461: Remove catch all rule in NR routing table

### LINKED ISSUE #462: Remove veth ifaces from network
In favour of macvlan interfaces

LeeSmet:
Might fix #456
Might fix #457 

Closes #461
Closes #462

### DIFF
--- a/pkg/network/ndmz/ndmz.go
+++ b/pkg/network/ndmz/ndmz.go
@@ -34,10 +34,9 @@ const (
 	BridgeNDMZ = "br-ndmz"
 	netNSNDMZ  = "ndmz"
 
-	vethGWSide = "ipv4-rt"
-	vethBrSide = "to-gw"
-
 	ndmzNsMACDerivationSuffix = "-ndmz"
+
+	publicIfaceName = "public"
 )
 
 //Create create the NDMZ network namespace and configure its default routes and addresses
@@ -222,7 +221,7 @@ func applyFirewall() error {
 	return nil
 }
 
-// AttachNR links a network resource to the DMZ
+// AttachNR links a network resource to the NDMZ
 func AttachNR(networkID string, nr *nr.NetResource) error {
 	nrNSName, err := nr.Namespace()
 	if err != nil {
@@ -234,35 +233,26 @@ func AttachNR(networkID string, nr *nr.NetResource) error {
 		return err
 	}
 
-	vethNR := "public"
-	vethDMZ := fmt.Sprintf("n-%s", nr.ID())
-
-	if !ifaceutil.Exists(vethDMZ, nil) || !ifaceutil.Exists(vethNR, nrNS) {
-		log.Debug().
-			Str("nr side", vethNR).
-			Str("dmz side", vethDMZ).
-			Msg("create veth pair to connect network resource and ndmz")
-
-		_ = ifaceutil.Delete(vethDMZ, nil)
-		_ = ifaceutil.Delete(vethNR, nrNS)
-
-		if _, _, err = ip.SetupVethWithName(vethDMZ, vethNR, 1500, nrNS); err != nil {
-			return errors.Wrap(err, "failed to create veth pair for to connect network resource and ndmz")
-		}
+	if _, err = macvlan.Create(publicIfaceName, BridgeNDMZ, nrNS); err != nil {
+		return err
 	}
 
-	err = nrNS.Do(func(_ ns.NetNS) error {
+	return nrNS.Do(func(_ ns.NetNS) error {
 		addr, err := allocateIPv4(networkID)
 		if err != nil {
-			return errors.Wrap(err, "ip allocation for network resource veth error")
+			return errors.Wrap(err, "ip allocation for network resource")
 		}
 
-		lvethNR, err := netlink.LinkByName((vethNR))
+		pubIface, err := netlink.LinkByName(publicIfaceName)
 		if err != nil {
 			return err
 		}
 
-		if err := netlink.AddrAdd(lvethNR, &netlink.Addr{IPNet: addr}); err != nil && !os.IsExist(err) {
+		if err := netlink.AddrAdd(pubIface, &netlink.Addr{IPNet: addr}); err != nil && !os.IsExist(err) {
+			return err
+		}
+
+		if err = netlink.LinkSetUp(pubIface); err != nil {
 			return err
 		}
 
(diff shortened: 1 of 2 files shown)


pkg/network/nr/net_resource.go
```diff
@@ -249,19 +249,21 @@ func (nr *NetResource) Delete() error {
 	return nil
 }
 
-func (nr *NetResource) routes() ([]*netlink.Route, error) {
-	routes := make([]*netlink.Route, 0, len(nr.resource.Peers)+1)
+func (nr *NetResource) routes() ([]netlink.Route, error) {
```
zaibon: I do not understand this change. The commit message seems to only talk about pointer. 
But there is actually a logic change here. 
The gateway of the route is now different. Is this change valid ?

pkg/network/ndmz/ndmz.go
```diff
@@ -59,6 +59,14 @@ func Create(nodeID pkg.Identifier) error {
 		if _, err := sysctl.Sysctl("net.ipv6.conf.all.forwarding", "1"); err != nil {
 			return errors.Wrapf(err, "failed to enable ipv6 forwarding in gateway namespace")
 		}
+		lo, err := netlink.LinkByName("lo")
```
zaibon: there is a function in ifaceutil package to do this: https://github.com/threefoldtech/zos/blob/70d05f1a5d2b0af1cff3fecb1d43d0a8502fb26a/pkg/network/ifaceutil/interface.go#L120-L131

pkg/network/nr/net_resource.go
```diff
@@ -249,24 +250,49 @@ func (nr *NetResource) Delete() error {
 	return nil
 }
 
-func (nr *NetResource) routes() ([]netlink.Route, error) {
-	routes := make([]netlink.Route, 0, len(nr.resource.Peers)+1)
+// func (nr *NetResource) routes() ([]netlink.Route, error) {
```
zaibon: no point keeping old code in comment, delete it please.

zaibon (changes requested):

zaibon:
@delandtj @LeeSmet can you also explain the reason behing using macvlan instead of veth pair in network resource now.

pkg/network/ndmz/ndmz.go
```diff
@@ -226,35 +233,26 @@ func AttachNR(networkID string, nr *nr.NetResource) error {
...
-			Str("dmz side", vethDMZ).
-			Msg("create veth pair to connect network resource and ndmz")
-
-		_ = ifaceutil.Delete(vethDMZ, nil)
-		_ = ifaceutil.Delete(vethNR, nrNS)
-
-		if _, _, err = ip.SetupVethWithName(vethDMZ, vethNR, 1500, nrNS); err != nil {
-			return errors.Wrap(err, "failed to create veth pair for to connect network resource and ndmz")
-		}
+	if _, err = macvlan.Create(publicIfaceName, BridgeNDMZ, nrNS); err != nil {
```
zaibon: How do you handle the case where the interface already exists ?

pkg/network/nr/net_resource.go
```diff
@@ -308,17 +316,24 @@ func (nr *NetResource) createNetNS() error {
...
+	})
 	return nil
 }
 
-// createVethPair creates a veth pair inside the namespace of the
-// network resource and sends one side back to the host namespace
-// the host side of the veth pair is attach to the bridge of the network resource
-func (nr *NetResource) createVethPair() error {
-
+// We're no setting up a veth pair any more, instead, we use a macvlan
```
zaibon: Comment is not correct. 
Golang comment always starts with the name of the function e.g: `attachToNRBridge attaches a ..`.
Also don't describe it with thing like "any more". The code doesn't show any history so a reader have no clue what was here before. Instead describe why it does what it does.

zaibon (changes requested):

pkg/network/nr/net_resource.go
LeeSmet: yes there were 2 changes in this code, the first fixed a bug related to how golang handles loops over slices and pointers, the second changed some of the logic to properly set up all routes as provided in the reservation and remove the catch all rule

pkg/network/ndmz/ndmz.go
LeeSmet: This function is only called after creating a new network namespace and created the `public` interface inside the NS. So the only way to have an iface already exists here would be to call this function multiple times, which indicates a logic error on our side I think

LeeSmet: This function is only called after creating a new network namespace and created the public interface inside the NS. So the only way to have an iface already exists here would be to call this function multiple times, which indicates a logic error on our side I think.

pkg/network/nr/net_resource.go
zaibon: :+1:

pkg/network/ndmz/ndmz.go
zaibon: I don't think this is right.
A network resource can be already created when a new network reservation comes in. 
The current code deals with this gracefully and only update the wireguard configuration with the information of the new reservation.

Which mean a network resource can exists and this code be called anyone. All the methods related to network resource creation needs to be idempotent

zaibon: In this case I wonder if a `err != nil && !os.IsExists(err)` wouldn't be enough. But need to doulbe check what exactly happen in the `macvlan.Create` method

LeeSmet: I guess this is the code which would error:

https://github.com/threefoldtech/zos/blob/3b2f79066da4226e7df3bc5846656de599b00a91/pkg/network/macvlan/macvlan.go#L47-L49

which returns a custom error. So if we want to check for an os.IsExists error in the calling code, we also need to return that in the macvlan create code

zaibon: Actually no, this line will never return a exist error cause it is created random name and then only it is renamed from inside the network namespace.
So what you propose won't work.

I think we have to do the same as what the previous code was doing which is manually check if the interface already exists with `!ifaceutil.Exists(publicIfaceName, nrNS)`

zaibon (approved):

PR #463, opened 2020-01-04
networkd: watch address update on host network interfaces

### LINKED ISSUE #455: networkd: registers node interface config only once
when a IPv6 Router advertisement daemon is slow to react (+10 seconds) , networkd registers an incomplete network config.

networkd should reissue it's config if interfaces have received some IP address, or when the network has changed ...

zaibon:
fixes #455

The idea here is to allow networkd to watch a list of nic for address modification.
Any time an address is added or remove from a watched interfaces, networkd send the full list of interface info updated to BCDB.

Someting to consider here is that I have no clue how spamy subscribing to netlink is. So we migh have to add some kind of rate limiting or something to now DOS BCDB.

### DIFF
--- a/cmds/networkd/addr_watcher.go
+++ b/cmds/networkd/addr_watcher.go
@@ -1,0 +1,152 @@
+package main
+
+import (
+	"context"
+
+	"github.com/rs/zerolog/log"
+
+	"time"
+
+	"github.com/cenkalti/backoff/v3"
+	"github.com/threefoldtech/zos/pkg"
+	"github.com/threefoldtech/zos/pkg/network"
+	"github.com/threefoldtech/zos/pkg/network/ifaceutil"
+	"github.com/threefoldtech/zos/pkg/network/types"
+	"github.com/vishvananda/netlink"
+)
+
+type WatchedLinks struct {
+	linkNames map[string]struct{}
+	bcdb      network.TNoDB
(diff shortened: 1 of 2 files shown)


cmds/networkd/addr_watcher.go
```diff
@@ -0,0 +1,150 @@
...
+	var output []types.IfaceInfo
+
+	links, err := netlink.LinkList()
+	if err != nil {
+		log.Error().Err(err).Msgf("failed to list interfaces")
+		return nil, err
+	}
+
+	for _, link := range ifaceutil.LinkFilter(links, []string{"device", "bridge"}) {
+		//TODO: why bringing the interface up? shouldn't we just check if it's up or not?
```
delandtj: a NIC of which the MII has no handshake detected, doesn't matter if it's up or down, so we bring them up, in case there is some IPv6 RA on that link.

cmds/networkd/addr_watcher.go
```diff
@@ -0,0 +1,150 @@
...
+		if err := netlink.LinkSetUp(link); err != nil {
```
delandtj: but not sure if it waits for the syscall to complete (i.e. MII detected), or it 'just' sends the syscall, and continues it's merry way. If the latter, we should wait timeout time to see if there is effectively a link.

PR #470, opened 2020-01-08
re-add ipv6 support for network resources

zaibon:


pkg/network/ndmz/ndmz.go
```diff
@@ -111,23 +147,43 @@ func Delete() error {
...
 
-		var (
-			master netlink.Link
-			err    error
-		)
-
+		// find which interface to use as master for the macvlan
 		if namespace.Exists(types.PublicNamespace) {
-			master, err = getPublicIface()
+			pubNS, err := namespace.GetByName(types.PublicNamespace)
```
LeeSmet, line 155: this err is not checked

LeeSmet (changes requested):
LGTM except 1 small remark

pkg/network/ndmz/ndmz.go
delandtj: well yea.. it Exists(), so could as well be _

LeeSmet: Not sure about that, the implementation of GetByName seems to also open a file, which could error even if it exists, as far as I know.

PR #474, opened 2020-01-09
add 'delete' command to tfuser

zaibon:


delandtj:
tested and works, 
only ... `--seed=` is missing, so anyone can delete my stuff :'(
(not that I don't trust you guys)

PR #477, opened 2020-01-10
Install binaries even if u running from overlay

### LINKED ISSUE #454: identityd: if booting from overlay make sure to install binaries from tf-zos-bins
Currently when the system is booted from overlay (for development) identityd makes sure it doesn't check for updates to not override the binaries coming from the overlay. But identityd still need to install the other packages from tf-zos-bins to make sure the system is fully operations.

# Workaround.
I usually trick the system that is booted from the latest flist, so it will install all packages from tf-zos-bins normally. But the problem with this approach if suddenly there is a build/update available on the hub, it will override the system binaries. 

```bash
echo "tf-zos/zos:development:latest.flist" > /tmp/flist.name
wget -O /tmp/flist.info https://hub.grid.tf/api/flist/tf-zos/zos:development:latest.flist/light
pkill identityd
```

muhamadazmy:
Fixes #454

### DIFF
--- a/cmds/identityd/main.go
+++ b/cmds/identityd/main.go
@@ -189,6 +189,23 @@ func main() {
 		}
 	}()
 
+	zinit, err := zinit.New(zinitSocket)
+	if err != nil {
+		log.Fatal().Err(err).Msg("failed to connect to zinit")
+	}
+
+	// with volume allocation is set to nil this flister can be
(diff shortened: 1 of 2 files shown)


muhamadazmy:
Don't merge atm, i think there is a bug somewhere

PR #481, opened 2020-01-10
Encrypt zdb password

### LINKED ISSUE #396: zdb provision: password for zdb namespace should be encrypted in the reseravtion
At the moment the password for a 0-DB namespace reservation is in clear text into the reservation.
This needs to change. The password can be encrypted with the public key of the node.

muhamadazmy:
Fixes #396


zaibon (approved):
LGTM. only question is why did you choose to use base58 to encode the password after encryption ?
In the network reservation we do use hex to encode the password, see: https://github.com/threefoldtech/zos/blob/master/tools/tfuser/cmds_network.go#L150

I think we should be consistent in the way we encrypt/encode things

muhamadazmy:
@zaibon node-id uses base58 to encrypt the node-id from a public key, I just used the same encoding.

tools/tfuser/main.go
```diff
@@ -253,6 +253,10 @@ func main() {
 									Name:  "password, p",
 									Usage: "optional password",
 								},
+								cli.StringFlag{
```
zaibon, line 256: I don't think we should add this flag here.
The node ID is already specify in the provision command.
So I think the encryption needs to happen during provision command. 
- load json
- if it's a zdb reservation, load the zdb data 
- encrypt the password using the node ID
- sign the full reservation with encrypted password
- send the signed reservation to BCDB

zaibon (changes requested):

tools/tfuser/main.go
muhamadazmy: Yeah, I considered that first but then i realized
- The password will be unencrypted in the schema file
- The provision need to know the type of the reservation and have special handling for the zdb case.

PR #487, opened 2020-01-20
VM daemon

muhamadazmy:


etc/zinit/vmd.yaml
```diff
@@ -0,0 +1,3 @@
+exec: vmd -broker unix:///var/run/redis.sock
+after:
+  - node-ready
```
zaibon: shouldn't it also depend on `networkd` and `storaged` ?

go.mod
```diff
@@ -36,31 +45,35 @@ require (
...
 	github.com/tcnksm/go-input v0.0.0-20180404061846-548a7d7a8ee8
 	github.com/threefoldtech/zbus v0.1.2
 	github.com/urfave/cli v1.22.1
 	github.com/vishvananda/netlink v1.0.0
-	github.com/vishvananda/netns v0.0.0-20180720170159-13995c7128cc
+	github.com/vishvananda/netns v0.0.0-20191106174202-0a2b9b5464df
 	github.com/vmihailenco/msgpack v4.0.4+incompatible // indirect
 	github.com/whs/nacl-sealed-box v0.0.0-20180930164530-92b9ba845d8d
 	go.etcd.io/bbolt v1.3.3 // indirect
+	go.mongodb.org/mongo-driver v1.2.1 // indirect
```
zaibon, line 67: is this a indirect dependency from the firecracker-sdk ?

pkg/network.go
```diff
@@ -41,6 +41,14 @@ type Networker interface {
 	// hw is an optional hardware address that will be set on the new interface
 	ZDBPrepare(hw net.HardwareAddr) (string, error)
 
+	// SetupTap sets up a tap device is the namespace for the networkID. It is hooked
```
zaibon:
```suggestion
	// SetupTap sets up a tap device in the network namespace of the networkID. It is hooked
```

pkg/network.go
```diff
@@ -41,6 +41,14 @@ type Networker interface {
...
+	// to the network bridge. The name of the tap interface is returned
```
zaibon, line 45:
```suggestion
	// to the bridge of the network resource. The name of the tap interface is returned
```

pkg/network/networker.go
```diff
@@ -226,6 +230,81 @@ func (n networker) ZDBPrepare(hw net.HardwareAddr) (string, error) {
...
+		return "", errors.Wrap(err, "failed to load network resource")
+	}
+
+	nsName, err := netRes.Namespace()
+	if err != nil {
+		return "", errors.Wrap(err, "could not get network namespace name")
+	}
+	log.Info().Str("ns name", nsName).Msg("Network namespace")
+
+	ns, err := namespace.GetByName(nsName)
```
zaibon: you need to defer the `Close` function on the `ns` object.
```go
defer ns.Close()
```

pkg/network/networker.go
```diff
@@ -226,6 +230,81 @@ func (n networker) ZDBPrepare(hw net.HardwareAddr) (string, error) {
...
+		return errors.Wrap(err, "failed to load network resource")
+	}
+
+	nsName, err := netRes.Namespace()
+	if err != nil {
+		return errors.Wrap(err, "could not get network namespace name")
+	}
+	log.Info().Str("ns name", nsName).Msg("Network namespace")
+
+	ns, err := namespace.GetByName(nsName)
```
zaibon: you need to defer the `Close` function on the `ns` object.
```go
defer ns.Close()
```

pkg/network/tuntap/tap.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"fmt"
+
+	"github.com/containernetworking/plugins/pkg/ns"
+	"github.com/containernetworking/plugins/pkg/utils/sysctl"
+	"github.com/pkg/errors"
+	"github.com/vishvananda/netlink"
+)
+
+const disableIPv6Template = "net.ipv6.conf.%s.disable_ipv6"
```
zaibon, line 11: I've seen this else where in the code. Let's extract it into the `ifaceutl` package so we can reuse it instead of copying the code.

pkg/vm.go
```diff
@@ -0,0 +1,113 @@
...
+// VMNetworkInfo structure
+type VMNetworkInfo struct {
+	// Namespace name
+	Namespace string
+	// Tap device name
+	Tap string
+	// Mac address of the device
+	MAC net.HardwareAddr
+	// Address of the device in the form of cidr
+	AddressCIDR string
```
zaibon: why not using a proper type for it, `net.IPNet` ?

pkg/vm.go
```diff
@@ -0,0 +1,113 @@
...
+	// Gateway gateway address
+	GatewayIP string
```
zaibon: why not using a proper type for it, `net.IP` ?

pkg/vm.go
```diff
@@ -0,0 +1,113 @@
...
+	// Nameservers dns servers
+	Nameservers []string
```
zaibon: why not using a proper type for it, `[]net.IP` ?

pkg/vm.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// Disk specifies vm disk params
+type Disk struct {
```
zaibon: Can we rename this type to `VMDisk` or `VDisk` ? Something that express more that it is only related to VMs ?

pkg/vm.go
```diff
@@ -0,0 +1,113 @@
...
+	Name string
+	// CPU is number of cores assigned to the VM
+	CPU uint8
+	// Memory size in Mib
+	Memory int64
+	// Network is network info
+	Network VMNetworkInfo
+	// an allocated storage for the vm operation
+	// where files/virtual disks can be allocated
+	Storage string
```
zaibon: I'm not sure I understand what this fields represent. Is it a path where the vm disk file will be located ?

pkg/vm.go
```diff
@@ -0,0 +1,113 @@
...
+
+	if missing(vm.RootImage) {
+		return fmt.Errorf("root-image is required")
+	}
+
+	if missing(vm.Storage) {
+		return fmt.Errorf("storage is required")
+	}
+
+	if vm.Memory <= 0 {
```
zaibon: We should also have a minimum. What is the required amount of memory for the VM to boot ?

pkg/vm/manager.go
```diff
@@ -0,0 +1,404 @@
...
+}
+
+var (
+	_ pkg.VMModule = (*vmModuleImpl)(nil)
+
+	errScanFound = fmt.Errorf("found")
+)
+
+// NewVMModule creates a new instance of vm manager
+func NewVMModule(root string) (pkg.VMModule, error) {
```
zaibon, line 47: It is better to return a actual type instead of an interface.
```suggestion
func NewVMModule(root string) (*vmModuleImpl, error) {
```

pkg/vm/manager.go
```diff
@@ -0,0 +1,404 @@
...
+	if err := os.MkdirAll(FCSockDir, 0755); err != nil {
+		return nil, err
+	}
+
+	return &vmModuleImpl{
+		root: root,
+	}, nil
+}
+
+func (m *vmModuleImpl) makeDisk(name string, size int64) error {
```
zaibon: This needs to be moved into `storaged`

zaibon (changes requested):

etc/zinit/vmd.yaml
muhamadazmy: yes, this should be set to
```yaml
- boot
- networkd
```

go.mod
muhamadazmy: I assume so, i will run `mod tidy` and see if it stays

**EDIT**: `go mod tidy` did not remove this :man_shrugging:

pkg/vm.go
muhamadazmy: Yeah, this will be gone after the changes to `storaged` but now it's basically the location where I do `truncate` of vm disks.

pkg/vm.go
muhamadazmy: depends on your image, but I think a 512M should be enough

pkg/vm/manager.go
muhamadazmy: I understand, but this is a private type. To make this possible I will have to make it a public type . like `VMModule` but then, it will be possible to developers to create an instance of this type directly without calling the `NewVMModule` method!

I am not sure which is worse actually.

pkg/vm/manager.go
muhamadazmy: Yup, we will do that, I just need to finalize few things in vmd (trying to run the k3s image) before I move to storaged

pkg/vm.go
zaibon: Let's take the k3os image as reference for the moment since this is going to be the first use case.

pkg/vm/manager.go
zaibon: Indeed you're right, being a private type if is better to just return the interface then. My bad.

pkg/network/tuntap/tap.go
```diff
@@ -29,10 +29,14 @@ func CreateTap(name string, master string, netns ns.NetNS) (*netlink.Tuntap, err
 		Mode: netlink.TUNTAP_MODE_TAP,
 	}
 
-	if err := netlink.LinkAdd(tap); err != nil {
+	if err = netlink.LinkAdd(tap); err != nil {
 		return nil, errors.Wrap(err, "could not add tap device")
 	}
 
+	if err = netlink.LinkSetNsFd(tap, int(netns.Fd())); err != nil {
```
zaibon: Do you need this even if you specify the namespace line 26 ?

pkg/provision/kubernetes.go
```diff
@@ -0,0 +1,311 @@
...
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/threefoldtech/zos/pkg"
+	"github.com/threefoldtech/zos/pkg/stubs"
+)
+
+// Kubernetes reservation data
+type Kubernetes struct {
+	// Size of the vm, this defines the amount of vCpu, memory, and the disk size
```
zaibon, line 23: we should add the detail of the sizes here or a link to the document describing the sizes.

pkg/provision/kubernetes.go
```diff
@@ -0,0 +1,311 @@
...
+	Size uint8 `json:"size"`
+
+	// NetworkID of the network namepsace in which to run the VM. The network
+	// must be provisioned previously.
+	NetworkID pkg.NetID `json:"network_id"`
```
zaibon, line 29: Can't we reuse the same type used in container reservation to configure the network ?
So we are consistent between both ?

pkg/provision/kubernetes.go
```diff
@@ -0,0 +1,311 @@
...
+	SSHKeys []string `json:"ssh_keys"`
+
+	PlainClusterSecret string `json:"-"`
+}
+
+// TODO: move flist
+const k3osFlistURL = "https://hub.grid.tf/lee/k3os.flist"
+
+func kubernetesProvision(ctx context.Context, reservation *Reservation) (interface{}, error) {
+	return nil, kubernetesProvisionImpl(ctx, reservation)
```
zaibon: The result of the deployment doesn't return any information ?
How does the user can reach its cluster with kubectl once it is deploy then ?

pkg/provision/kubernetes.go
```diff
@@ -0,0 +1,311 @@
...
+}
+
+func kubernetesProvisionImpl(ctx context.Context, reservation *Reservation) error {
```
zaibon: This function doesn't really bring anything, I would just move the code into `kubernetesProvision` function and delete this one.

pkg/storage/disk.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	file, err := os.Create(path)
+	if err != nil {
+		return "", err
+	}
+
+	defer file.Close()
+
+	return path, syscall.Fallocate(int(file.Fd()), 0, 0, size*1024*1024)
```
zaibon: 1024*1024 can be a constant.

pkg/storage/filesystem/btrfs.go
```diff
@@ -516,10 +516,19 @@ func (v *btrfsVolume) Usage() (usage Usage, err error) {
 		return
 	}
 
+	size := group.MaxRfer
+	if size == 0 {
+		// in case no limit is set on the subvolume, we assume
+		// it's size is the size of the files on that volumes
+		size, err = FilesUsage(v.Path())
```
zaibon, line 523: Is this a viable solution for a multi TB disk ? I'm afraid this will be quite slow ?
@delandtj ?

pkg/vm/manager.go
zaibon: should be deleted now ?

pkg/vm/strategy.go
```diff
@@ -0,0 +1,104 @@
...
+	"context"
+	"os"
+	"path/filepath"
+	"syscall"
+
+	"github.com/firecracker-microvm/firecracker-go-sdk"
+	"github.com/pkg/errors"
+)
+
+// TODO: this is a dead code file. Was used when we use the firecracker-go-sdk to start
```
zaibon: You want to keep this file in case they fix the bug we have with the sdk and we change to use it again ?

zaibon (dismissed):

pkg/vm/manager.go
muhamadazmy: yeah, it's not used anymore, will remove the dead code

pkg/flist/flist.go
```diff
@@ -113,8 +113,21 @@ func New(root string, storage pkg.VolumeAllocater) pkg.Flister {
 	return newFlister(root, storage, cmd(exec.Command))
 }
 
+// NamedMount implements the Flister.NamedMount interface
+func (f *flistModule) NamedMount(name, url, storage string, opts pkg.MountOptions) (string, error) {
+	return f.mount(name, url, storage, opts)
```
muhamadazmy, line 118: There should be a check here to make sure this mount point is already not used.

muhamadazmy: Also since user can provide a name, we need to make sure that the name is a valid "filesystem" name and that it doesn't container relative path directories "like ../" otherwise it's prone to attacks by carefully crafting a mount endpoint to manipulate parts of the system

pkg/flist/flist.go
```diff
@@ -204,6 +212,11 @@ func (f *flistModule) getMountOptions(pidPath string) (options, error) {
 	return result, nil
 }
 
+// NamedUmount implements the Flister.NamedUmount interface
+func (f *flistModule) NamedUmount(name string) error {
+	return f.Umount(filepath.Join(f.mountpoint, name))
```
muhamadazmy, line 254: Same check on the name must be done like in the mount, to avoid umounting of other parts of the system by carefully crafting a name.

pkg/provision/kubernetes.go
muhamadazmy: They don't match one to one except for the NetworkID, but containers 'Network' struct also has List of IPs while a machine has only one.

pkg/provision/kubernetes.go
muhamadazmy: I assume this part was missed out in implementation. What information is needed to be returned to the User so I can add it?

pkg/provision/kubernetes.go
muhamadazmy: Yes, but this is a convention I suggest (and use) in the other types of provision to make sure the method always return the correct type. You see, the Provision method signature return (interface{}, error) this can allow the code to return Anything or anytime which can make the method inconsistent. To allow the compiler to check for the return type we move the impl to an 'Impl' method, and then add the single line wrapper that return the `(interface{}, error) required signature

pkg/storage/disk.go
muhamadazmy: :+1:

pkg/storage/filesystem/btrfs.go
muhamadazmy: Unfortunately this is the only accurate way.

pkg/vm.go
muhamadazmy: Hmm, i actually started with using both net.IPNet and net.IP types first and then switched to strings, don't quiet remember why, but at this moment I assume it makes client code easier. (provisiond) ? should I switch back to `net.` types.

pkg/vm/strategy.go
muhamadazmy: That's was the plan yes, but we can always restore it, what do u think?

pkg/provision/kubernetes.go
zaibon: Actually for container we only assign 1 address. Maybe we should then modify the container network config to also only accept one IP.

pkg/provision/kubernetes.go
zaibon: Any information the sysadmin requires to be able to use kubectl against the cluster. I do not know what this is.

pkg/provision/kubernetes.go
zaibon: Yes I know, but here it doesn't return anything. So there is no point doing it at all.

pkg/vm.go
zaibon: I prefer to use proper types. Anything can be stored into string, this just force more conversion along the way I think.

pkg/vm/strategy.go
zaibon: I'm ok with keeping it for now

pkg/provision/kubernetes.go
zgorizzo69: it is the file generated by k3s when launched for the first time it is  written to /etc/rancher/k3s/k3s.yaml inside that file there is a server ip (clusters / cluster /server:) we should make sure the ip in here points to the real ip of the master node adn not 127.0.0.1

muhamadazmy: Hmm, that is gonna be tricky to read a file from inside the VM. only idea that comes to mind is to mount the raw disk in in `ro` mode and then try to read that file.

pkg/provision/kubernetes.go
muhamadazmy: But we going to change that as discussed above.

zaibon: right... Forgot this :-)

pkg/provision/kubernetes.go
muhamadazmy: We decided that the user can use his ssh key to retrieve the required file

PR #488, opened 2020-01-23
provisiond: avoid to try to decommission reservation that are not provisioned

zaibon:
This is specially important for node booting cause the first time they
ask for reservation they get the full list of reservation for a node.
Skipping not provisioned reservation makes this process nearly a no-op.

### DIFF
--- a/pkg/provision/engine.go
+++ b/pkg/provision/engine.go
@@ -13,12 +13,13 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
-// ReservationReadWriter define the interface to store
+// ReservationCache define the interface to store
 // some reservations
-type ReservationReadWriter interface {
+type ReservationCache interface {
 	Add(r *Reservation) error
 	Get(id string) (*Reservation, error)
 	Remove(id string) error
+	Exists(id string) (bool, error)
 	Counters() pkg.ProvisionCounters
 }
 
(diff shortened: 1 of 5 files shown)


pkg/provision/engine.go
```diff
@@ -13,12 +13,13 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
-// ReservationReadWriter define the interface to store
+// ReservationCache define the interface to store
 // some reservations
-type ReservationReadWriter interface {
+type ReservationCache interface {
```
muhamadazmy, line 18: Good, this name was really confusing.

muhamadazmy (approved):
LGTM. But the tests are failing.

muhamadazmy:
Also one improvement suggestion:
- If the BCDB did not send reservations that are expired, this will save a lot of un-needed data to be sent over the wire, specially after (hopefully) years of operation. 

The node will still detect expiration with normal methods (check the local reservation expiration, etc...) so I think this would be the better approach rather than skipping the reservation locally, unless I am missing something.

zaibon:
You cannot really received only not expired reservation. Cause if for some reason the node still have some reservation provisioned, it needs to be aware of reservation so it can decommission them.
That is one of the reason the node ask for all of them.

PR #489, opened 2020-01-28
update readme to include build path

zgorizzo69:


docs/tfuser/readme.md
```diff
@@ -9,6 +9,36 @@ It has 3 main commands:
 - **provision**: used to send schema created with the generate command to a node
 
 ## Examples
+### Build tfuser
```
zaibon: I would move this just before the `Examples` title instead.

zaibon (changes requested):

docs/tfuser/readme.md
zgorizzo69: feel free to merge but imho you should first learn to build/make the tool before explaining how to use it

zaibon: yes I think so too, this is why I asked to move this build step **before** the example section.

zaibon (approved):

PR #492, opened 2020-01-29
identityd: only copy file from flist when they are regular file

zaibon:


maxux (approved):

pkg/upgrade/upgrade.go
```diff
@@ -377,7 +377,7 @@ func copyRecursive(source string, destination string, skip ...string) error {
 			if err := os.MkdirAll(dest, info.Mode()); err != nil {
 				return err
 			}
-		} else {
+		} else if info.Mode().IsRegular() {
```
muhamadazmy, line 380: This is okay for now, but may be we need to fix how `symlinks` are handled, but skip special files like devices, fifo, and unix-sockets (these types are supported by the flist format)

Also, same logic should be applied to bootstrap

PR #496, opened 2020-01-30
K8s testnet

zaibon:


zaibon:
I've no idea why but with these change I fail to start a vm. The error I get is:
```
failed to apply provision error="failed to install k3s: could not run vm: error: The argument '--api-sock <api_sock>' was provided more than once, but cannot be used multiple times\n\nUSAGE:\n    firecracker --api-sock <api_sock> --config-file <config-file> --id <id> --seccomp-level <seccomp-level>\n\nFor more information try --help\n: failed to spawn vm machine process '17-2'" id=17-2
```

PR #510, opened 2020-02-03
Fix #509: remove tap if network is already expired

### LINKED ISSUE #509: kubernetes decomission fails if network is already decomissioned
The net resource is opened to delete the tap device, which fails since the net resource is no longer there

LeeSmet:


zaibon (approved):
LGTM. way more elegant like this :+1:

PR #515, opened 2020-02-06
networkd: allow container to choose to have a public ipv6

zaibon:
We now have 2 mode for container networking.
- normal mode, the container is only accessible over the user overlay
  network.
- public 6 mode, the container has 2 networks, one for the user overlay
  network and one for public IPv6


delandtj (approved):
Coooool

mooleshacat:
how we change it?

PR #516, opened 2020-02-11
Some modifications to tfuser

muhamadazmy:
This will make sure we can reuse the same reservation generated
and apply the same schema on multiple nodes with no issue


go.mod
```diff
@@ -37,7 +37,7 @@ require (
 	github.com/go-openapi/validate v0.19.5 // indirect
 	github.com/go-redis/redis v6.15.5+incompatible
 	github.com/gogo/googleapis v1.3.0 // indirect
-	github.com/gomodule/redigo v1.7.0
+	github.com/gomodule/redigo v2.0.0+incompatible
```
zaibon: I do not think this is needed in this PR.

zaibon (changes requested):

go.mod
muhamadazmy: no, but not sure where this came from actually. will revert it.

zaibon (approved):

PR #517, opened 2020-02-11
Updated docs Case2 run VM command

maximevanhees:
Documentation command did not specify what the name of the bridge should be. Changed the command to run the VM so it's in line with the rest of the documentation.

Added `-b zos0` in case 2, step 2
`./vm.sh -n myzos-01 -b zos0 -c farmer_id=2 printk.devmsg=on runmode=dev`

EDIT: thanks to Christophe de Carvalho

### DIFF
--- a/qemu/README.md
+++ b/qemu/README.md
@@ -141,7 +141,7 @@ sudo ip link set forzos up
 1. Now run your vm
 
 ```bash
-./vm.sh -n myzos-01 -c farmer_id=2 printk.devmsg=on runmode=dev
+./vm.sh -n myzos-01 -b zos0 -c farmer_id=2 printk.devmsg=on runmode=dev
 ```
 
 where `runmode` is one of `dev` , `test`  or `prod`, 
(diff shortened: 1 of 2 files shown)


zaibon (changes requested):
I would have fixed this in another way. 
Since all the documentation is referring to a bridge called `zos0`, lets change the default value in the vm.sh script to `zos0`: https://github.com/threefoldtech/zos/blob/4da76c3c483a2724abac4206b9472d350d29f9de/qemu/vm.sh#L8

So all the documentation stays correct and there is no need for the extra `-b` argument in the start command.

maximevanhees:
> I would have fixed this in another way.
> Since all the documentation is referring to a bridge called `zos0`, lets change the default value in the vm.sh script to `zos0`:
> 
> https://github.com/threefoldtech/zos/blob/4da76c3c483a2724abac4206b9472d350d29f9de/qemu/vm.sh#L8
> 
> So all the documentation stays correct and there is no need for the extra `-b` argument in the start command.

Of course. That makes more sense. I'll change it.

zaibon (changes requested):
LGTM
Congrats on your first PR :1st_place_medal:

PR #518, opened 2020-02-12
add staticcheck to CI

zaibon:


pkg/gedis/commands_provision.go
```diff
@@ -431,13 +431,8 @@ func k8sReservation(i interface{}, nodeID string) types.TfgridWorkloadsReservati
...
 	}
 
-	for i, ip := range k.MasterIPs {
-		k8s.MasterIps[i] = ip
-	}
-
-	for i, key := range k.SSHKeys {
-		k8s.SSHKeys[i] = key
-	}
+	copy(k8s.MasterIps, k.MasterIPs)
```
muhamadazmy, line 434: Why not `k8s.MasterIps = k.MasterIPs` ? I know this will make both arrays pointer to the same array, but do we care?

muhamadazmy (approved):

PR #520, opened 2020-02-12
network: ensure we do not clean ipam leases cache during an update of networkd

zaibon:


pkg/app/boot.go
```diff
@@ -0,0 +1,27 @@
...
+	if err := os.MkdirAll(bootedPath, 0770); err != nil {
+		return err
+	}
+	path := filepath.Join(bootedPath, name)
+	_, err := os.OpenFile(path, os.O_RDONLY|os.O_CREATE|os.O_TRUNC|os.O_EXCL, 0666)
+	return err
+}
+
+// IsFirtBoot checks if the a file has been created by MarkBooted function
+func IsFirtBoot(name string) bool {
```
muhamadazmy: typo, should be `IsFirstBoot`

muhamadazmy (changes requested):

PR #521, opened 2020-02-12
Check if node is virtualized

### LINKED ISSUE #337: Detect if 3Node is physical hardware or virtual machine
right now we can not detect if capacity is generated by physical or virtual hardware. virtual hardware can be heavily oversubscribed and would help to cheat the system. that's why we need a detector which makes it clear what we are dealing with.

- [x] research how to detect virtual machines
- [ ] implement VM detection in 0-OS
  - [ ] add new IsVM method on the `ResourceOracle` type https://github.com/threefoldtech/zos/blob/f90d4fdfa516266d74eb391118d5fe4b9b63c8fd/pkg/capacity/capacity.go#L19-L21. The method will call `virt-what` command and parse its output
  - [ ] update main function of capacityd to send the IsVM value to BCDB: https://github.com/threefoldtech/zos/blob/f90d4fdfa516266d74eb391118d5fe4b9b63c8fd/cmds/capacityd/main.go#L61-L62
 - [ ] Update nodes schema so it store the VM detection result: https://github.com/threefoldtech/zos/blob/f90d4fdfa516266d74eb391118d5fe4b9b63c8fd/pkg/gedis/types/directory/tfgrid_node_2.go#L10

maximevanhees:
Checks if the 3Node is a virtual machine or not using _virt-what_

Thanks to @muhamadazmy for the help!


pkg/capacity/capacity.go
```diff
@@ -99,3 +103,19 @@ func (r *ResourceOracle) Disks() (d Disks, err error) {
 
 	return
 }
+
+// IsVM detect if machine is virtual
+func (r *ResourceOracle) IsVM() (bool, error) {
```
zaibon: After reflection, instead of returning just a boolean, I would prefer we return the list of thing virt-what have found.
So we can report the result to BCDB.

qemu/overlay/bin/virt-what
```diff
@@ -0,0 +1,399 @@
+#!/bin/sh -
```
zaibon: you should not commit this file.

pkg/capacity/store.go
```diff
@@ -53,15 +53,17 @@ func NewHTTPStore(url string) *HTTPStore {
...
 	x := struct {
 		Capacity Capacity `json:"capacity"`
 		DMI      dmi.DMI  `json:"dmi"`
 		Disks    Disks    `json:"disks"`
+		IsVM     bool     `json:"is_vm"`
 	}{
 		Capacity: c,
 		DMI:      d,
 		Disks:    disks,
+		IsVM:     isVM,
```
zaibon: Since you modify the structure we sent to BCDB, you also need to add this field into the struct that will be used to deserialize the request server side, which happens here:
https://github.com/threefoldtech/zos/blob/acf49fbc5780985fcba995838606fe8a35ab852f/tools/bcdb_mock/node_handlers.go#L127-L131
And ultimately add it into this type:
https://github.com/threefoldtech/zos/blob/acf49fbc5780985fcba995838606fe8a35ab852f/pkg/gedis/types/directory/tfgrid_node_2.go#L58-L64

zaibon (changes requested):

pkg/capacity/capacity.go
```diff
@@ -99,3 +103,15 @@ func (r *ResourceOracle) Disks() (d Disks, err error) {
 
 	return
 }
+
+//GetHypervisor gets the name of the hypervisor used on the node
+func (r *ResourceOracle) GetHypervisor() (string, error) {
```
zaibon: This funstion should return a slice of string ([]string) since virt-what can output multiple line.

pkg/capacity/store.go
```diff
@@ -29,12 +29,12 @@ func NewBCDBStore(gedis *gedis.Gedis) *BCDBStore {
 }
 
 // Register sends the capacity information to BCDB
-func (s *BCDBStore) Register(nodeID pkg.Identifier, c Capacity, d dmi.DMI, disks Disks) error {
+func (s *BCDBStore) Register(nodeID pkg.Identifier, c Capacity, d dmi.DMI, disks Disks, hypervisor string) error {
```
zaibon:
```suggestion
func (s *BCDBStore) Register(nodeID pkg.Identifier, c Capacity, d dmi.DMI, disks Disks, hypervisor []string) error {
```

pkg/capacity/store.go
```diff
@@ -53,15 +53,17 @@ func NewHTTPStore(url string) *HTTPStore {
 }
 
 // Register sends the capacity information to BCDB
-func (s *HTTPStore) Register(nodeID pkg.Identifier, c Capacity, d dmi.DMI, disks Disks) error {
+func (s *HTTPStore) Register(nodeID pkg.Identifier, c Capacity, d dmi.DMI, disks Disks, hypervisor string) error {
```
zaibon:
```suggestion
func (s *HTTPStore) Register(nodeID pkg.Identifier, c Capacity, d dmi.DMI, disks Disks, hypervisor []string) error {
```

pkg/gedis/commands_capacity.go
```diff
@@ -46,12 +46,13 @@ func (g *Gedis) UpdateUsedNodeCapacity(node pkg.Identifier, mru, cru, hru, sru u
 }
 
 // SendHardwareProof sends a dump of hardware and disks to BCDB
-func (g *Gedis) SendHardwareProof(node pkg.Identifier, dmi interface{}, disks interface{}) error {
+func (g *Gedis) SendHardwareProof(node pkg.Identifier, dmi interface{}, disks interface{}, hypervisor string) error {
```
zaibon:
```suggestion
func (g *Gedis) SendHardwareProof(node pkg.Identifier, dmi interface{}, disks interface{}, hypervisor []string) error {
```

pkg/gedis/types/directory/tfgrid_node_2.go
```diff
@@ -61,6 +61,7 @@ type TfgridNodeProof1 struct {
 	DiskHash     string                 `json:"disk_hash"`
 	Hardware     map[string]interface{} `json:"hardware"`
 	Disks        map[string]interface{} `json:"disks"`
+	Hypervisor   string                 `json:"hypervisor"`
```
zaibon:
```suggestion
	Hypervisors   []string                 `json:"hypervisor"`
```

tools/bcdb_mock/nodes_store.go
```diff
@@ -140,14 +140,15 @@ func (s *nodeStore) updateUptime(nodeID string, uptime int64) error {
 	return nil
 }
 
-func (s *nodeStore) StoreProof(nodeID string, dmi dmi.DMI, disks capacity.Disks) error {
+func (s *nodeStore) StoreProof(nodeID string, dmi dmi.DMI, disks capacity.Disks, hypervisor string) error {
```
zaibon:
```suggestion
func (s *nodeStore) StoreProof(nodeID string, dmi dmi.DMI, disks capacity.Disks, hypervisors []string) error {
```

zaibon (changes requested):

qemu/overlay/bin/virt-what
muhamadazmy: Same for `virt-what-cpuid-helper`

tools/bcdb_mock/nodes_store.go
muhamadazmy: @zaibon this method signature is becoming too long, shouldn't we create (or reuse) one of the structures that already have these attributes? (same applies for other methods regarding capacity registration)

zaibon: Yep I was thinking the same.

pkg/capacity/capacity.go
```diff
@@ -99,3 +103,15 @@ func (r *ResourceOracle) Disks() (d Disks, err error) {
...
+//GetHypervisor gets the name of the hypervisor used on the node
+func (r *ResourceOracle) GetHypervisor() (s []string, err error) {
+	out, err := exec.Command("virt-what").CombinedOutput()
+
+	if err != nil {
+		return s, fmt.Errorf("could not detect if VM or not: %s", string(out))
+	}
+
+	out = bytes.TrimSpace(out)
+	return s, nil
```
zaibon:
```suggestion
      return strings.Split(string(out), "\n"), nil
```
you forgot to return the actual content of the command result, now you just return an empty slice of string

tools/bcdb_mock/nodes_store.go
```diff
@@ -134,14 +134,15 @@ func (s *nodeStore) updateUptime(nodeID string, uptime int64) error {
...
+func (s *nodeStore) StoreProof(nodeID string, dmi dmi.DMI, disks capacity.Disks, hypervisor []string) error {
 	node, err := s.Get(nodeID)
 	if err != nil {
 		return err
 	}
 
 	proof := directory.TfgridNodeProof1{
-		Created: schema.Date{Time: time.Now()},
+		Created:    schema.Date{Time: time.Now()},
+		Hypervisor: hypervisor,
```
zaibon, line 145: Actually this is not enough. You will see in the code down, that it is possible that we do not store the `proof` variable:
https://github.com/threefoldtech/zos/blob/a7a8e9b6ea2a507f4e648c2d75cbb576bc0d6ae6/tools/bcdb_mock/nodes_store.go#L170-L174
I would replace this block with
```go
// don't save the proof if we already have one with the same
	// hash/content
	for i := range node.Proofs {
		if proof.Equal(node.Proofs[i]) {
			// always update hypervisor content
			node.Proofs[i].Hypervisor = hypervisor
			return nil
		}
	}
```
So we ensure that we always update the content of hypervisor even if the proof is already known and we do not store a new one.

zaibon (changes requested):

pkg/capacity/capacity.go
```diff
@@ -105,13 +105,18 @@ func (r *ResourceOracle) Disks() (d Disks, err error) {
...
 
 	if err != nil {
-		return s, fmt.Errorf("could not detect if VM or not: %s", string(out))
+		return nil, fmt.Errorf("could not detect if VM or not: %s", string(out))
 	}
 
-	out = bytes.TrimSpace(out)
-	return s, nil
+	str := strings.TrimSpace(string(out))
+	if len(str) == 0 {
```
zaibon, line 116: :+1:

zaibon (approved):

PR #538, opened 2020-02-18
internet: improve logic used to choose the slave interface of zos bridge

### LINKED ISSUE #537: improve logic of internet service to make sure we grab the proper ip for zos bridge
Internet service is the first service that is started on the node, its role is to scan all network interfaces and try to detect which one receives an IP address that provide internet access to the node.

The current logic is the following:
- list all physical interfaces of the node
- filter out interfaces that doesn't have a cable plugged in it
- attach the interface to the zos bridge
- start a DHCP probe to try to receive an IP from the farm DHPC server
- if we got an IP (whether it's an ipv6 or ipv4), exit the loop
- if we didn't received anything, detach the interfaces and try with another one

This flow has a problem, in a farm that uses 2 different network for management and public, it can happen that the DHPC of the public network answer first. So the zos bridge is attached on an interfaces that is supposed to be used for public network. 

The proposed solution to improve this is the following:
- create a temp namespace per interface that has carrier and is physical
` ip netns add $interfacename`
- send interface in that namespace
`ip link set $interfacename netns $interfacename`
- in all namespaces, `ip l set $interfacenale,lo up` , start dhcp and enable ipv6
- wait a $timeout bit
- collect ip addrs, routes

At this stage we have now a list of IP given by the DHCP.

- In this list of IP, detect which one is the management IP (RFC 1918)
- attach the interface that received the management IP on the zos bridge and assign the IP address to it.
With this flow we have the change to actually check the IPs we received from DHCP and make the appropriate decision on which one to use for zos.

zaibon:
fixes #537


pkg/network/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,331 @@
...
+			}
+		}()
+
+		var addrs4 = newAddrSet()
+		var addrs6 = newAddrSet()
+		for stay := true; stay; {
+			select {
+			case <-ctx.Done():
+				// exit for loop
+				stay = false
```
muhamadazmy: You can use `labels` here like this 
```
myloop:
for {
   select {
       case x:
           break myloop
   }
}
```

pkg/network/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,331 @@
...
+			// for _, route := range cfg.Routes6 {
+			// 	if route.Gw != nil {
+			// 		selected6 = append(selected6, cfg)
+			// 	}
+			// }
+		}
+	}
+
+	if len(selected4) < 1 {
+		return "", fmt.Errorf("not route with default gateway found")
```
muhamadazmy:
```suggestion
		return "", fmt.Errorf("no route with default gateway found")
```

pkg/network/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,331 @@
...
+
+	return selected4[0].Name, nil
+}
+
+type ifaceFilter func(links []netlink.Link) []netlink.Link
+
+// filterPhysical is a ifaceFilter that filter out the links that are
+// not physical interface
+func filterPhysical(links []netlink.Link) []netlink.Link {
+	out := links[:0]
```
muhamadazmy, line 237: Why this slicing if it will create an empty slice anyway? 

```suggestion
	var out []netlink.Link
```

pkg/network/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,331 @@
...
+		if l.Type() == "device" {
+			out = append(out, l)
+		}
+	}
+	return out
+}
+
+// filterPlugged is a ifaceFilter that filter out the links that does not have a cable plugged in it
+func filterPlugged(links []netlink.Link) []netlink.Link {
+	out := links[:0]
```
muhamadazmy, line 248: Same comment
```suggestion
	var out []netlink.Link
```

pkg/network/dhcp/dhcp.go
```diff
@@ -1,25 +1,29 @@
...
-	if err != nil {
-		return false, err
+// Probe is used to do some DHCP request on a interface
+type Probe struct {
+	ctx context.Context
+	cmd *exec.Cmd
+}
+
+// NewPrope returns a Probe
+func NewPrope(ctx context.Context) *Probe {
```
muhamadazmy: I have read somewhere (can't find the article) that context are supposedly only used for "request" based context. In other words, you should not store the context as an attribute anywhere. Instead it must be passed to a method (not a constructor)

Hence this should be:
```go
func NewPrope() *Prope {}
func (d *Probe) Start(ctx context.Context) erro {}
```
instead.

*Note*: I actually don't see where this context is used anyway!

pkg/network/ndmz/ndmz.go
```diff
@@ -135,12 +136,37 @@ func Create(nodeID pkg.Identifier) error {
 			return errors.Wrapf(err, "ndmz: failed to enable enable_defrtr=1 in ndmz namespace")
 		}
 		// run DHCP to interface public in ndmz
-		received, err := dhcp.Probe(types.PublicIface, netlink.FAMILY_V4)
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
+		probe := dhcp.NewPrope(ctx)
+
+		defer func() {
+			cancel()
```
muhamadazmy: I noticed that this `cancel` is the only place where it's actually called. Hence there is not other trigger to stop this routing from finishing that that it actually just finish execution. 

I am not sure this is the desired scenario, but if that is the case, then you can just directly use context.Background() or no context at all!

cmds/internet/main.go
```diff
@@ -78,15 +82,52 @@ func bootstrap() error {
...
+			log.Error().Err(err).Msg("failed to select a valid interface for zos bridge")
+			return err
+		}
+
+		br, err := bootstrap.CreateDefaultBridge(types.DefaultBridge)
+		if err != nil {
 			return err
 		}
 
+		time.Sleep(time.Second) // this is dirty
```
muhamadazmy, line 103: Why this is needed?

muhamadazmy (changes requested):

cmds/internet/main.go
zaibon: well I'm not sure. I've seen that sometime the call to `netlink.LinkByName` failed to find the interface. 
I suspect it does too fast and the interface has not had the time to be moved back to the host network namespace after the temporary network namespace has been deleted

pkg/network/bootstrap/bootstrap.go
zaibon: shouldn't the label be after the loop ? Now I will just enter the loop again no ?

pkg/network/bootstrap/bootstrap.go
zaibon: This is not the same, the idea here is to have a slice that points to the same underlying array so we can filter without copying. See https://github.com/golang/go/wiki/SliceTricks#filtering-without-allocating

pkg/network/dhcp/dhcp.go
zaibon: > Note: I actually don't see where this context is used anyway!

Indeed :-P I'll just remove it

pkg/network/ndmz/ndmz.go
zaibon: It is actually used line 158 to timeout and exist the loop there. But since context is just not use by `dhcp.Probe` I can just replace this ctx with a `time.After()` and be done with it.

pkg/network/bootstrap/bootstrap.go
muhamadazmy: no, this is not a JUMP it's a break, for the block labeled `myloop`

muhamadazmy (approved):

PR #539, opened 2020-02-18
Ndmz better pub ipv6

delandtj:
a better way to handle the pub iface in ndmz, so that the kernel handles it by itself


zaibon (approved):
LGTM :+1:

PR #540, opened 2020-02-19
Report primitives

### LINKED ISSUE #524: Report how many primitives are deployed on the node to BCDB
At the moment, the node only report the total amount of resource units it has to BCDB.
While this information is useful for the farmer and the minting, it is not really useful for users.

User needs to know how much capacity if actually still available on the node to be able to choose on which one to reserve capacity.

So I would like the node to report the number of primitives provisioned on the node.
Based on this information it is also possible to compute how much resource units have been reserved from a node.

Code wise, provision package already have a counter for each primitive that is provision on the node: https://github.com/threefoldtech/zos/blob/master/pkg/provision/local_store.go#L20
So the work would be to send the value of this counter to BCDB any time there is a new provisioning or a decommission.

FSStore stuct keep a copy of all the reservation provisionned on the node on disk: 
https://github.com/threefoldtech/zos/blob/master/pkg/provision/local_store.go#L63
Using this, it is possible to parse the content of reach reservation and compute how much resource unit it used by the reservation.

Here is how it works:
- containers uses: CPU and memory which is define by https://github.com/threefoldtech/zos/blob/2e540936e3ce8be1f4d1f62269489366baf9749e/pkg/provision/container.go#L65-L70
- volumes uses only disk space, so SRU for SSD and HRU for HDD : https://github.com/threefoldtech/zos/blob/2e540936e3ce8be1f4d1f62269489366baf9749e/pkg/provision/volume.go#L29-L34
- network: doesn't consume any resource unit
- 0-DB: uses only disk space, so SRU for SSD and HRU for HDD : https://github.com/threefoldtech/zos/blob/2e540936e3ce8be1f4d1f62269489366baf9749e/pkg/provision/zdb.go#L31-L39
- VMs: uses CPU, memory and disk space: https://github.com/threefoldtech/zos/blob/2e540936e3ce8be1f4d1f62269489366baf9749e/pkg/provision/kubernetes.go#L311-L322

maximevanhees:


zaibon (commented):
So far so good. Only remark I would make so far is I would like to have the logic to extract the number of resource unit from a reservation to be extracted into it's own package, it does not really belong into the FSStore type.

PR #545, opened 2020-02-24
networkd: fix bug where we read from a closed channel 

zaibon:


cmds/networkd/addr_watcher.go
```diff
@@ -56,6 +57,43 @@ func (w WatchedLinks) CallBack(update netlink.AddrUpdate) error {
...
+
+				log.Debug().Msgf("addr update received %+v", update)
+				if err := w.callBack(update); err != nil {
+					log.Error().Err(err).Msg("addr watcher: error during callback")
+				}
+			}
+		}
+	}()
+
+	err := <-cErr
```
muhamadazmy: If u waiting on an error anyway (not returing) why this has to be in a go routine ?

muhamadazmy (changes requested):

cmds/networkd/addr_watcher.go
muhamadazmy: I think no need at all to have this in a go routine if you are blocking on the return of that go routine anyway, no ?

PR #547, opened 2020-02-25
Always append zinit logs to the end of logfile

maxux:
When using > to redirect logs, target file offset will be kept even if
file is truncated. This result by incoherent/invalid log filesize.

Using >> will always append to the end of the file and if logfile is
truncated, size will be reset.


delandtj:
Good catch !

muhamadazmy (approved):

delandtj (approved):
YES!

PR #555, opened 2020-03-02
updated macos developer docs

maximevanhees:
By changing a go environment variable, there is no need to deploy an Ubuntu Docker container.


zaibon:
the problem with this approach is that you actually build linux binaries. 
So for component that are run into 0-OS itself this is ok, but you actually cannot build `tfuser`, `bcdb_mock` and other tools and use them on mac.

PR #558, opened 2020-03-03
Improve storage initializing

### LINKED ISSUE #557: HRU and SRU displayed in the wrong way
have my little box on the grid with version 2.0. should show SRU 250 and the rest in HRU. right now all is displayed as HRU. Please check what is causing this. thx

image

maxux:
Wwhen some new devices are available, some list issue occures affecting expected volumes list on runtime.

This fixes #557 and report when filesystem creation fails.

### DIFF
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -11,7 +11,7 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/pkg/errors"
-	"github.com/rs/zerolog/log"
+	log "github.com/rs/zerolog/log"
 	"github.com/shirou/gopsutil/disk"
 	"github.com/threefoldtech/zos/pkg"
 	"github.com/threefoldtech/zos/pkg/storage/filesystem"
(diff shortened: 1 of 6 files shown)


pkg/storage/storage.go
```diff
@@ -114,6 +114,19 @@ func (s *storageModule) BrokenDevices() []pkg.BrokenDevice {
...
 }
 
+func (s *storageModule) Dump() {
+	log.Debug().Int("volumes", len(s.volumes)).Msg("dumping volumes")
+
+	for i := range s.volumes {
+		devices := s.volumes[i].Devices()
+		for id := range devices {
+			d := devices[id]
+			log.Debug().Str("path", d.Path).Str("labem", d.Label).Str("type", string(d.DiskType)).Msg("dump")
```
zaibon:
```suggestion
			log.Debug().Str("path", d.Path).Str("label", d.Label).Str("type", string(d.DiskType)).Send()
```

zaibon: .Send() allow to print the log without the need for a message body. So it just print a line with the arguments.

pkg/storage/storage.go
```diff
@@ -169,17 +182,24 @@ func (s *storageModule) initialize(policy pkg.StoragePolicy) error {
 		return err
 	}
 
-	// collect free disks, sort by read time so faster disks are first
-	sort.Sort(filesystem.ByReadTime(disks))
+	// collect free disks, without sorting 'disks', this break s.volumes list
+	// sort.Sort(filesystem.ByReadTime(disks))
```
zaibon:
```suggestion
```

zaibon (changes requested):

zaibon (approved):

PR #561, opened 2020-03-05
Plan bee

muhamadazmy:


tools/bcdb_mock/pkg/directory/farms_store.go
```diff
@@ -0,0 +1,54 @@
...
+// GetByID gets a farm by ID
+func (s *FarmAPI) GetByID(ctx context.Context, db *mongo.Database, id int64) (directory.Farm, error) {
+	var filter directory.FarmFilter
+	filter = filter.WithID(schema.ID(id))
+
+	return filter.Get(ctx, db)
+}
+
+// Add add farm to store
+// TODO: support update farm information ?
```
zaibon, line 51: yes indeed, we need to be able to update some of the field of a farm. see: https://github.com/threefoldtech/jumpscaleX_threebot/blob/ab4ddcb7014dd226affe86d5d891e4881e81ed87/ThreeBotPackages/tfgrid/directory/actors/farms.py#L62-L75

tools/bcdb_mock/pkg/directory/node_handlers.go
```diff
@@ -0,0 +1,242 @@
...
+	log.Debug().Str("node", nodeID).Uint64("uptime", input.Uptime).Msg("node uptime received")
+
+	if err := s.updateUptime(r.Context(), db, nodeID, int64(input.Uptime)); err != nil {
+		return nil, mw.NotFound(err)
+	}
+
+	return nil, nil
+}
+
+func (s *NodeAPI) updateUsedResources(r *http.Request) (interface{}, mw.Response) {
```
zaibon: This should be `updateReservedResources` we do not track "Used" resources, only total and reserved

tools/tfuser/cmds_identity.go
```diff
@@ -24,10 +28,16 @@ func cmdsGenerateID(c *cli.Context) error {
...
+		return errors.Wrap(err, "failed to register user")
+	}
+
 	if err := k.Save(c.String("output")); err != nil {
 		return errors.Wrap(err, "failed to save seed")
 	}
-	fmt.Printf("new identity generated: %s\n", k.Identity())
-	fmt.Printf("seed saved at %s\n", output)
+
+	fmt.Printf("Your ID is: %d\n", id)
```
zaibon, line 40: I think we should also store the user ID somewhere no ?

PR #563, opened 2020-03-06
use wipefs to clean up v1 disks

muhamadazmy:


maxux:
Could you use `-b` to get backup in addition ?

PR #565, opened 2020-03-09
Cache issue

maximevanhees:
Current flow:
1. Try to make cache on SSD, if not available on HDD
2. When failing to do both, create cache-limited file (used as flag) in /tmp/flags
3. Check if flag "limited-cache" is true, if so, do not provision
4. Retry the provision every 30 sec by looking if flag is still set


maxux:
For this specific PR, I would handle that in a different way.

Keeping track of system status using temporary files is IMO not adapted. Since you can check the real status as easily than checking file presence, adding this layer add possible inconsistency (file exists and should not, file doesn't exists and should) while you could simply avoid it completely.

In this specific case, I would have change the `filesystem.IsMountPoint` to not return a `bool` but returns a string, which contains the mountpoint type if found.

This is a simple change on the `filesystem.IsMountPoint` function since it already reads `/proc/mounts`, if entry is found, just returns the 3rd item on the line (which is the filesystem type).

The `filesystem.IsMountPoint` is only called 1 time, so it's not a big change on the whole code base. After that, you don't have to check the flag status, just check if `/var/cache` is mounted (which is already done), and type is `tmpfs` then you know what to do and your check is always up-to-date without inconstancy possible :)

maximevanhees:
Following @maxux 's comment. If I now need to change it, just let me know. Thanks for the reply!

muhamadazmy:
@maxux While i fully agree with your concern, but this raise another issue. The provision (or other daemons) must now be aware of storaged operation and also depends on its implementation. This means that changes to storaged can break provisiond (for example if we changed the limited cache underlying filesystem or decided not to mount anything at all)

With the file flags, this allow us to separate concerns, and make sure changes to storaged doesn't break provisiond.

maxux:
> for example if we changed the limited cache underlying filesystem or decided not to mount anything at all

I don't see how a flag file can help that. And since all implementation share same code, you could reuse same function in the code to check status.

cmds/provisiond/main.go
```diff
@@ -51,6 +52,14 @@ func main() {
 		log.Logger.Level(zerolog.DebugLevel)
 	}
 
+	// keep checking if limited-cache flag is set
+	if app.CheckFlag("limited-cache") {
+		log.Info().Msg("failed cache reservation! Retrying every 30 seconds...")
```
zaibon: Let's log using Error() instead of info so it's easier for us to grep on the logs.

muhamadazmy: Please also use the defined constant, not string "limited-cache"

cmds/provisiond/main.go
```diff
@@ -51,6 +52,14 @@ func main() {
 		log.Logger.Level(zerolog.DebugLevel)
 	}
 
+	// keep checking if limited-cache flag is set
+	if app.CheckFlag(app.LimitedCache) {
+		log.Error().Msg("failed cache reservation! Retrying every 30 seconds...")
+		for app.CheckFlag("limited-cache") {
```
muhamadazmy: What about this one :P

maximevanhees: oopsie, will change now!

pkg/app/flag.go
```diff
@@ -0,0 +1,51 @@
...
+
+// CheckFlag checks the status of a flag based on a key
+func CheckFlag(key string) bool {
+	_, err := os.Stat(filepath.Join(flagsDir, key))
+	return !os.IsNotExist(err)
+}
+
+// DeleteFlag deletes (unsets) a given flag based on a key
+func DeleteFlag(key string) error {
+	d, err := os.Open(flagsDir)
```
muhamadazmy: Why do u open the directory here ?

pkg/storage/storage.go
```diff
@@ -340,39 +342,59 @@ func (s *storageModule) ensureCache() error {
...
-		filesystems, err := s.volumes[idx].Volumes()
-		if err != nil {
-			return err
-		}
-		for jdx := range filesystems {
-			if filesystems[jdx].Name() == cacheLabel {
-				log.Debug().Msgf("Found existing cache at %v", filesystems[jdx].Path())
-				cacheFs = filesystems[jdx]
+	// check if mounted
+	if !filesystem.IsMountPoint(cacheTarget) {
```
muhamadazmy: I would do this differently actually

```go
if filesystem.IsMoutPoint(cacheTarget) {
return nil
}
```

This I won't have to indent all this code in an if statement

pkg/app/flag.go
maximevanhees: added check to make sure directory exist, in case the code would execute when there is no flag set. A "just-to-be-sure". Should it be removed?

muhamadazmy: First of all, a check if the directory exists would be more efficient if u do `os.Stat` on the file, not a Open, then close. Secondly, a RemoveAll call does not fail if the file does not exist in the first place, so a check is not really needed at all.

cmds/provisiond/main.go
maximevanhees: fixed

cmds/provisiond/main.go
maximevanhees: fixed

pkg/app/flag.go
maximevanhees: fixed

pkg/storage/storage.go
maximevanhees: fixed

pkg/app/flag.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// CheckFlag checks the status of a flag based on a key
+func CheckFlag(key string) bool {
+	_, err := os.Stat(filepath.Join(flagsDir, key))
+	return !os.IsNotExist(err)
+}
+
+// DeleteFlag deletes (unsets) a given flag based on a key
+func DeleteFlag(key string) error {
```
muhamadazmy, line 41: this function is vulnerable to "path" injection. I can pass a key as "." and then the filepath.Join(flagsDir, key) will return "flagsDir" itself witch will cause the function to delete the entire flagsDir director. A more dangerous attack can be done by setting key to "../" or other calls that can cause the function to delete specific paths on the system knowing how to go there from the flagsDir. 

A check can be as simple as making sure the key doesn't have special "/" character. Or making sure that the final file path to delete is always under flagsDir.

muhamadazmy (approved):

muhamadazmy (approved):

PR #567, opened 2020-03-11
don't block provision engine when a wrongly formatted reservation comes

### LINKED ISSUE #554: provisiond: provision engine can be blocked by a bad formatted reservation
The code that poll for new reservation handle error by just logging the error and restart the loop:
https://github.com/threefoldtech/zos/blob/328df4540f655658acebe9a0d1451f5c046c9f3d/pkg/provision/source.go#L65-L69

The problem is for some reason a reservation that cannot be deserialized is sent from BCDB, then we endup in a loop that will always be blocked on this reservation forever.

I think we need a way to somehow skip the badly formatted reservation and continue processing the others.

zaibon:
in

### DIFF
--- a/pkg/provision/source.go
+++ b/pkg/provision/source.go
@@ -2,6 +2,7 @@ package provision
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 	"time"
@@ -65,6 +66,11 @@ func (s *pollSource) Reservations(ctx context.Context) <-chan *Reservation {
 			res, err := s.store.Poll(pkg.StrIdentifier(s.nodeID), next)
 			if err != nil && err != ErrPollEOS {
 				log.Error().Err(err).Msg("failed to get reservation")
+				// if this is not a temporary error, then skip the reservation entirely
+				// and try to get the next one
+				if !errors.Is(err, ErrTemporary) {
+					next++
+				}
 				continue
 			}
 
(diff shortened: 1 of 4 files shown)


pkg/provision/source.go
```diff
@@ -65,6 +65,7 @@ func (s *pollSource) Reservations(ctx context.Context) <-chan *Reservation {
 			res, err := s.store.Poll(pkg.StrIdentifier(s.nodeID), next)
 			if err != nil && err != ErrPollEOS {
 				log.Error().Err(err).Msg("failed to get reservation")
+				next++
```
LeeSmet: Do we always want  to skip this reservation? I.E is not possible that poll returns a temporary network error or something

zaibon: I guess we could do that. but the reservation is never going to be fixed. What should be the behavior here if we get a "temporary error" ?

LeeSmet: Well, if its a network error where the explorer is temporarily unreachable (for this node), then it would be possible for this reservation to become valid in the future right? So skipping it here would mean that it would no longer get provisioned then

zaibon: Ok I see what you mean now. But we are talking about 2 different problem here.
- if the network is the problem, in this case we should indeed not increment next and just wait to retry
- the reservation format is invalid and we cannot unmarshal it. In this case the reservation should be skipped

LeeSmet (approved):

PR #570, opened 2020-03-12
networkd: make reserved wireguard port persistent between network restart

### LINKED ISSUE #556: used wireguard port are not properly reported on testnet
```python
for n in nodes: 
   2     print(n.node_id, n.wg_ports)                                                                                                                                                                     
2fi9ZZiBGW4G9pnrN656bMfW6x55RSoHDeMrd9pgSA8T []
72CP8QPhMSpF7MbSvNR1TYZFbTnbRiuyvq5xwcoRNAib []
D1sY7tgHX6fVvjrpCotr4SmezVyRShLTufqhEpBQxK1v []
6m9gFPTYLRiCH7FD8yU5VWwXZdfegKUvXZgrJoS1EJfQ []
2m3nHPSAMyZFSeg5HPozic2NGBMtrXrBkhtNcVmd5Ss6 []
3dAnxcykEDgKVQdTRKmktggL2MZbm3CPSdS9Tdoy4HAF []
3BwDVNRigau8Zk6BQkJt5PLxrHQnJ5NTfNvAtuubC1eS []
A7FmQZ72h7FzjkJMGXmzLDFyfyxzitDZYuernGG97nv7 []
7fHSAHEvUGtUcYSqLtpGq8ANssPikTyyHC52FddDYF4Y []
FjwyHVvfATkVb4Puh4x6jCMS79TVVgSYagAuZTxWrsbj []
HT9yNvHRQs65dyRJeztAdEnXKweCX2bhgvD9i1WFPUNs - 1,000

9211BFV7MFwktD2b8jHE9Ub3fHRtaYQyBBfwT9kEKA7q []
HugtVL51BFNLbZbbxnWu2GEe8hV97YVPac19zy5wwNpT []
9KAbX21NGbZYupBJ6EeeWx3ZTKDx7ADevr8qtmEa5WkC []
9kcLeTuseybGHGWw2YXvdu4kk2jZzyZCaCHV9t6Axqqx []
3h4TKp11bNWjb2UemgrVwayuPnYcs2M1bccXvi3jPR2Y []
FUq4Sz7CdafZYV2qJmTe3Rs4U4fxtJFcnV6mPNgGbmRg []
5Pb5NMBQWLTWhXK2cCM8nS6JZrnP2HaTP452TfMMYT9p []
DUF2knurdMuX2eJVp9o7tXq4eNBy2fbxBoWhrMXWPEtF []
8zdqjFD7GLsSSfsTgFYcGusw91gQ3tdx7jbUhJep2a5X []
6chi1iSczxfF4U2iyCcJwkwWnwzcDgQHzCRExK9r4V1j []
AT4hKXnFWPNsnE38sUoERDVzxL9mNysqDWQSdHkum7YD - 8,192

J1Xb2piba1vZM2pSmWP24CPgREnoTce7EuUJysttCyz6 - 8,193
- 8,194
- 1,000

9161hj9VnM41HVvTr7nHQKcgFBx8Yf6AjWduMCrBXQip - 8,193
- 8,194
- 1,000

33zKcb6YBSuYDVrTsm6AjBuTpujP4na7kL6ZrShcVKbF - 8,194
- 1,000
- 8,193

FLeByELHK22evRYqJokHsCMCF9wQpLtuyehUp1LgR5eM []
49RnnxyRJ7r2So6UMBV2UDyLtoyRjmAf8NEnDnzpMtmY []
ByTfXHH7RogHNLpTpBWbn34ofC2tfCBVhFKJxYnMBoNi []
A34YUGenHKyhjDMAUKZe4cVDtJM2wQ4n4XRkfGUUEYdy - 1,000
- 8,192

HYwvrxCy5z1QPALC5gTAiAMYWLH7orQ8xWEDbEyLiJV7 - 8,192
- 1,000

CBDY1Fu4CuxGpdU3zLL9QT5DGaRkxjpuJmzV6V5CBWg4 - 8,192
- 1,000

Hb6oVe2B5v9UBzDcDeQfZGn5bwFeM2R3rJh6U93AWfiN - 8,192
- 1,000

54S1qFXxWgnjmvEFVvqUbR7dHhvCshLbQrp2UpmE7GhZ - 8,192
- 1,000

9WhwSbM2xBNb9E3ws3PNJfyeajnKXWDZAMBLZMXCA9jf - 8,192
- 1,000

2hgRioV9ZKe8Apnm84TZn8Bn5XczyPU2nkPmozUY4rYw - 8,192
- 1,000

6gBWuYT8MTCJiRJp3mLiPLSXDJpDCGrLyi4HpVissb5j - 8,192
- 1,000

59R1GU6qh4RiWQRvUFm4GzwBbTM2JStyTFmgpzruq9Gt []
292NEoCJ98FfhB62qZXXKqBQyLUij54roeH5VwBuXsSp []
HARGRBPyxF315bXDBCdaTnAD9cqMpwxvQnnzkZjhpVCA []
Ddstj2hPbGip8Ci26TUaEzY52DHZs3JphLFXy9UgWSRS []
FZZxePXsMfWpoepiPtdmc7EjRs3i8cmFF8xXJsFHKFB4 []
Aiux6s3V8wg398FHzWWy3LC81C8vuP4JLkck7uR9T3pD []
D9pVi21QMozECuZ6inVKQZP5UFrtusKUhW5mfW5fRxbK []
35A85gq6FRBFjMfdHDE6d7j5BYr786h4afy1wNcmdzre []
4TksThCbMYo8THcsADEckaCwurHSbn6EvDapfXvga97r []
JE3GQ7LhStnoTR5mSfJZE1gcLhTr2oMiURVH1C1ipsgo []
7Qa7fxRPtMc5R72cnNms4XWD9PJkeQJ6iD8iTU9T8p2o []
9if6GSLuz1awA9EAywgYWDdgyH7Aq79X68kv9adJNAFL []
```
This is an output of all the wireguard port marked as used for all the nodes from testnet.
While when I check the amount of network resource deployed on some nodes there should be more wireguard port marked as used.

This also make new provision of network fails cause user doesn't know which port is used or not. Which make there reservation failed with 
```
[+] provisiond: 2020-03-02T14:10:15Z error failed to apply provision error="failed to create network resource for network U7Xfof9T9CHuL: failed to configure network resource: failed to configure wireguard interface: failed to bring wireguard interface w-U7Xfof9T9CHuL up: address already in use" id=1159-1```
cause the port picked but the user is actually already used

### PR DESCRIPTION
**Pull request by zaibon**
fixes #556

Before this commit networkd would loose all the reserved port after
a restart and thus publish a wrong list of reserved ports on BCDB.

We now keep this list of reserved port on cache disk so it survive
a restart.

We still do clean up after a reboot since all the network resources are
also going to be re-created so the list of reserved port will be rebuild
from scratch

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -1,0 +1,36 @@
+package cache
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"syscall"
+
+	"github.com/threefoldtech/zos/pkg/storage"
+)
+
+// ForeverDir creates a new cache directory that is stored on the persistent cache.
+// This means data stored in this directory will survive reboot
+// It is the caller's responsibility to remove the directory when no longer needed.
+func ForeverDir(name string) (string, error) {
+	name = filepath.Join(storage.CacheTarget, "cache", name)
+	return name, os.MkdirAll(name, 0700)
+}
+
+// VolatileDir creates a new cache directory that is stored on a tmpfs.
+// This means data stored in this directory will NOT survive a reboot.
+// Use this when you need to store data that needs to survice deamon reboot but not between reboots
+// It is the caller's responsibility to remove the directory when no longer needed.
+// If the directory already exist, the the function does nothing and return successfully
+func VolatileDir(name string, size uint64) (string, error) {
+	const volatileBaseDir = "/var/run/cache"
+	name = filepath.Join(volatileBaseDir, name)
+	if err := os.MkdirAll(volatileBaseDir, 0700); err != nil {
+		return name, err
+	}
+
+	if err := os.Mkdir(name, 0700); err != nil {
+		return "", err
+	}
+	return name, syscall.Mount("", name, "tmpfs", 0, fmt.Sprintf("size=%d", size))
+}
--- a/pkg/network/networker.go
+++ b/pkg/network/networker.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/threefoldtech/zos/pkg/app"
 	"github.com/threefoldtech/zos/pkg/network/ndmz"
 	"github.com/threefoldtech/zos/pkg/network/tuntap"
 
@@ -35,7 +36,8 @@ import (
 
 const (
 	// ZDBIface is the name of the interface used in the 0-db network namespace
-	ZDBIface = "zdb0"
+	ZDBIface  = "zdb0"
+	wgPortDir = "wireguard_ports"
 )
 
 type networker struct {
@@ -46,15 +48,33 @@ type networker struct {
 }
 
 // NewNetworker create a new pkg.Networker that can be used over zbus
-func NewNetworker(identity pkg.IdentityManager, tnodb TNoDB, storageDir string) pkg.Networker {
+func NewNetworker(identity pkg.IdentityManager, tnodb TNoDB, storageDir string) (pkg.Networker, error) {
+
+	wgDir := filepath.Join(storageDir, wgPortDir)
+
+	if app.IsFirstBoot("networkd") {
+		log.Info().Msg("first boot, empty wireguard ports cache")
+		if err := os.RemoveAll(wgDir); err != nil {
+			return nil, err
+		}
+
+		if err := app.MarkBooted("networkd"); err != nil {
+			return nil, errors.Wrap(err, "fail to mark networkd as booted")
+		}
+	}
+
+	if err := os.MkdirAll(wgDir, 0770); err != nil {
+		return nil, err
+	}
+
 	nw := &networker{
 		identity:   identity,
 		storageDir: storageDir,
 		tnodb:      tnodb,
-		portSet:    set.NewUint(),
+		portSet:    set.NewUint(storageDir),
 	}
 
-	return nw
+	return nw, nil
 }
 
 var _ pkg.Networker = (*networker)(nil)
@@ -387,6 +407,12 @@ func (n networker) Addrs(iface string, netns string) ([]net.IP, error) {
 
 // CreateNR implements pkg.Networker interface
 func (n *networker) CreateNR(network pkg.Network) (string, error) {
+	defer func() {
+		if err := n.publishWGPorts(); err != nil {
+			log.Warn().Err(err).Msg("failed to publish wireguard port to BCDB")
+		}
+	}()
+
 	var err error
 	var nodeID = n.identity.NodeID().Identity()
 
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

File: pkg/set/uint.go [hunk 1]
```diff
@@ -18,27 +25,37 @@ func (e ErrConflict) Error() string {
...
 	sync.RWMutex
-	m map[uint]struct{}
+	root string
 }
 
 // NewUint creates a new set for uint
-func NewUint() *UintSet {
+// path if the directory where to store the set on disk
+// the directory pointed by path must exists already
+func NewUint(path string) *UintSet {
```
muhamadazmy, line 34: If this set is created for a location under `/tmp/ports` for example, then we won't need to check for the first-boot status. Since the /tmp folder is gonna be free after new boot anyway. The only concern is free space limitation under /tmp but I think these port files are extremely small anyway. 

Other than that, i think code is okay

muhamadazmy (approved):

pkg/set/uint.go
zaibon: Right, indeed. that would solve the problem...
And actually we could use the same trick there too then: https://github.com/threefoldtech/zos/blob/04ffb3459581917e76eb24cde639e547821bd0e8/pkg/network/ndmz/ndmz.go#L60-L93

What about I implement a package which purpose is to return path that are usable for different use case.
- persistent forever
- persistent between daemon restart

muhamadazmy: Sounds like a good idea to me actually. We just need to make sure either to mount a good enough tmpfs size  for run-time data (/tmp) to be able to hold all temporary data.

pkg/network/networker.go
```diff
@@ -35,26 +39,102 @@ import (
...
+		return nil, fmt.Errorf("failed to create wireguard_port cache directory: %w", err)
+	}
+
+	// create a directory on tmpfs to store the network deployed
+	nwDir, err := cache.VolatileDir(networkDir, 20*mib)
+	if err != nil && !os.IsExist(err) {
+		return nil, fmt.Errorf("failed to create networks cache directory: %w", err)
+	}
+
+	ipamLease, err := cache.VolatileDir(ipamLeaseDir, 10*mib)
```
muhamadazmy: Can't we just create a single VolatileDir for the networker that is like 50M and that's it instead of 3 different mounts?

pkg/cache/cache.go
```diff
@@ -0,0 +1,42 @@
+package cache
```
muhamadazmy, line 1: Did this need a new package? can't we have this under pkg/app ?

muhamadazmy (approved):

pkg/network/networker.go
zaibon: indeed

PR #571, opened 2020-03-12
Fixes leak of persisted resources

### LINKED ISSUE #560: provisiond: leaking of persisted resources
When a node reboot, provisiond makes sure it's local cache (local store) is cleared. This works well for volatile resources like containers, or networks. put for persisted resources, this will cause the resource to not ever get deprovisioned.
- u create a volume
- the machine deploys it
- machines shutsdown
- you mark volume as to-delete
- node starts again
- machine starts with fresh local store, it says resource is not provisioned
- the subvolume remains forever

i think the clean up of local fsstore should deal with volatile resources (like networks and containers) and persisted resources like subvolumes and zdb that survive a reboot

only delete the ones that are volatlie but never delete info about persisted resources

DylanVerstraete:


pkg/provision/local_store.go
```diff
@@ -102,13 +105,34 @@ func NewFSStore(root string) (*FSStore, error) {
...
+			return err
+		}
+		if reservation.Type != VolumeReservation && reservation.Type != ZDBReservation {
+			log.Info().Msgf("Removing %s from cache", path)
+			return os.Remove(path)
+		}
+		return nil
+	})
+	if err != nil {
+		log.Debug().Msgf("error walking the path %q: %v\n", rootPath, err)
```
zaibon:
```suggestion
		log.Error().Msgf("error walking the path %q: %v\n", rootPath, err)
```

pkg/provision/local_store.go
```diff
@@ -102,13 +105,34 @@ func NewFSStore(root string) (*FSStore, error) {
...
 
+func (s *FSStore) removeAllButPersistent(rootPath string) error {
+	err := filepath.Walk(rootPath, func(path string, info os.FileInfo, r error) error {
+		if r != nil {
+			return r
+		}
+		if info.IsDir() {
+			return nil
+		}
+		reservation, err := s.Get(filepath.Base(path))
```
zaibon: Since all you care from the reservation serialized on disk is the type, you can have an inline type that you use inside this function.
Something like
```go
res := struct{
	Type ReservationType `json:"type"`
}{}
```
So the unmarshalling of the data goes faster.

zaibon (changes requested):

pkg/provision/local_store.go
```diff
@@ -262,6 +286,41 @@ func (s *FSStore) Get(id string) (*Reservation, error) {
...
+	validV1 := versioned.MustParseRange(fmt.Sprintf("<=%s", reservationSchemaV1))
+
+	if validV1(reader.Version()) {
+		if err := json.NewDecoder(reader).Decode(&res); err != nil {
+			return nil, err
+		}
+	} else {
+		return nil, fmt.Errorf("unknown reservation object version (%s)", reader.Version())
+	}
+	return &res.Type, nil
```
muhamadazmy: Why u return a pointer , not res.Type ?

pkg/provision/local_store.go
```diff
@@ -262,6 +286,41 @@ func (s *FSStore) Get(id string) (*Reservation, error) {
 	return s.get(id)
 }
 
+// GetType retrieves a specific reservation's type using its ID
+// if returns a non nil error if the reservation is not present in the store
+func (s *FSStore) GetType(id string) (ReservationType, error) {
```
zaibon: this can be a private function I don't think this method is used anywhere else then in this package.

zaibon (changes requested):

zaibon (approved):

PR #572, opened 2020-03-13
Decomission workload

zaibon:


pkg/provision/provision.go
```diff
@@ -58,4 +57,21 @@ var (
...
+}
+
+// NewErrTemporary wrap an error and mark it as temporary
+func NewErrTemporary(err error) error {
+	return ErrTemporary{err: err}
+}
+
+// Error implements the errors.Error interface
+func (e ErrTemporary) Error() string {
+	return e.Error()
```
muhamadazmy: This will cause a stack overflow.

pkg/gedis/commands_provision.go
```diff
@@ -218,6 +218,7 @@ func reservationFromSchema(w ptypes.TfgridReservationWorkload1) (*provision.Rese
 		Signature: []byte(w.Signature),
 		Data:      w.Workload,
 		Tag:       provision.Tag{"source": "BCDB"},
+		ToDelete:  w.ToDelete,
```
muhamadazmy, line 221: I had similar fix in plan-bee :)

muhamadazmy (changes requested):

pkg/provision/provision.go
zaibon: wow nice catch, should be `e.err.Error()`.

pkg/provision/provision.go
```diff
@@ -58,4 +57,26 @@ var (
...
+}
+
+// Error implements the errors.Error interface
+func (e ErrTemporary) Error() string {
+	return e.err.Error()
+}
+
+// Is implements errors.Is interface
+func (e ErrTemporary) Is(target error) bool {
+	return target == ErrTemporary{}
```
muhamadazmy: This won't work because that's a check on the value, not the type. which means target will only == ErrTemprary{} if and only if traget was created as NewErrTemporary(nil) 

I suggest changing that to 
```go
_, ok := target.(ErrTemporary)
return true
```

PR #578, opened 2020-03-16
container: support user defined logs endpoint

maxux:
This PR allows user to forward container stdout/stderr to external endpoint. Right now, only redis is supported. You can define on container reservation a new field `logs` which contains information where to push logs:
```
        "logs": [
          {
            "type": "redis",
            "data": {
              "endpoint": "redis://10.241.0.232:6370",
              "channel": "debug"
            }
          }
        ],
```

This endpoint is a reachable redis, and channel is a pub/sub where logs will be sent (one line at a time).


pkg/logger/logger_redis.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Stderr handle a stderr single line
+func (c *ContainerLoggerRedis) Stderr(line string) error {
+	_, err := c.conn.Do("PUBLISH", c.ChannelStderr, line)
```
muhamadazmy: Shouldn't we have a different channel for stderr or at least prefix the line with "E:" the same way u do with console logger?

tools/bcdb_mock/models/generated/workloads/workloads.go
```diff
@@ -93,9 +93,20 @@ type TfgridWorkloadsReservationContainer1 struct {
 	Volumes           []TfgridWorkloadsReservationContainerMount1    `bson:"volumes" json:"volumes"`
 	NetworkConnection []TfgridWorkloadsReservationNetworkConnection1 `bson:"network_connection" json:"network_connection"`
 	StatsAggregator   []TfgridWorkloadsReservationStatsaggregator1   `bson:"stats_aggregator" json:"stats_aggregator"`
+	Logs              []TfgridWorkloadsReservationLogs1              `bson:"logs" json:"logs"`
 	FarmerTid         int64                                          `bson:"farmer_tid" json:"farmer_tid"`
 }
 
+type TfgridWorkloadsReservationLogs1 struct {
```
muhamadazmy, line 100: Can we add the corresponding schema structures to the files under `models/schema/workloads`. No regeneration is needed, just to be able to port this back to jumpscale.

muhamadazmy (changes requested):
Excellent PR. but I have one comment that requires a change (regarding schema)

pkg/container/container.go
```diff
@@ -173,7 +173,39 @@ func (c *containerModule) Run(ns string, data pkg.Container) (id pkg.ContainerID
...
+	filelog, err := logger.NewContainerLoggerFile(filepath)
+	if err != nil {
+		return id, err
+	}
+
+	loggers.Add(filelog)
+
+	// set user defined endpoint logging
+	for _, l := range data.Logs {
+		if l.Type == "redis" {
```
zaibon: make "redis" a constant in the logger package

zaibon: you probably want to use a switch here, so where there is new type of logger we can just add a case

pkg/logger/logger.go
```diff
@@ -0,0 +1,28 @@
...
+// ContainerLogger defines a logging backend
+type ContainerLogger interface {
+	// Send stdout line to the backend
+	Stdout(line string) error
+
+	// Send stderr line to the backend
+	Stderr(line string) error
+
+	// Close stdout handler
+	CloseStdout()
```
zaibon: Is there any reason you want to have a separate method to close stdout and stderr ? We will always want to close both stream no ?

pkg/container/container.go
zaibon: make redis a constant in the logger package

zaibon: you probably want to use a switch here, so where there is new type of logger we can just add a case

pkg/logger/logger_console.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"fmt"
+	"github.com/rs/zerolog/log"
+	"os"
+)
+
+// ContainerLoggerConsole does nothing else that print
+// logs on console stdout/stderr, there are no config
+type ContainerLoggerConsole struct {
+	ContainerLogger
```
zaibon: you do not need to embed the interface since you type already implements the required method to satisfy the interface

pkg/logger/logger_file.go
```diff
@@ -0,0 +1,60 @@
...
+	fd       *os.File
+}
+
+// NewContainerLoggerFile open file and prepare logs writing
+func NewContainerLoggerFile(filepath string) (*ContainerLoggerFile, error) {
+	log.Debug().Str("filepath", filepath).Msg("initializing localfile logging")
+
+	f, err := os.Create(filepath)
+	if err != nil {
+		return &ContainerLoggerFile{}, err
```
zaibon:
```suggestion
		return nil, err
```

pkg/logger/logger_file.go
```diff
@@ -0,0 +1,60 @@
...
+
+	return &ContainerLoggerFile{
+		Filepath: filepath,
+		fd:       f,
+	}, nil
+}
+
+// Stdout handle a stdout single line
+func (c *ContainerLoggerFile) Stdout(line string) error {
+	_, err := c.fd.WriteString(line + "\n")
```
zaibon:
```suggestion
	_, err := fmt.Fprintf(c.fd, "%s\n", line)
```

pkg/logger/logger_file.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Stderr handle a stderr single line
+func (c *ContainerLoggerFile) Stderr(line string) error {
+	_, err := c.fd.WriteString(line + "\n")
```
zaibon:
```suggestion
      _, err := fmt.Fprintf(c.fd, "%s\n", line)
```

pkg/logger/logger.go
zaibon: I would also return an error here. So we can at least log the error if any.

pkg/logger/logger_redis.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"github.com/gomodule/redigo/redis"
+	"github.com/rs/zerolog/log"
+)
+
+// ContainerLoggerRedis send stdout/stderr to a
+// Redis PubSub channel
+type ContainerLoggerRedis struct {
+	ContainerLogger
```
zaibon: no need to embed the interface

pkg/logger/logger_redis.go
```diff
@@ -0,0 +1,65 @@
...
+
+	Endpoint      string
+	ChannelStdout string
+	ChannelStderr string
```
zaibon: These property doesn't have to be public
```suggestion
	endpoint      string 
  	channelStdout string
	channelStderr string
```

pkg/logger/logger_file.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/rs/zerolog/log"
+	"os"
+)
+
+// ContainerLoggerFile write stdout/stderr to
+// a defined file
+type ContainerLoggerFile struct {
+	ContainerLogger
+
+	Filepath string
```
zaibon:
```suggestion
	filepath string
```

pkg/logger/loggers.go
```diff
@@ -0,0 +1,102 @@
...
+	}, nil
+}
+
+// Add adds a defined backend on the list
+func (c *ContainerLoggers) Add(backend ContainerLogger) {
+	c.loggers = append(c.loggers, backend)
+}
+
+// Wait waits for logs to be closed to clean handlers
+func (c *ContainerLoggers) Wait() {
```
zaibon: you do not need this method at all. You can just copy the content if this method into the literal function line 96

tools/tfuser/main.go
```diff
@@ -244,6 +244,15 @@ func main() {
 							Name:  "public6",
 							Usage: "when enabled, the container will have a public IPv6 interface",
 						},
+						cli.StringFlag{
+							Name:  "logs",
+							Usage: "custom redis endpoint to send process logs (eg: redis://example.tld:1234)",
+						},
+						cli.StringFlag{
```
zaibon, line 251: I guess we could get rid of this flag if we use the `path` of the redis URI. Something like
`redis://host:port/channel`

zaibon (changes requested):
the PR overall looks ok. 
I think there is a way to make things a bit simpler though.
Instead of adding a new ContainerLogger interface, we could just use the [WithStream](https://github.com/containerd/containerd/blob/master/cio/io.go#L117) method. 
This methods takes any `io.Writer`. This means any new logger implementation should just implement the `io.Writer` interface. 

The result would be something like
```go
mw := io.MultiWriter(
	logFile,
	redisLogger,
       // any other object that implement io.Writer...
)
container.NewTask(ctx,cio.NewCreator(cio.WithStreams(nil, mw, mw)))
```

pkg/logger/logger.go
maxux: Since it's 2 different handler for reading them, I keep both for closing. I guess they will always close at the same time, but nothing guarantee it. And for redis, internally we support two two channels for stdout/stderr, we could also support two files for stdout/stderr, which are not the same, etc. I guess it's better to keep separation ?

pkg/logger/logger_console.go
maxux: I can have an empty struct ?

pkg/logger/logger_redis.go
maxux: We support a different channel, it's just not exposed publicly on the reservation.

tools/tfuser/main.go
maxux: I don't know if channel can be specified like that. I'll check log. I maybe need to parse uri and connect not using internal uri connector.

maxux:
Latest implementation use the `MultiWriter` that @zaibon suggested, reducing amount of code and complexity. Thanks :)

Still some improvement to fix and everything will be updated.

tools/tfuser/main.go
maxux: It's implemented like that now :)

pkg/logger/logger.go
maxux: Handled by io.Writer now

pkg/logger/logger_redis.go
maxux: Now you can specify different channel for `stdout` and `stderr`

tools/bcdb_mock/models/generated/workloads/workloads.go
maxux: Please review my changes to see if models seems right :)

maxux:
From 0d6d56f commit message:

- Move `pkg/logger` to `pkg/container/logger` since logger package are only for container purpose
- Update name of struct inside logger package for shorter names
  - `ContainerLoggerConsole` -> `Console`
  - `ContainerLoggerFile` -> `File`
  - `ContainerLoggerRedis` -> `Redis`
  - `NewContainerLoggers` -> `NewLoggers`
  - `NewContainerLoggerFile` -> `NewFile`
  - `NewContainerLoggerRedis` -> `NewRedis`
  - `NewContainerLoggerConsole` -> `NewConsole`
- Update provision to replace endpoint/channel with redis url for stdout and stderr
- Update tfuser to support new provision fields
  - Removing `--logs` and `--channel`
  - Adding `--stdout` and `--stderr` options
- Argument are now always using `redis://host:port/channel` format

zaibon (approved):

PR #584, opened 2020-03-18
Fixes empty reservation chache bug

### LINKED ISSUE #581: provisiond: lstat /var/cache/modules/provisiond/reservations
When booting a VM, provisiond is stuck on:
```
provisiond: lstat /var/cache/modules/provisiond/reservations: no such file or directory
```

Workaround: `mkdir /var/cache/modules/provisiond/reservations` fixes the issue
This occures on master, even with f41cfc241 reverted

DylanVerstraete:
issue: https://github.com/threefoldtech/zos/issues/581

### DIFF
--- a/pkg/provision/local_store.go
+++ b/pkg/provision/local_store.go
@@ -109,7 +109,13 @@ func NewFSStore(root string) (*FSStore, error) {
 }
 
 func (s *FSStore) removeAllButPersistent(rootPath string) error {
-	err := filepath.Walk(rootPath, func(path string, info os.FileInfo, r error) error {
+	// if rootPath is not present on the filesystem, return
+	_, err := os.Stat(rootPath)
+	if err != nil {
+		return nil
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/provision/local_store.go
```diff
@@ -109,7 +109,13 @@ func NewFSStore(root string) (*FSStore, error) {
 }
 
 func (s *FSStore) removeAllButPersistent(rootPath string) error {
-	err := filepath.Walk(rootPath, func(path string, info os.FileInfo, r error) error {
+	// if rootPath is not present on the filesystem, return
+	_, err := os.Stat(rootPath)
```
muhamadazmy, line 113: This fix is not very useful, since you still return an error if the rootPath does not exist.
A correct solution would be 

```suggestion
	_, err := os.Stat(rootPath)
	if os.IsNotExist(err) {
	   return nil
	} else if err != nil {
	   return err
	}
```

So basically we return nil if the path does not exist.

zaibon (approved):

PR #586, opened 2020-03-18
Fixes debug mode logs

### LINKED ISSUE #582: provisiond: enable debug by default when node is running in dev
When node is running in dev mode, debug logs should be enabled by default for all daemon. Right now, only `provisiond` is reducing verbosity by default.

DylanVerstraete:
fixes https://github.com/threefoldtech/zos/issues/582

Sets logger to debug on a global level if debug env is activated.


zaibon:
This requires more thinking on a how to deals with logs in general.

PR #587, opened 2020-03-18
Adds error message that farmer id is not valid

### LINKED ISSUE #580: unclear error when using wrong farmer id
As I didn't know that I couldn't use a random farmer ID anymore, I've had some troubles starting my Qemu vm. 

This is the current error message: 
`[+] identityd: 2020-03-17T13:29:56Z info start registration of the node version="not booted from flist"
[+] identityd: 2020-03-17T13:29:56Z error  error="wrong response status code received: 500 Internal Server Error"
[+] identityd: 2020-03-17T13:29:56Z warn registration failed error="wrong response status code received: 500 Internal Server Error" sleep=46.982049071s
`
Should we add a better error message telling the farmer_id is not correct?

PS: Thank you very much @muhamadazmy and @LeeSmet for the help!

DylanVerstraete:
closes: https://github.com/threefoldtech/zos/issues/580

### DIFF
--- a/cmds/identityd/main.go
+++ b/cmds/identityd/main.go
@@ -461,7 +461,7 @@ func registerNode(nodeID pkg.Identifier, farmID pkg.FarmID, version string, stor
 
 	_, err := store.RegisterNode(nodeID, farmID, version, loc)
 	if err != nil {
-		log.Error().Err(err).Send()
+		log.Error().Msgf("%v, farmID: %d is not valid", err, farmID)
 		return err
 	}
 	return nil


cmds/identityd/main.go
```diff
@@ -461,7 +461,7 @@ func registerNode(nodeID pkg.Identifier, farmID pkg.FarmID, version string, stor
 
 	_, err := store.RegisterNode(nodeID, farmID, version, loc)
 	if err != nil {
-		log.Error().Err(err).Send()
+		log.Error().Msgf("%v, farmID: %d is not valid", err, farmID)
```
zaibon, line 464: you do not know if the error mean the farmer ID is invalid. It could be anything. 
What assumption make you think this is the only possible reason for the error ?

DylanVerstraete: True, maybe the API should return a description of the error.

PR #589, opened 2020-03-18
bdcb_mock: make proofs opt-in

grimpy:
To get nodes with proofs you need to explicitly request it in the client


zaibon:
implemented in https://github.com/threefoldtech/zos/commit/5d5b4be8d7bf97f4989bd1f30e99ebe70180e00f instead

PR #598, opened 2020-03-20
boostrap: don't use error log when skipping symlink

### LINKED ISSUE #596: we should really remove these non-errors
People go bananas over them

image

zaibon:
fixes #596


bootstrap/bootstrap/src/zfs.rs
```diff
@@ -75,7 +75,7 @@ impl Zfs {
                 debug!("installing file {:?}", dst);
                 std::fs::copy(&src, &dst)?;
             } else {
-                error!("unsupported file type: ({:?}): {:?}", src, typ)
+                error!("skipping: ({:?}): {:?}", src, typ)
```
muhamadazmy: May be also make it a `debug!` message

PR #602, opened 2020-03-22
bcdb: authenticate request to the API

zaibon:


tools/bcdb_mock/mw/auth.go
```diff
@@ -0,0 +1,103 @@
...
+	uid, err := strconv.ParseInt(id, 10, 64)
+	if err != nil {
+		return nil
+	}
+
+	f := types.UserFilter{}
+
+	f = f.WithID(schema.ID(uid))
+
+	col := m.db.Collection(types.UserCollection)
```
muhamadazmy: Why not `f.Get()` ?

tools/bcdb_mock/mw/auth.go
```diff
@@ -0,0 +1,103 @@
...
+
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+type mongoKeyGetter struct {
+	db *mongo.Database
+}
+
+func (m mongoKeyGetter) GetKey(id string) interface{} {
```
muhamadazmy: Shouldn't this be different for different endpoints. for example, the node endpoints where the node initialize the code should check against the Node public key, not the user public key

tools/bcdb_mock/mw/auth.go
```diff
@@ -0,0 +1,103 @@
...
+
+	challenge := "Signature"
+	if len(challengeParams) > 0 {
+		challenge += fmt.Sprintf(" %s", strings.Join(challengeParams, ", "))
+	}
+
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		err := v.Verify(req)
+		if err != nil {
+			w.Header()["WWW-Authenticate"] = []string{challenge}
```
muhamadazmy: I don't get this actually, does this mean that the user (or the node) has to do 2 trips to execute an action. First one to get Unauthorized + the header to sign, then a request with the correct signature ?

tools/bcdb_mock/mw/auth.go
zaibon: You are correct.

tools/bcdb_mock/mw/auth.go
zaibon: Well if you send a request with the proper header right away then you will not get the 401 error.
Here we just notify the client that if he wants to authenticate, it needs to use these header to create the signature.

muhamadazmy: Yes, I understand but if the client will not know what are the required headers unless at least one call is made. right? also does the challenge change with every request?

tools/bcdb_mock/pkg/directory/farms_handlers.go
```diff
@@ -90,3 +91,36 @@ func (s *FarmAPI) getFarm(r *http.Request) (interface{}, mw.Response) {
 
 	return farm, nil
 }
+
+func (s *FarmAPI) delete(r *http.Request) (interface{}, mw.Response) {
```
muhamadazmy: Why do we need that?
I mean in block chain database, the "delete" should not be allowed (or at least the data should not get deleted from the database, may be only marked as deleted)

tools/bcdb_mock/pkg/directory/node_handlers.go
```diff
@@ -185,21 +236,70 @@ func (s *NodeAPI) updateUptimeHandler(r *http.Request) (interface{}, mw.Response
...
-
 	db := mw.Database(r)
 	if err := s.updateReservedCapacity(r.Context(), db, nodeID, resources); err != nil {
 		return nil, mw.NotFound(err)
 	}
 
 	return nil, nil
 }
+
+func (s *NodeAPI) delete(r *http.Request) (interface{}, mw.Response) {
```
muhamadazmy: Same question regarding farme.delete()

muhamadazmy (approved):
This looks really good, I approved but I had some questions that are not blocking

tools/bcdb_mock/pkg/directory/farms_handlers.go
zaibon: You are right I should not delete anything but use another mechanism to marked them as deleted. I'll remove from this PR and open an issue to discuss how to handle clean up in a better way.

tools/bcdb_mock/mw/auth.go
zaibon: The challenge doesn't not change with every request no. And this needs to be documented. Normal user will always include this during each requests.

PR #614, opened 2020-03-23
Fixes dangling files in provision cache

### LINKED ISSUE #611: provisiond: is emtpy reservation cached file are found provisiond fails to start

DylanVerstraete:


pkg/provision/local_store.go
```diff
@@ -121,6 +122,14 @@ func (s *FSStore) removeAllButPersistent(rootPath string) error {
 		if r != nil {
 			return r
 		}
+		// if a file with size 0 is present we can assume its empty and remove it
+		if info.Size() == 0 {
+			err := os.Remove(path)
+			if err != nil {
```
zaibon: Seems in both case you do need to return, so you can just do
```go
return os.Remove(path)
```

zaibon (changes requested):

zaibon (approved):

PR #622, opened 2020-03-24
output clearer error message when farm ID is wrong

### LINKED ISSUE #580: unclear error when using wrong farmer id

zaibon:
we handle case when farm ID has a wronf formaat or when the farm ID does
not exists

Here are the new error message outputted by identityd:
```
[+] identityd: 2020-03-24T16:14:28Z fatal failed to create identity manager error="failed to parse node environment: wrong format for farm ID: strconv.ParseUint: parsing \"123,123\": invalid syntax"
```
```
[+] identityd: 2020-03-24T16:11:31Z error  error="farmer with ID 123123123 does not exists"
```

fixes #580

### DIFF
--- a/pkg/environment/environment.go
+++ b/pkg/environment/environment.go
@@ -4,6 +4,7 @@ import (
 	"os"
 	"strconv"
 
+	"github.com/pkg/errors"
 	"github.com/threefoldtech/zos/pkg"
 
 	"github.com/threefoldtech/zos/pkg/kernel"
@@ -130,7 +131,7 @@ func getEnvironmentFromParams(params kernel.Params) (Environment, error) {
 		env.Orphan = false
 		id, err := strconv.ParseUint(farmerID[0], 10, 64)
 		if err != nil {
-			return env, err
+			return env, errors.Wrap(err, "wrong format for farm ID")
 		}
 		env.FarmerID = pkg.FarmID(id)
 	}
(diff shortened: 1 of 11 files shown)


pkg/identity/store.go
```diff
@@ -62,6 +62,10 @@ func (s *httpIDStore) RegisterNode(node pkg.Identifier, farm pkg.FarmID, version
 	}
 	defer resp.Body.Close()
 
+	if resp.StatusCode == http.StatusNotFound {
```
muhamadazmy: I think one of the issues is that bcdb returns 500 if farmer id is not found, which is also need to be fixed in the mock to return proper error code

zaibon: indeed, and this is also part of this PR: https://github.com/threefoldtech/zos/pull/622/files#diff-cd866ca7f60c9a054ceee3a4b934b6b2R38

PR #623, opened 2020-03-25
Feature payment flow

### LINKED ISSUE #605: explorer: implements escrow and payment flow management
```
1. user register a reservation
2. for each user-farmer pair (ufp), creates an address
3. user send tokens to each ufp
4. explorer sees the tokens have arrived
5. explorer verifies the amount of token is correct
5.1 if no, refund to the payer
5.2 if yes, explorer set the reservation state to DEPLOY
6. nodes actually provision
7. once all the workloads result are deployed, release money to the farmer
7.1 if one of the workload result is ERROR, explorer cancel reservation, money if refunded to payers


for later:
- allow price per nodes
- allow more complex payment requirements
```

DylanVerstraete:


tools/bcdb_mock/pkg/escrow/escrow.go
```diff
@@ -0,0 +1,332 @@
...
+	reservationEscrows, err := types.GetAllActiveReservationPaymentInfos(ctx, e.db)
+	if err != nil {
+		return errors.Wrap(err, "failed to load active reservations from escrow")
+	}
+	for _, escrowInfo := range reservationEscrows {
+		allPaid := true
+		for _, escrowAccount := range escrowInfo.Infos {
+			balance, _, err := e.wallet.GetBalance(escrowAccount.EscrowAddress, escrowInfo.ReservationID)
+			if err != nil {
+				allPaid = false
```
muhamadazmy: What is the point of setting this if you going to return an error anyway

tools/bcdb_mock/pkg/escrow/escrow.go
```diff
@@ -0,0 +1,332 @@
...
+				return errors.Wrap(err, "failed to verify escrow account balance")
+			}
+			if balance < escrowAccount.TotalAmount {
+				allPaid = false
+				return errors.Wrapf(err, "escrow account %s for reservation id %d is not funded yet", escrowAccount.EscrowAddress, escrowInfo.ReservationID)
+			}
+		}
+		if allPaid {
```
muhamadazmy: This condition doesn't make sense because if allPaid != true, the function would return already and never reach this block. hence the check can be dropped

tools/bcdb_mock/pkg/escrow/resource.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	cloudUnits struct {
+		// TODO: should we use big.Float here to avoid floating point errors?
+		cu float64
+		su float64
+	}
+)
+
+// cost price of cloud units:
```
muhamadazmy: I am not sure, but shouldn't this be configurable

tools/bcdb_mock/pkg/escrow/types/escrow.go
```diff
@@ -0,0 +1,137 @@
...
+	paymentInfos := make([]ReservationPaymentInformation, 0)
+	err = cursor.All(ctx, &paymentInfos)
+	if err != nil {
+		err = errors.Wrap(err, "failed to decode active payment information")
+	}
+	return paymentInfos, err
+}
+
+// GetAllAddresses gets all in use addresses from the escrowcollections
+func GetAllAddresses(ctx context.Context, db *mongo.Database) ([]string, error) {
```
muhamadazmy: This probably requires pagination otherwise this can blow up since it doesn't use any filtering. I assume this collection can has so many addresses

tools/bcdb_mock/pkg/workloads/types/reservation.go
```diff
@@ -394,6 +394,17 @@ func ReservationCreate(ctx context.Context, db *mongo.Database, r Reservation) (
 	return id, nil
 }
 
+// ReservationGetByID gets a reservation by id
+func ReservationGetByID(ctx context.Context, db *mongo.Database, id schema.ID) (Reservation, error) {
+	var filter ReservationFilter
+	filter = filter.WithID(id)
+
+	var reservation Reservation
+	res := db.Collection(ReservationCollection).FindOne(ctx, filter)
```
muhamadazmy: Why not `filter.Get` ?

tools/bcdb_mock/pkg/escrow/escrow.go
```diff
@@ -0,0 +1,332 @@
...
+
+// RegisterReservation registers a workload reservation
+func (e *Escrow) RegisterReservation(reservation workloads.Reservation) ([]types.EscrowDetail, error) {
+	job := reservationRegisterJob{
+		reservation:  reservation,
+		responseChan: make(chan reservationRegisterJobResponse),
+	}
+	e.reservationChannel <- job
+
+	response := <-job.responseChan
```
muhamadazmy: If this is going to block anyway on the response, why is it needed to do it over channel. Can't it be something like 

```go
job.Process()
```

zaibon (approved):

PR #625, opened 2020-03-25
Type cleanup

### LINKED ISSUE #600: Clean up duplicated types in the code base.

muhamadazmy:
To review this PR please follow this order:
- Review the client code (all new) under tools/client 
- Gedis types and commands has been dropped (probably commands will be reintroduced later again, but under client)
- Check changes in daemons to use the new client instead of the scattered sotres
- Type converts has been changed to use types from mock


tools/client/client.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/threefoldtech/zos/pkg/capacity"
+	"github.com/threefoldtech/zos/pkg/capacity/dmi"
+	"github.com/threefoldtech/zos/pkg/schema"
+	"github.com/threefoldtech/zos/tools/bcdb_mock/models/generated/directory"
+	"github.com/threefoldtech/zos/tools/bcdb_mock/models/generated/phonebook"
+	"github.com/threefoldtech/zos/tools/bcdb_mock/models/generated/workloads"
+)
+
+// Client structure
+type Client struct {
```
zaibon, line 16: Should we split the client into collection types instead of pacakge ? 
So instead of having `client.Directory.NodeGet()` have `client.Nodes.Get()` ?

tools/client/client.go
```diff
@@ -0,0 +1,115 @@
...
+	NodeSetInterfaces(id string, ifaces []directory.TfgridDirectoryNodeIface1) error
+	NodeSetPorts(id string, ports []uint) error
+	NodeSetPublic(id string, pub directory.TfgridDirectoryNodePublicIface1) error
+
+	//TODO: this method call uses types from zos that is not generated
+	//from the schema. Which is wrong imho.
+	NodeSetCapacity(
+		id string,
+		resources directory.TfgridDirectoryNodeResourceAmount1,
+		dmiInfo dmi.DMI,
```
zaibon, line 41: I think the reason for this is this `dmi.DMI` type is impossible to map with jsx schema.
It is also a type ever used by JS at all. So I do not mind too much.

tools/client/http.go
```diff
@@ -0,0 +1,145 @@
...
+	cl http.Client
+}
+
+func newHTTPClient(raw string) (*httpClient, error) {
+	u, err := url.Parse(raw)
+	if err != nil {
+		return nil, errors.Wrap(err, "invalid url")
+	}
+
+	return &httpClient{u: u}, nil
```
zaibon, line 37: Now that we do not use the http.DefaultClient anymore, let's also use sane value for the [Transport type](https://pkg.go.dev/net/http?tab=doc#Transport)

tools/client/http.go
```diff
@@ -0,0 +1,145 @@
...
+		if err := dec.Decode(&output); err != nil {
+			return errors.Wrapf(err, "failed to load error while processing invalid return code of: %s", response.StatusCode)
+		}
+
+		return fmt.Errorf("%s: %s", response.Status, output.E)
+	}
+
+	if output == nil {
+		//discard output
+		return nil
```
zaibon, line 79:
```suggestion
		output = ioutil.Discard
```
So we are sure to consume eveything from the body in any case and allow to reuse the connection ?

tools/client/http.go
```diff
@@ -0,0 +1,145 @@
...
+	return nil
+}
+
+func (c *httpClient) post(u string, input interface{}, output interface{}, expect ...int) error {
+	var buf bytes.Buffer
+	if err := json.NewEncoder(&buf).Encode(input); err != nil {
+		return errors.Wrap(err, "failed to serialize request body")
+	}
+
+	response, err := http.Post(u, httpContentType, &buf)
```
zaibon:
```suggestion
	response, err := c.cl.Post(u, httpContentType, &buf)
```

tools/client/http.go
```diff
@@ -0,0 +1,145 @@
...
+func (c *httpClient) put(u string, input interface{}, output interface{}, expect ...int) error {
+	var buf bytes.Buffer
+	if err := json.NewEncoder(&buf).Encode(input); err != nil {
+		return errors.Wrap(err, "failed to serialize request body")
+	}
+	req, err := http.NewRequest(http.MethodPut, u, &buf)
+	if err != nil {
+		return errors.Wrap(err, "failed to build request")
+	}
+	response, err := http.DefaultClient.Do(req)
```
zaibon:
```suggestion
	response, err := c.cl.Do(req)
```

tools/client/http.go
```diff
@@ -0,0 +1,145 @@
...
+
+	return c.process(response, output, expect...)
+}
+
+func (c *httpClient) get(u string, query url.Values, output interface{}, expect ...int) error {
+	if len(query) > 0 {
+		u = fmt.Sprintf("%s?%s", u, query.Encode())
+	}
+
+	response, err := http.Get(u)
```
zaibon:
```suggestion
	response, err := c.cl.Get(u)
```

tools/client/http.go
```diff
@@ -0,0 +1,145 @@
...
+
+func (c *httpClient) delete(u string, query url.Values, output interface{}, expect ...int) error {
+	if len(query) > 0 {
+		u = fmt.Sprintf("%s?%s", u, query.Encode())
+	}
+	req, err := http.NewRequest(http.MethodDelete, u, nil)
+	if err != nil {
+		return errors.Wrap(err, "failed to build request")
+	}
+	response, err := http.DefaultClient.Do(req)
```
zaibon:
```suggestion
	response, err := c.cl.Do(req)
```

tools/client/signer.go
```diff
@@ -0,0 +1,90 @@
...
+		return "", "", err
+	}
+
+	return hex.EncodeToString(msg), hex.EncodeToString(sig), nil
+}
+
+// Sign constructs a message from all it's arguments then sign it
+func (s *Signer) Sign(o ...interface{}) ([]byte, []byte, error) {
+	var buf bytes.Buffer
+	for _, x := range o {
```
zaibon, line 52: How does a range on an empty interface works ?
It walk over all the fields of the underlying type ?

cmds/identityd/main.go
```diff
@@ -436,22 +438,40 @@ func identityMgr(root string) (pkg.IdentityManager, error) {
...
+
+	return cl.Directory, nil
 }
 
-func registerNode(nodeID pkg.Identifier, farmID pkg.FarmID, version string, store identity.IDStore, loc geoip.Location) error {
+func registerNode(nodeID pkg.Identifier, farmID pkg.FarmID, version string, store client.Directory, loc geoip.Location) error {
 	log.Info().Str("version", version).Msg("start registration of the node")
 
-	_, err := store.RegisterNode(nodeID, farmID, version, loc)
+	err := store.NodeRegister(directory.TfgridDirectoryNode2{
```
zaibon: I think we miss something that set the NodeIdV1 field

pkg/provision/converter.go
```diff
@@ -8,17 +8,24 @@ import (
...
-	generated "github.com/threefoldtech/zos/pkg/gedis/types/provision"
 	"github.com/threefoldtech/zos/pkg/network/types"
+	"github.com/threefoldtech/zos/tools/bcdb_mock/models/generated/workloads"
 )
 
 // ContainerToProvisionType converts TfgridReservationContainer1 to Container
-func ContainerToProvisionType(c generated.TfgridReservationContainer1) (Container, string, error) {
+func ContainerToProvisionType(c workloads.TfgridWorkloadsReservationContainer1) (Container, string, error) {
+	env := func(m map[string]interface{}) map[string]string {
+		o := make(map[string]string)
```
zaibon:
```suggestion
		o := make(map[string]string,len(m))
```
So we avoid to allocated potentially multiple time during the loop

pkg/provision/converter_test.go
```diff
@@ -38,12 +38,12 @@ func TestTfgridReservationContainer1_ToProvisionType(t *testing.T) {
 		NodeID            string
 		Flist             string
 		HubURL            string
-		Environment       map[string]string
+		Environment       map[string]interface{}
```
zaibon: Why this type change ? environment has always value as string

pkg/provision/converter_test.go
```diff
@@ -130,13 +132,20 @@ func TestTfgridReservationContainer1_ToProvisionType(t *testing.T) {
 			wantErr: false,
 		},
 	}
+	// convertEnv := func(m map[string]string) map[string]interface{
```
zaibon: clean up commented code

pkg/provision/engine.go
```diff
@@ -36,7 +38,7 @@ type defaultEngine struct {
 	nodeID string
 	source ReservationSource
 	store  ReservationCache
-	fb     Feedbacker
+	cl     *client.Client
```
zaibon, line 41: Isn't this a bit sad that we loose the interface here ?

tools/bcdb_mock/models/schema/workloads/tfgrid_workloads_reservation_container_1.toml
```diff
@@ -33,6 +33,7 @@ mountpoint = (S)
 @url = tfgrid.workloads.reservation.network.connection.1
 network_id = (S)
 ipaddress = (ipaddress)
+public_ip6 = (B)
```
zaibon, line 37: We're also missing the capacity type here:
```
@url = tfgrid.workloads.reservation.container.capacity.1
# Number of vCPU
cpu = (I) 
# memory in MiB
memory = (I)
```

zaibon (changes requested):
First round of review.
Can we also try to get rid of the weird type prefix `TfgridReservation...` we have on all the types ?
It really makes the code java like...

tools/client/client.go
muhamadazmy: I thought of that, but I thought may be it's better if we have the client per mock "pkg" so directory, phonebook and workloads, otherwise I will have to always pass around 2 (or more) insntance of the interfaces 
for example, modules that need to work against reservations will usually need workloads and reservations, etc ... so it's cleaner to have related stuff grouped in one interface

tools/client/http.go
muhamadazmy: I don't get what u mean here!

tools/client/http.go
muhamadazmy: Idd, but this thin should be ioutil.ReadAll(response.Body)

tools/client/http.go
muhamadazmy: I think i had a split brain at some point in time :D

pkg/provision/converter_test.go
muhamadazmy: I know, but because the schema limitation, we can only generate map[string]interface{} (there is no way in schema to define type of the value of the map!)

Note: Actually i support that in our schema pkg, but it's not compatible with jsx

pkg/provision/engine.go
muhamadazmy: It is, but we also need 2 different interfaces here. so It's either we pass 2 instances of interfaces (instead of one) or just pass one client

tools/bcdb_mock/models/schema/workloads/tfgrid_workloads_reservation_container_1.toml
muhamadazmy: where is that coming from?

tools/client/http.go
zaibon: well you already consume the body in the `dec.Decode` underneath, but this you are right better going it like you propose

pkg/provision/engine.go
zaibon: Or we just define a Interface that embed the 2 interface we need ?

tools/bcdb_mock/models/schema/workloads/tfgrid_workloads_reservation_container_1.toml
zaibon: From here https://github.com/threefoldtech/zos/blob/master/pkg/provision/container.go#L68

cmds/identityd/main.go
```diff
@@ -436,22 +438,40 @@ func identityMgr(root string) (pkg.IdentityManager, error) {
...
+	return cl.Directory, nil
 }
 
-func registerNode(nodeID pkg.Identifier, farmID pkg.FarmID, version string, store identity.IDStore, loc geoip.Location) error {
+func registerNode(nodeID pkg.Identifier, farmID pkg.FarmID, version string, store client.Directory, loc geoip.Location) error {
 	log.Info().Str("version", version).Msg("start registration of the node")
 
-	_, err := store.RegisterNode(nodeID, farmID, version, loc)
+	err := store.NodeRegister(directory.Node{
+		NodeId:    nodeID.Identity(),
```
zaibon, line 462: Still missing the NodeIDV1

pkg/provision/engine.go
muhamadazmy: I don't see the added benefit of that, that would be an interface with so many methods that we don't use. 
I mean if the provision engine requires a client to the explorer, i don't see this as an odd code. Plus the client itself, does not provide any functionality, it's just a carrier for the actual interfaces. I would just leave it like that

zaibon (changes requested):

zaibon (approved):

PR #636, opened 2020-03-29
explorer-client: Update listing for farm and reservation

grimpy:

### DIFF
--- a/tools/client/workloads.go
+++ b/tools/client/workloads.go
@@ -19,9 +19,9 @@ func (w *httpWorkloads) Create(reservation workloads.Reservation) (id schema.ID,
 	return
 }
 
-func (w *httpWorkloads) List(nextAction workloads.NextActionEnum, customerTid int64, page *Pager) (reservation []workloads.Reservation, err error) {
+func (w *httpWorkloads) List(nextAction *workloads.NextActionEnum, customerTid int64, page *Pager) (reservation []workloads.Reservation, err error) {
 	query := url.Values{}
-	if nextAction != 0 {
+	if nextAction != nil {
 		query.Set("next_action", fmt.Sprint(nextAction))
 	}
 	if customerTid != 0 {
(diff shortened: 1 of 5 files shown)


zaibon (approved):
LGTM, but CI is not happy

tools/client/workloads.go
```diff
@@ -19,8 +19,14 @@ func (w *httpWorkloads) Create(reservation workloads.Reservation) (id schema.ID,
 	return
 }
 
-func (w *httpWorkloads) List(page *Pager) (reservation []workloads.Reservation, err error) {
+func (w *httpWorkloads) List(nextAction workloads.NextActionEnum, customerTid int64, page *Pager) (reservation []workloads.Reservation, err error) {
 	query := url.Values{}
+	if nextAction != 0 {
```
muhamadazmy: Actually nextAction of value 0 is a valid value. (check available values for the enum)
To work around this u can either do
- (check if value is < 0) and then caller has to pass of `-1` to donate (not set) (I don't like this very much)
- Or, you can change the input to `*NextActionEnum` then a nil value is possible, and then u only pass pointer to the value if u want to filter with next action. then u can do `if nextAction != nil` then `query.Set("next_action", *nextAction)`

muhamadazmy: I just realized that the first option is not even possible because the Enum type is of size `uint8` so no support for minus value. Hence only possible option is using pointer value

tools/explorer/pkg/workloads/types/reservation.go
```diff
@@ -45,26 +46,23 @@ const (
...
 
-// ReservationQuery type to parse reservation query string
-type ReservationQuery struct {
-	CustomerID int
-	NextAction int
-}
-
 // Parse parese the query string
-func (rq *ReservationQuery) Parse(r *http.Request) mw.Response {
+func Parse(r *http.Request, filter ReservationFilter) (ReservationFilter, mw.Response) {
```
muhamadazmy: Why this return a mw.Response and not an error? I can see you only return bad request 
so It's better to return an error, and in the handler if u receive an error u wrap it in the bad request response.

grimpy: no reaon at all corrected

muhamadazmy (approved):

PR #642, opened 2020-03-30
Update docs

maximevanhees:
Just some very small minor docs updates when I saw something was outdated.


zaibon (changes requested):
I think instead you can just overwrite `0-OS_v2_architecture.png` with your version instead of adding new files so you do not have to touch the readme

zaibon:
@maximevanhees ping :-D

zaibon (approved):

PR #656, opened 2020-04-01
Multisig on escrow accounts

### LINKED ISSUE #651: Add multisig to escrow accounts
Add multisig support to escrow account to recover funds if the database is wiped.

We also need some sort of tool to perform multisig actions and sign them

DylanVerstraete:
- Adds Signers to escrow account in order to recover funds if something bad happens.
- Adds a cli tool to recover these funds


tools/explorer/pkg/stellar/stellar.go
```diff
@@ -144,9 +151,111 @@ func (w *Wallet) CreateAccount() (keypair.Full, error) {
 		return keypair.Full{}, errors.Wrap(hError.Problem, "error submitting transaction")
 	}
 
+	if len(w.signers) == 5 {
```
zaibon: So we only create the mutlisign transaction if we have 5 signer address define. Shouldn't we allow creating the transaction as soon as we have 3 or more signers ? Since we requires only 3 to be able to create transaction later on ?

zaibon: And shouldn't this all block be move into it's own function ?

zaibon (changes requested):

tools/explorer/pkg/stellar/stellar.go
DylanVerstraete: Well we said we were going to add 5 signers to the escrow accounts which of only 3 out of 5 must sign off on a transaction.

LeeSmet (commented):
Code wise it looks good. But I agree with the comments from scrum, we should not enforce exactly 5 signers, rather a minimum of 3 signers , and print a warning at startup otherwise.

Now that the multisig is also setup when the escrow is created, I feel @zaibon is right and the setup method should be refactored, and each piece of functionality moved into its own sub method, which would clear up the code a bit. 

Otherwise, I noticed the code to create a multisig tx is added in the stellar wallet implementation in the explorer. I'm not sure I agree with this, since the explorer itself will never use it. Personally I'd rather see this moved to the new tool you created. But thats more of a "purist" remark so I won't deny this PR just for that

zaibon:
> Otherwise, I noticed the code to create a multisig tx is added in the stellar wallet implementation in the explorer. I'm not sure I agree with this, since the explorer itself will never use it. Personally I'd rather see this moved to the new tool you created. But thats more of a "purist" remark so I won't deny this PR just for that

I kinda had the same feeling about it, but does that means we have to copy all the wallet code ?
If so maybe it's not worth it.

DylanVerstraete:
> I kinda had the same feeling about it, but does that means we have to copy all the wallet code ?
> If so maybe it's not worth it.

Yes I had the same feeling during developing it. I was trying to wrap the wallet struct in:

```type multisigWallet stellar.Wallet```

But all the fields in stellar.Wallet are unexported so I could not create an instance of it.

@LeeSmet what would be the solution here? Export the fields in the stellar.Wallet?

**Comment by LeeSmet** [comment]
No, I would create a new wallet struct in the multisig tool, which basically just has a stellar wallet as its single field. The helper methods to get the passphrase for the network can be exported on the base wallet, as these do provide utility. This way no fields need to be exported in the original wallet, the functions can move to the multisig package, and you can access the required helpers

File: tools/explorer/pkg/stellar/stellar.go [hunk 2]
```diff
@@ -105,53 +139,108 @@ func (w *Wallet) CreateAccount() (keypair.Full, error) {
...
 		hError := err.(*horizonclient.Error)
-		return keypair.Full{}, errors.Wrap(hError.Problem, "error submitting transaction")
+		return errors.Wrap(hError.Problem, "error submitting transaction")
 	}
+	return nil
+}
 
-	return *newKp, nil
+func (w *Wallet) setupEscrowMultisig(newKp *keypair.Full, sourceAccount hProtocol.Account, client *horizonclient.Client) error {
+	if len(w.signers) >= 3 {
```
LeeSmet: I'd prefer it if this condition is inverted, i.e.:

```
if len(w.signers) < 3 {
    // not enough signers, don't add multisig
    return nil
}
```

This way it is immediately clear that this function will take no further action in this case, and that this is indeed a terminal condition

LeeSmet (changes requested):
looks pretty good. Apart from my other comment, I'd also like it if the `setupTrustline` and `setupEscrowMultisig` functions don't each create a transaction, but rather they each add their operations onto a shared transaction, so we create 1 fewer tx already here. And if at all possible, it would be really nice if this tx is also shared by `activeEscrowAccount`, though I'm not sure if that is possible

tools/explorer/pkg/stellar/stellar.go
DylanVerstraete: @LeeSmet  What do you mean by `activeEscrowAccount` ?

LeeSmet (approved):

LeeSmet:
LGTM, though you'll have to rebase master first

PR #671, opened 2020-04-02
explorer: allow the explorer to run with escrow disabled

zaibon:


tools/explorer/pkg/workloads/reservation.go
```diff
@@ -372,6 +378,15 @@ func (a *API) workloads(r *http.Request) (interface{}, mw.Response) {
 			}
 		}
 
+		// if escrow is not enabled, we skip the pay state go to deploy directly
```
LeeSmet: Why is there specific logic being added for the `Free` escrow (and why here)? I did noticed that the `Free` escrow has only blanket implementations for the escrow interface - I think the logic to move from `Pay` to `Deploy` state should be moved to the `RegisterReservation` method. Then we have identical flow for either the rea or "fake" escrow (from the reservation standpoint ofc).

LeeSmet (changes requested):

tools/explorer/pkg/escrow/escrow.go
```diff
@@ -2,491 +2,52 @@ package escrow
...
-	}
-	kp, err := e.wallet.KeyPairFromSeed(addressInfo.Secret)
-	if err != nil {
-		log.Error().Msgf("failed to parse escrow address secret: %s", err)
-		return errors.Wrap(err, "could not load escrow address info")
-	}
-	if err = e.wallet.PayoutFarmers(*kp, paymentInfo, id); err != nil {
-		log.Error().Msgf("failed to pay farmer: %s for reservation %d", err, id)
-		return errors.Wrap(err, "could not pay farmer")
+		reservation.NextAction = workloads.NextActionDeploy
```
LeeSmet: Reservation is not a pointer, and the object is not returned (or used anywhere else in this flow), so I don't think this will do anything

LeeSmet (changes requested):

LeeSmet (approved):
lgtm, I'm happy if CI is happy :)

Issue #646, opened 2020-03-30
tfuser provision requires seed flag to be set twice


LeeSmet:
When trying to create a provision using tfuser, you now need to set the `--seed` flag both after `tfuser` and after `provision`, since the flag is declared on both, and is required on both. If the `seed` flag is set only once, it will fail with an error noting a required flag is missing.

zaibon:
This would be solved with #660

zaibon:
fixed in #673

PR #673, opened 2020-04-02
Fix tfuser and tffarmer with new seed option

### LINKED ISSUE #660: tfuser: rework how the identity "seed" is stored
At the moment all the developer tooling `tfuser` and `tffarmer` rely on a file where the seed used to generate their key pair is stored. But the key pair is not enough to identity a user on the explorer anymore. the `threebot-id`, an integer, is also required. At the moment the commands that needs this `threebot-id` have just use flag for it. This is very cumbersome to use cause the user needs to know what his `threebot-id` is. 

Instead I propose that the "seed" file is converted into something that stores both information. The seed and the `threebot-id`.

So instead of just dumping the seed into a file, the `tfuser id` command would write a struct like:
```go
type Indentity struct{
    // base64 version of the seed
    Seed string
   // ThreebotID return by the explorer users/create endpoint
   ThreebotID uint
```

Having this, both `tffarmer` and `tfuser` just have to use one single flag that point to this file to have all the required information to identity the user of the commands.

maxux:
- Add tffarmer custom seed field
- Fix tfuser to create new seed without specifying a seed existing
- Fix seed needed twice for provision #646
- Improve how seed is required or not on tfuser
- Support seed 1.1.0 which bundle seed mnemonic and threebot id
- Update `tffarmer` to support new seed and use threebot id from it
- Update `tfuser` to use new seed and use threebot id from seed
- Move `tfuser id` to `tfuser id create` command
- Add `tfuser id show` and `tfuser id convert` commands

Seed argument is not mandatory but can be specified globally. If the seed is mandatory for a subcommand, it will be rejected if not available.


pkg/identity/userid.go
```diff
@@ -0,0 +1,104 @@
...
+// UserIdentity defines serializable strict to identify a user
+type UserIdentity struct {
+	// Seed encoded in base64
+	Mnemonic string `json:"mnemonic"`
+	// ThreebotID generated by explorer
+	ThreebotID uint64 `json:"threebotid"`
+}
+
+// UserData contains all information needed to identify someone
+type UserData struct {
```
zaibon: This struct could be avoided I think. You already have all the information required in UserIdentity. Instead of having another struct just to have access to the KeyPair, you can just add a KeyPair method on the UserIdentity struct.

pkg/identity/userid.go
```diff
@@ -0,0 +1,104 @@
...
+	err = json.Unmarshal(buf, &ui)
+	if err != nil {
+		return ud, err
+	}
+
+	return LoadUserIdentityObject(&ui)
+}
+
+// LoadUserIdentityObject converts a UserIdentity object to UserData
+func LoadUserIdentityObject(ui *UserIdentity) (UserData, error) {
```
zaibon: This can also be removed if you add the KeyPair method on UserIdentity struct

pkg/identity/userid.go
```diff
@@ -0,0 +1,104 @@
...
+	ud.Key = kp
+
+	// Link ThreebotID
+	ud.ThreebotID = ui.ThreebotID
+
+	return ud, nil
+}
+
+// SaveUserData dumps UserData into a file
+func SaveUserData(u *UserData, path string) error {
```
zaibon: Same here.

tools/seedconv/main.go
```diff
@@ -0,0 +1,52 @@
+package main
```
zaibon: Instead of add a new CLI move this into a sub command of `tfuser id`

tools/tfuser/cmds_identity.go
```diff
@@ -5,46 +5,78 @@ import (
...
 	description := c.String("description")
+	mnemonic := c.String("mnemonic")
 
-	k, err := identity.LoadKeyPair(output)
+	eud, err := identity.LoadUserIdentity(output)
 	if err == nil {
-		fmt.Printf("a seed already exists at %s\n", output)
-		fmt.Printf("identity: %s\n", k.Identity())
+		fmt.Printf("A seed already exists at: %s\n", output)
+		fmt.Printf("Identity: %s\n", eud.Key.Identity())
```
zaibon: while we are at it, let's also print the threebot ID here.

tools/tfuser/main.go
```diff
@@ -6,17 +6,27 @@ import (
...
 
 var (
-	bcdb *client.Client
+	bcdb     *client.Client
+	bcdbAddr string
+	mainSeed string
 )
 
-func main() {
+func requireSeed(c *cli.Context) error {
```
zaibon, line 22: This is not needed, you can set the `Required` property of the `cli.StringFlag` to true on line 50 and have the same result.

zaibon (approved):

pkg/identity/userid.go
maxux: Ah, yes. Wasn't sure we could do that in Go, it's much better, for sure. I'll change that.

pkg/identity/userid.go
maxux: Yep, cool

pkg/identity/userid.go
maxux: This can be set as a method, right ?

tools/seedconv/main.go
maxux: True, was an external tool to make things easier to test, I can now move it.

tools/tfuser/main.go
maxux: No, this was the initial issue. It you set it as required, you NEED to specify it for ALL subcommands, and you can't specify it for `id` subcommand. This is the exact problem of the main issue :p

pkg/identity/userid.go
zaibon: Well the UserData struct won't exist anymore so this method has no reason o exists

tools/tfuser/main.go
zaibon: Right 👌

pkg/identity/userid.go
```diff
@@ -0,0 +1,95 @@
...
+	// Mnemonic words of Private Key
+	Mnemonic string `json:"mnemonic"`
+	// ThreebotID generated by explorer
+	ThreebotID uint64 `json:"threebotid"`
+	// Internal keypair not exported
+	key KeyPair
+}
+
+// Key returns the internal KeyPair
+func (u *UserIdentity) Key() KeyPair {
```
zaibon, line 32: Instead of having a method to change the key, I would have a constructor instead. So you still have a way to build a `UserIdentity` from a specific key pair. But once the object is created you cannot mess it up by changing the key.

It seems this function is only used here https://github.com/threefoldtech/zos/pull/673/files#diff-4b63591c005a6e0aa77c5d468ba7b616R99

tools/tfuser/cmds_identity.go
```diff
@@ -1,50 +1,141 @@
...
+	// Load original seed file
+	kp, err := identity.LoadLegacyKeyPair(source)
+	if err != nil {
+		log.Fatal().Err(err).Msg("load key pair")
+	}
+
+	// Create new object
+	ui := &identity.UserIdentity{}
+	ui.SetKey(kp)
+	ui.ThreebotID = uint64(tid)
```
zaibon:
```suggestion
	ui := identity.NewUserIdentity(pk, tid)
```

zaibon (changes requested):

PR #678, opened 2020-04-03
Close #674: Rework escrow address usage

### LINKED ISSUE #674: explorer: change the way addresses are generated for reservation payment.
Follow up from https://github.com/threefoldtech/zos/issues/666

> In a perfect world, wallets would allow to scan a QR code that creates a transaction to pay all the farmers. But reality is different and most wallet just allow creation of transaction 1 to 1.

So to solve this we need to take another approach. Instead of creating an address per user-farmer pair. There will be a single address per user. The explorer will then be responsible to split the total amount between farmers.
This solution keeps all the same security properties the current solution has and also allow better UX for the user.

LeeSmet:
Generate 1 escrow address for a user, which is used to fund all
reservations for all farmers. We later split the payment, so that each
farmer gets the correct cut, in the explorer itself.

### DIFF
--- a/tools/explorer/pkg/escrow/escrow.go
+++ b/tools/explorer/pkg/escrow/escrow.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
 	"github.com/stellar/go/xdr"
 	"github.com/threefoldtech/zos/pkg/schema"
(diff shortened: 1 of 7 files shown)


tools/explorer/pkg/escrow/escrow.go
```diff
@@ -169,100 +173,109 @@ func (e *Escrow) checkReservations() error {
...
-			reservation, _ = pl.Next()
-			if !reservation.IsAny(workloadtypes.Pay) {
-				// Do not continue, but also take no action to drive the reservation
-				// as much as possible from the main explorer part.
-				slog.Warn().Msgf("reservation is paid, but no longer in pay state")
-				continue
-			}
+		reservation, err := workloadtypes.ReservationFilter{}.WithID(escrowInfo.ReservationID).Get(e.ctx, e.db)
+		if err != nil {
+			slog.Error().Err(err).Msgf("failed to load reservation")
```
zaibon: missing return ?

tools/explorer/pkg/escrow/escrow.go
```diff
@@ -169,100 +173,109 @@ func (e *Escrow) checkReservations() error {
...
 
-			slog.Info().Msg("all farmer are paid, trying to move to deploy state")
-			// update reservation
-			if err = workloadtypes.ReservationSetNextAction(e.ctx, e.db, escrowInfo.ReservationID, workloadtypes.Deploy); err != nil {
-				slog.Error().Err(err).Msgf("failed to set reservation in deploy state")
-				// FIXME: we do nothing with the error ?
-			}
+		pl, err := workloadtypes.NewPipeline(reservation)
+		if err != nil {
+			slog.Error().Err(err).Msgf("failed to process reservation in pipeline")
```
zaibon: missing return ?

zaibon (changes requested):
found possible missing error return otherwise LGTM

tools/explorer/pkg/escrow/escrow.go
```diff
@@ -170,63 +169,89 @@ func (e *Escrow) checkReservations() error {
...
-		}
-
-		balance, _, err := e.wallet.GetBalance(escrowInfo.Address, escrowInfo.ReservationID)
-		if err != nil {
-			slog.Error().Err(err).Msgf("failed to verify escrow account balance")
+		if err := e.checkReservationPaid(escrowInfo); err != nil {
+			log.Error().
+				Str("address", escrowInfo.Address).
+				Int64("reservation_id", int64(escrowInfo.ReservationID)).
+				Err(err)
```
zaibon: This line won't print anything zerolog requires to have `Msg`, `Msgf` or just `Send` if you don't want to add a message.

zaibon (changes requested):

zaibon (approved):

PR #688, opened 2020-04-07
explorer: handle the duration_provision field better in reservation flow

### LINKED ISSUE #686: User couldn't get his money back and the result of deployment in case of deployment failure
## Branches
libs, threebot: **unstable_reservation_payments**
other repos: **unstable**
## Environment 
Devnet

## Steps
1- Check wallet balance
```
JSX> wallet.get_balance()                                                                            
Balances
  81.7865580 TFT:GA47YZA3PKFUZMPLQ3B5F2E3CJIB57TGGU7SPCQT2WAEYKN766PWIMB3
  9999.9999400 XLM

```
2-  Try to deploy a container with fake network_id, everything should work fine and you should be able to submit your reservation request expecting reservation result to be [ ]
```
JSX> r = zos.reservation_create()                                                                                                                                                                           
JSX> nodes = zos.nodes_finder.nodes_search()                                                                                                                                                                
JSX> node6 = next(filter(zos.nodes_finder.filter_public_ip6, nodes))                                                                                                                                        
JSX> zos.container.create( 
   2         reservation=r, 
   3         node_id=node6.node_id, 
   4         network_name="fake", 
   5         ip_address="172.24.1.10", 
   6         flist="https://hub.grid.tf/zaibon/zaibon-ubuntu-ssh-0.0.2.flist", 
   7         entrypoint="/sbin/my_init", 
   8     )                                                                                                                                                                                                  
## tfgrid.workloads.reservation.container.1
 - workload_id         : 1
 - node_id             : HT9yNvHRQs65dyRJeztAdEnXKweCX2bhgvD9i1WFPUNs
 - flist               : https://hub.grid.tf/zaibon/zaibon-ubuntu-ssh-0.0.2.flist
 - hub_url             : 
 - environment         : {}
 - secret_environment  : {}
 - entrypoint          : /sbin/my_init
 - interactive         : False
 - volumes             : []
 - network_connection  :
    - network_id: fake    - ipaddress: 172.24.1.10    - public_ip6: False - stats_aggregator    : []
 - farmer_tid          : -

    ## tfgrid.workloads.reservation.container.capacity.1
     - cpu                 : 1
     - memory              : 1,024
    



JSX> expiration = j.data.time.epoch + (60 * 5)                                                                    
JSX> registered_reservation = zos.reservation_register(r, expiration)                                                                                                                                       
JSX> registered_reservation                                                                                                                                                                                 
## tfgrid.workloads.reservation.create.1
 - reservation_id      : 32

    ## tfgrid.workloads.reservation.escrow.1
     - address             : GDEZWIYPAV3OKKQYMQKJGSPAHBOBYRG7MPTYH54QXKVDSIP2KNCPKJUY
     - details             :
        - farmer_id: 2    - total_amount: 158,667,460



JSX> result = zos.reservation_result(registered_reservation.reservation_id)                                                                                                                                 
JSX> result                                                                                                                                                                                                 
[]

```
3- Pay the farmer , should succeed
```
JSX> res = zos.billing.payout_farmers(wallet, registered_reservation)
JSX> res                                                             
['69fccdde9254e36bd602f96f2a51b7f3d34c1fe7953a5204e991ef4bcfa4ce80']


```
4- Check tha wallet balance again, reservation amount should be deducted 
```
JSX> wallet.get_balance()                                                                                                                                                                                   
Balances
  65.9198120 TFT:GA47YZA3PKFUZMPLQ3B5F2E3CJIB57TGGU7SPCQT2WAEYKN766PWIMB3
  9999.9999300 XLM
```
5- Get the reservation result again

## Actual Result
- result still empty []
- user couldn't get back his tokens back 
```
JSX> time.sleep(60)                                                                                                                                                                                         
JSX> result = zos.reservation_result(registered_reservation.reservation_id)                                                                                                                                 
JSX> result                                                                                                                                                                                                 
[]

```
## Expected Result

1- To get a result that the state is "Error" with error message
i should get  a message like this (tried it on testnet without payment to see how the error message will be ) 
```
message = "couldn't load network with id (NPHavN6REaxH2):
```
2- User should get back his money

```
JSX> wallet.get_balance()                                                                                                                                                                                   
Balances
  65.9198120 TFT:GA47YZA3PKFUZMPLQ3B5F2E3CJIB57TGGU7SPCQT2WAEYKN766PWIMB3
  9999.9999300 XLM
```

zaibon:
a reservation is now cancelled if it is not deployed before the
duration_provision time.
a paid reservation that is cancelled is refunded

fixes #686

### DIFF
--- a/tools/explorer/pkg/escrow/escrow.go
+++ b/tools/explorer/pkg/escrow/escrow.go
@@ -95,7 +95,7 @@ func (e *Escrow) Run(ctx context.Context) error {
 			return nil
 
 		case <-ticker.C:
-			// log.Info().Msg("scanning active escrow accounts balance")
+			log.Info().Msg("scanning active escrow accounts balance")
 			if err := e.checkReservations(); err != nil {
 				log.Error().Err(err).Msgf("failed to check reservations")
 			}
@@ -147,10 +147,13 @@ func (e *Escrow) refundExpiredReservations() error {
 		return errors.Wrap(err, "failed to load active reservations from escrow")
 	}
 	for _, escrowInfo := range reservationEscrows {
+		log.Info().Int64("id", int64(escrowInfo.ReservationID)).Msg("expired escrow")
+
 		if err := e.refundEscrow(escrowInfo); err != nil {
 			log.Error().Err(err).Msgf("failed to refund reservation escrow")
 			continue
 		}
+
 		escrowInfo.Canceled = true
 		if err = types.ReservationPaymentInfoUpdate(e.ctx, e.db, escrowInfo); err != nil {
 			log.Error().Err(err).Msgf("failed to mark expired reservation escrow info as cancelled")
--- a/tools/explorer/pkg/workloads/types/pipeline.go
+++ b/tools/explorer/pkg/workloads/types/pipeline.go
@@ -98,12 +98,20 @@ func (p *Pipeline) Next() (Reservation, bool) {
 		return p.r, false
 	}
 
+	slog := log.With().Int64("id", int64(p.r.ID)).Logger()
+
 	// reseration expiration time must be checked, once expiration time is exceeded
 	// the reservation must be deleted
 	if p.r.Expired() || p.checkDeleteSignatures() {
 		// reservation has expired
 		// set its status (next action) to delete
-		log.Debug().Int64("id", int64(p.r.ID)).Msg("expired or to be deleted")
+		slog.Debug().Msg("expired or to be deleted")
+		p.r.NextAction = generated.NextActionDelete
+		return p.r, true
+	}
+
+	if p.r.DataReservation.ExpirationProvisioning.Before(time.Now()) && !p.r.IsSuccessfullyDeployed() {
+		log.Debug().Msg("provision expiration reached and not fully provisionned")
 		p.r.NextAction = generated.NextActionDelete
 		return p.r, true
 	}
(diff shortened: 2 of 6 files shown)


tools/explorer/pkg/workloads/types/pipeline.go
```diff
@@ -98,12 +98,21 @@ func (p *Pipeline) Next() (Reservation, bool) {
...
 	if p.r.Expired() || p.checkDeleteSignatures() {
 		// reservation has expired
 		// set its status (next action) to delete
-		log.Debug().Int64("id", int64(p.r.ID)).Msg("expired or to be deleted")
+		slog.Debug().Msg("expired or to be deleted")
+		p.r.NextAction = generated.NextActionDelete
+		return p.r, true
+	}
+
+	// log.Debug().Msgf("%s %s", p.r.DataReservation.ExpirationProvisioning.Unix(), time.Now().Unix())
```
LeeSmet: forgot to remove this? :p

tools/explorer/pkg/escrow/escrow.go
```diff
@@ -147,10 +147,16 @@ func (e *Escrow) refundExpiredReservations() error {
 		return errors.Wrap(err, "failed to load active reservations from escrow")
 	}
 	for _, escrowInfo := range reservationEscrows {
-		if err := e.refundEscrow(escrowInfo); err != nil {
-			log.Error().Err(err).Msgf("failed to refund reservation escrow")
-			continue
+		log.Info().Int64("id", int64(escrowInfo.ReservationID)).Msg("expired escrow")
+
+		if escrowInfo.Paid { //only refund if the reservation has been paid before
```
LeeSmet: The `paid` flag is used by the escrow to identify reservations which have been fully funded. This means that if I create a reservation, pay half of it, and then do nothing, `paid` will still be false. This way we allow for incremental payment, or payment from multiple sources, ... So this check should not be here, as it means in the scenario described your partial payment will be lost to the escrow and not refunded. 

Also, it is technically fine to call the refund method multiple times on an escrow

LeeSmet (changes requested):

tools/explorer/pkg/escrow/escrow.go
zaibon: How do you recommend to handle this then ? 
Cause here we do refund any expired transaction even when it has not been funded. So we are giving up free money

LeeSmet: No, this handles expired escrows. With the change in the db method to load expired escrows, an expired escrowed is now considered as follows:

Any escrow, tied to a reservation who's `expiration_provision` field is set to less than the current timestamp (i.e. we can no longer provision it). Also, we check the `released` and `canceled` fields are not set, so we are sure we did not cancel the reservation, or the reservation is not successful yet.

This means that we should refund ALL of these reservations, even if they are successfully paid already. If we only refund fully funded reservations, we indeed now refund reservations which have the provisioning expired while deploying (when node is down or so). But now we won't refund partially paid reservations, so this money will be stuck. Likewise, if we only pay not fully funded reservations back to the client, something like a node being down can cause the payment to not be refunded and end up with the money stuck. So we need to refund in both cases. 

Because we filter out reservations which have been canceled previously, or released (deployed successfully and farmers paid), there is no harm in doing this in either case. Also, the refund just checks how much money is in the escrow *for this reservation* and refunds it. So we only give to the client what he actually send us in the first place (again, for this reservation).

TL;DR: the check can just go away and we can refund everything we pulled from db in this loop

LeeSmet (approved):
LGTM :ship: :rocket:

PR #694, opened 2020-04-09
storaged: does not crash when lsblk response is empty

### LINKED ISSUE #675: storaged failed to start if lsblk returns no result
this is the same kind of error as https://github.com/threefoldtech/zos/issues/608

This is the possible root cause of  #616 and #617

maxux:
If system have no disks at all, `lsblk` (even in json) returns an empty string, which fails to be parsed as json.

This fixes prevent `storaged` to crash on empty response and let  starts with limited-cache feature.

In addition, this changes contains code which disable (unexpose) qemu overlay disk when system boots (after files have been copied).

This fixes #675

### DIFF
--- a/etc/zinit/init/node-ready.sh
+++ b/etc/zinit/init/node-ready.sh
@@ -9,10 +9,26 @@ setup_loopback() {
     ip link set lo up
 }
 
+disable_overlay() {
+    set -x
+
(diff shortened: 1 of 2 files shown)


zaibon (approved):
LGTM

PR #696, opened 2020-04-10
Encrypt generated escrow account seeds in explorer

### LINKED ISSUE #669: explorer: encrypt keypair of the generated accounts
extact from #666 

> If we want added security for the existing escrows, we could encrypt the keypair. This ties the escrows to the seed of the explorer (as this design would). Or, we could take that idea from this design, and setup the escrow addresses with a multisig where the explorer pubkey can sign on its own as well, and then throw away the key. This also ties the escrow to the used explorer seed. Both work I suppose, and this ("this" being the second option) would mean we save slightly less data in the db (we only need the address now).

LeeSmet:
Closes #669


zaibon (approved):
LGTM

PR #698, opened 2020-04-10
Multiple assets

DylanVerstraete:
- Adds FreeTFT, TFT and support for upcoming currencies
- Adds FreeTFT burning

### DIFF
--- a/tools/explorer/pkg/escrow/resource.go
+++ b/tools/explorer/pkg/escrow/resource.go
@@ -70,8 +70,8 @@ func (e Stellar) calculateReservationCost(rsuPerFarmerMap rsuPerFarmer) (map[int
 	return costPerFarmerMap, nil
 }
 
-func (e Stellar) processReservationResources(resData workloads.ReservationData) (rsuFarmer rsuPerFarmer, free bool, err error) {
-	free = false
+func (e Stellar) processReservationResources(resData workloads.ReservationData) (rsuFarmer rsuPerFarmer, freeToUse bool, err error) {
+	freeToUse = true
 	rsuPerNodeMap := make(rsuPerNode)
 	for _, cont := range resData.Containers {
 		rsuPerNodeMap[cont.NodeId] = rsuPerNodeMap[cont.NodeId].add(processContainer(cont))
@@ -91,12 +91,12 @@ func (e Stellar) processReservationResources(resData workloads.ReservationData) 
 		if err != nil {
 			return nil, false, errors.Wrap(err, "could not get node")
 		}
-		if node.FreeToUse {
-			free = true
+		if !node.FreeToUse {
+			freeToUse = false
 		}
 		rsuPerFarmerMap[node.FarmId] = rsuPerFarmerMap[node.FarmId].add(rsu)
 	}
-	return rsuPerFarmerMap, free, nil
+	return rsuPerFarmerMap, freeToUse, nil
 }
 
 func processContainer(cont workloads.Container) rsu {
--- a/tools/explorer/pkg/escrow/resource.go
+++ b/tools/explorer/pkg/escrow/resource.go
@@ -70,7 +70,8 @@ func (e Stellar) calculateReservationCost(rsuPerFarmerMap rsuPerFarmer) (map[int
 	return costPerFarmerMap, nil
 }
 
-func (e Stellar) processReservationResources(resData workloads.ReservationData) (rsuPerFarmer, error) {
+func (e Stellar) processReservationResources(resData workloads.ReservationData) (rsuFarmer rsuPerFarmer, free bool, err error) {
+	free = false
 	rsuPerNodeMap := make(rsuPerNode)
 	for _, cont := range resData.Containers {
 		rsuPerNodeMap[cont.NodeId] = rsuPerNodeMap[cont.NodeId].add(processContainer(cont))
@@ -88,11 +89,14 @@ func (e Stellar) processReservationResources(resData workloads.ReservationData) 
 	for nodeID, rsu := range rsuPerNodeMap {
 		node, err := e.nodeAPI.Get(e.ctx, e.db, nodeID, false)
 		if err != nil {
-			return nil, errors.Wrap(err, "could not get node")
+			return nil, false, errors.Wrap(err, "could not get node")
+		}
+		if node.FreeToUse {
+			free = true
 		}
 		rsuPerFarmerMap[node.FarmId] = rsuPerFarmerMap[node.FarmId].add(rsu)
 	}
-	return rsuPerFarmerMap, nil
+	return rsuPerFarmerMap, free, nil
 }
 
 func processContainer(cont workloads.Container) rsu {
--- a/tools/explorer/pkg/escrow/stellar.go
+++ b/tools/explorer/pkg/escrow/stellar.go
@@ -9,7 +9,6 @@ import (
 	"github.com/rs/zerolog/log"
 	"github.com/stellar/go/xdr"
 	"github.com/threefoldtech/zos/pkg/schema"
-	"github.com/threefoldtech/zos/tools/explorer/config"
 	gdirectory "github.com/threefoldtech/zos/tools/explorer/models/generated/directory"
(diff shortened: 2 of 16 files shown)


tools/explorer/pkg/escrow/resource.go
```diff
@@ -88,11 +89,14 @@ func (e Stellar) processReservationResources(resData workloads.ReservationData)
 	for nodeID, rsu := range rsuPerNodeMap {
 		node, err := e.nodeAPI.Get(e.ctx, e.db, nodeID, false)
 		if err != nil {
-			return nil, errors.Wrap(err, "could not get node")
+			return nil, false, errors.Wrap(err, "could not get node")
+		}
+		if node.FreeToUse {
```
zaibon: I don't think this is correct. Now if any of the workloads is deployed on a free to use node, then the full reservation is marked as free. It should be the opposite.

tools/explorer/pkg/escrow/stellar.go
```diff
@@ -482,11 +491,11 @@ func (e *Stellar) createOrLoadAccount(customerTID int64) (string, error) {
 	return res.Address, nil
 }
 
-func addressByAsset(addrs []gdirectory.WalletAddress, asset string) (string, error) {
+func getAddressFarmer(addrs []gdirectory.WalletAddress) (string, error) {
```
zaibon: I don't think you should hardcode the asset in here. Let's keep this function generic.

tools/explorer/pkg/escrow/stellar.go
```diff
@@ -352,11 +353,19 @@ func (e *Stellar) payoutFarmers(id schema.ID) error {
...
-			// FIXME: this is probably not ok, what do we do in this case ?
-			log.Error().Err(err).Msgf("failed to find address for %s for farmer %d", config.Config.Asset, farm.ID)
-			continue
+		// default use freeTFT issuer as destination
+		// if a reservation is free to use we send the tokens
+		// back to the issuer, this will freeze the tokens from being used again
+		destination := e.wallet.GetFreeTFTIssuer()
+
+		// if the reservation is not free we assume the currency is TFT
+		if !rpi.Free {
```
zaibon: I would prepare for the future and so something like
```go
if rpi.Free {
	destination := e.wallet.GetFreeTFTIssuer()
} else {
	destination, err = getAddressFarmer(farm.WalletAddresses, config.Config.Asset)
	if err != nil {
		// FIXME: this is probably not ok, what do we do in this case ?
		log.Error().Err(err).Msgf("failed to find address for %s for farmer %d", config.Config.Asset, farm.ID)
		continue
	}
}
```
So this code will work even if we use any other asset

zaibon (changes requested):

tools/explorer/pkg/escrow/resource.go
DylanVerstraete: I though we were going to validate the reservation upfront? So all the nodes for a specific reservation here would be freeToUse or not

LeeSmet (changes requested):
After looking through the PR a bit, I see some good stuff, but I also see that there is special handling in the escrow for freeTFT. I don't like this as it means the code is not generic after all, while this is possible. Since it would probably be a mess if I comment on the individual pieces of the PR, I won't do that, and instead give an overview of how I imagine this should be done here:

- The user now specifies which currency he pays in. This is  now implemented, but after sleeping a bit on this, I think we should make this an array rather than a single item. So a user can specify all currencies he can pay in. In case the first currency is not supported by all farmers, we check the next currency, etc. This way the user can create a reservation and fire all his owned currencies at once, which removes possible overhead of trying every currency separately (cc @zaibon your thoughts on this approach?). 

- We discussed and decided to verify that all nodes in the reservation are either free or paid to use on the reservation level (so not on the escrow). Since the logic that handles free reservations is already here, we can also add a few lines here that go: If all nodes free to use, only allow FreeTFT. so we can check that this is the only currency the user has provided. If multiple currencies are provided, the reservation is rejected. Or, we strip all currencies but FreeTFT. This is also possible but messes with the reservation, so I don't think this is actually a good idea. This means that all special logic for FreeTFT will be in the reservation (create) layer. We can work with a fully generic approach further down.

- When looking at the different currencies, we can actually see that there are only 2 differences (as of now): 
  - the asset and issuer, obviously
  - and the way in which payouts for these currencies to farmers happens

- We can tackle the second problem by creating a new type:
```go
type payoutKey struct {
    farmer: uint8
    burned: uint8
    foundation: uint8
}
```
The name can probably be improved a lot, but the idea is hopefully clear: we specify a structure in which we put which percentage of the actual cost of the reservations goes to which party. Since burning will be handles by returning the asset to the issuing account, all 3 of these actions are actually regular transfers, meaning we shouldn't have to touch the wallet code for the farmer payout, only the escrow part.

- Speaking of the wallet, its time to create an `asset` type in the wallet: 
```go
type Asset string
```
We will represent all currencies as an `Asset` type, which will take the form of `<code>:<issuer>`, and have some helper methods to extract the actual code and issuers from these assets. Assets can be declared as consts/vars in the escrow package, so other packages can access them should this be needed. We can then declare 3 maps: one for each network, where we use the asset as key, and a `payoutKey` (see previous point) as value. When a new escrow is created, we look at the network is passed, and load the correct map onto the wallet itself. This solves our first problem we had for the generic implementation.

- So now that we have all that, we follow the reservation flow. When the escrow receives a new reservation, it will now do an extra step: it will first extract which farmers are hosting these nodes, and create a list of all supported currencies. We then check which currencies are supported by all farmers and the client. Remember that we will check by iterating over the client currencies first, so we take the first matching currency from the client (so he can specify an order of preference in how to pay). And here's the catch: farmers won't have FreeTFT addresses. Recall that in the previous section we set up a map with the supported assets for our network as key, and the payout distribution as value. Now for FreeTFT the farmer payout value will be 0. So to make this generic, the function to determine the used currency will, for every client currency offered, first load the distribution value, and if the farmer field is 0, we just say all farmers agree on the fact they won't receive anything. This function returns the agreed upon `Asset`.

- So now we continue the usual flow, but this time we also pass along the `Asset` we extracted. Asset will be passed to the function which calculates the payment cost, so the right payment values can be loaded. Not that this matters now, since we use fixed values for everyone, but then it's already in place for the future. Neat.

- Obviously we save the `Asset` on the escrowinfo so we can retrieve it later. It will be needed e.g. when checking the balance, to make sure we don't pull in multiple assets (I'd need to check that method to be sure, but I doubt its 100% safe to use without modifications in multiple asset scenario's). But ofcourse it is also needed in the payoutFarmers method, as described in an earlier point. To finish, we also return the `Asset` to the user, as he might have provided multiple assets, so he'd probably like to know which one he needs to pay in (keep in mind that the escrow does not automatically refund payments made in the wrong currency). Here we will return a full `Asset` object. Technically, only the code would be enough since the client tool should have the issuer himself, but adding the issuer in the returned value adds some sanity checking options, and it just might prevent a possible fuck up somewhere in the chain.

- I already briefly mentioned the payoutFarmers method in the escrow. We will now get the asset next to the reservation cost per farmer. So we can calculate, for each farmer, how much he is owned. If this amount is greater than 0, we load in a suitable address from the farm object. Again this avoids having special logic to deal with FreeTFT and allows us to just add FreeTFTA or something in the future. Likewise, we can calculate the burn amount, the destination address is then extracted from the `Asset`. And finally, we calculate how much the foundation gets. Iirc we now send this to the seed used to start the explorer, though that actually feels a bit messy, so we should probably add a `-foundation-address` flag to start the explorer, and if its set use that (again cc @zaibon ?).

- To add more currencies in the future, it should now be as simple as declaring the `Asset`s (both testnet and mainnet I guess) in the escrow, and a `payoutKey` for them, and adding them into the respective maps.

That should be about everything I think.

PS: for the `ReservationCreate` type, to do this like I said on chat, you need to make the fields anonymous, so the type would become 
```go
type ReservationCreate struct {
    Reservation
    []string
}
```
You can look up `anonymous fields in structs in go` for more information

LeeSmet:
Closed in favor of #717

PR #707, opened 2020-04-12
review how an identity is passed to the explorer client

zaibon:
since we now need to authenticate 2 type of actor, nodes and users,
the client define an Identity interface whihc must be provided to the
constructor if the client need to authenticate the requests

2 type implement this interface, one for the nodes and one for the user.

This keep the http client generic

This PR also add a 'id import' command to tffuser

### DIFF
--- a/tools/tfuser/main.go
+++ b/tools/tfuser/main.go
@@ -14,7 +14,7 @@ import (
 
 var (
 	bcdb     *client.Client
-	mainui   identity.UserIdentity
+	mainui   *identity.UserIdentity
 	bcdbAddr string
 	mainSeed string
 )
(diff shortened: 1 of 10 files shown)


tools/tfuser/main.go
```diff
@@ -27,6 +29,18 @@ func requireSeed(c *cli.Context) error {
 	return nil
 }
 
+type userIdentity struct {
```
LeeSmet: This seems to be exactly the same struct as used in tffarmer, is there no way to only declare this once in a common file and then import it in both places?

LeeSmet (changes requested):

tools/tfuser/main.go
zaibon: it is indeed exactly the same, but I'm not sure where we could put it.

LeeSmet: Given that this identity is used by multiple tools packages, I'd say it could just be moved to a new `identity` pkg in the tools repo?

PR #710, opened 2020-04-13
keep track of storage units as bytes to allow counting part of a unit 

### LINKED ISSUE #703: provisiond: reserved resource does not count the SRU used by containers
This makes it hard to do proper node selection cause the reported storage reserved for a node is not correct.

250 MiB of SSD is reserved when deploying a container, but SRU is in GiB and we use a interger to store it, so before we count this we need to change the type of the resource unit to float

zaibon:
fixes #703


LeeSmet (approved):
LGTM when you fix tests :ship:

PR #717, opened 2020-04-14
Add multiple asset support in escrow

### LINKED ISSUE #663: explorer: support multiple currency in the escrow
At the moment the escrow can only deal with one currency at the time. This currency is defined in the command when starting the explorer.

This FR is about supporting multiple asset at the same time. This brings multiple question:
- How does a farmer define what currency he accepts ?
- Can a single reservation be paid with multiple currencies ?
- What currency should be set in the payment invitation ? All possible ? the preferred one from the farmer ?
- ...

### LINKED ISSUE #667: allow farmer to mark some node as free to use
In order to allow people to play with the capacity on mainnet without having to spend real money, a new stellar asset [FreeTFT](https://github.com/threefoldfoundation/tft-stellar/#freetft) [has been created](https://gist.github.com/robvanmieghem/324d7b33fb5961af0adb3225fa33d315#initial-grid-launch-only-with-stellar-tft).
The goal of this token is for any user to received a one time "gift" or a certain amount of FreeTFT.
Since these FreeTFT are meant to be free, the farmer need the ability to mark some of their node as free to use. Having this flag enabled on a node means the users can pay for reserved capacity on this node with FreeTFT. The FreeTFT will then be burn in the explorer instead of being transfered to the farmer.

LeeSmet:
Also adds support for FreeTFT
Closes #663
Closes #667

### DIFF
--- a/tools/explorer/pkg/escrow/escrow.go
+++ b/tools/explorer/pkg/escrow/escrow.go
@@ -14,7 +14,7 @@ import (
 // Escrow are responsible for the payment flow of a reservation
 type Escrow interface {
 	Run(ctx context.Context) error
-	RegisterReservation(reservation workloads.Reservation) (types.CustomerEscrowInformation, error)
+	RegisterReservation(reservation workloads.Reservation, supportedCurrencies []string) (types.CustomerEscrowInformation, error)
 	ReservationDeployed(reservationID schema.ID)
 	ReservationCanceled(reservationID schema.ID)
 }
@@ -33,7 +33,7 @@ func (e *Free) Run(ctx context.Context) error {
 }
 
 // RegisterReservation implements the escrow interface
-func (e *Free) RegisterReservation(reservation workloads.Reservation) (detail types.CustomerEscrowInformation, err error) {
+func (e *Free) RegisterReservation(reservation workloads.Reservation, _ []string) (detail types.CustomerEscrowInformation, err error) {
 
 	if reservation.NextAction == workloads.NextActionPay {
 		if err = workloadstypes.ReservationSetNextAction(context.Background(), e.db, reservation.ID, workloads.NextActionDeploy); err != nil {
--- a/tools/explorer/pkg/workloads/reservation.go
+++ b/tools/explorer/pkg/workloads/reservation.go
@@ -29,19 +29,24 @@ import (
 	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
-// API struct
-type API struct {
-	escrow escrow.Escrow
-}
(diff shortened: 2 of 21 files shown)


tools/explorer/pkg/escrow/escrow.go
```diff
@@ -14,7 +14,7 @@ import (
 // Escrow are responsible for the payment flow of a reservation
 type Escrow interface {
 	Run(ctx context.Context) error
-	RegisterReservation(reservation workloads.Reservation) (types.CustomerEscrowInformation, error)
+	RegisterReservation(reservation workloads.Reservation, supporteCurrencies []string) (types.CustomerEscrowInformation, error)
```
zaibon:
```suggestion
	RegisterReservation(reservation workloads.Reservation, supportedCurrencies []string) (types.CustomerEscrowInformation, error)
```

tools/explorer/pkg/workloads/reservation.go
```diff
@@ -58,10 +69,15 @@ func (a *API) validAddresses(ctx context.Context, db *mongo.Database, res *types
 		return err
 	}
 
-	validator := stellar.NewAddressValidator(config.Config.Network, config.Config.Asset)
-
 	for _, farm := range farms {
 		for _, a := range farm.WalletAddresses {
+			validator, err := stellar.NewAddressValidator(config.Config.Network, a.Asset)
```
zaibon, line 68: What is going to happen if someone has in its wallet addresses an asset not supported by the explorer? 
As example, now most of the farm have a fake TFChain asset code in their wallet addesses.

zaibon (approved):
LGTM. Just added one question regarding farm wallet address.

zaibon:
Can we also make sure to add some documentation regarding the new requirement when sending a reservation and also explain how exactly the different currencies are handled.

tools/explorer/pkg/workloads/reservation.go
LeeSmet: `NewAddressValidator` has been slightly reworked, it now creates a wallet, and tries to load the given asset code. If this fails, an `ErrAssetCodeNotSupported` is returned by the wallet, which is then wrapped and propagated up. 2 lines below, this code, we check if the returned errors `Is` an instance of this error, and ignore this address should this be the case. So we only validate addresses which are associated with a known asset(code). This should be the same behavior as the old implementation afaik

zaibon: Alright perfect.

zaibon (approved):

PR #725, opened 2020-04-15
tffarmer : allow to update farm information

zaibon:


tools/tffarmer/farm.go
```diff
@@ -2,42 +2,106 @@ package main
...
+
+	farm.ID, err = db.FarmRegister(farm)
 	if err != nil {
 		return err
 	}
 
-	fmt.Println("Farm registered successfully")
-	fmt.Printf("Name: %s\n", name)
-	fmt.Printf("Farm ID: %d\n", farmID)
+	fmt.Println("Farm updated successfully")
```
LeeSmet, line 43: why does this say "updated" in the register handler?

zaibon: because of copy pastaaaaaa

PR #731, opened 2020-04-16
Calculate starting balance of escrow account, closes #692

### LINKED ISSUE #692: Escrow: reduce amount of xlm used to active account
Currently new escrow accounts get activated by funding them with a static fee of 10 xlm. This amount can be reduced (need to calculate exactly how much we set), so we waste less xlm

At the same time, we already fund txes from escrow accounts using the base account. So we should investigate if we can recover these funds after the account is set up correctly

DylanVerstraete:


tools/explorer/pkg/stellar/stellar.go
```diff
@@ -32,10 +32,11 @@ type (
...
 	Wallet struct {
-		keypair *keypair.Full
-		network string
-		assets  map[Asset]struct{}
-		signers Signers
+		keypair        *keypair.Full
+		network        string
+		assets         map[Asset]struct{}
+		signers        Signers
+		minimumBalance float64
```
LeeSmet: calculating the minimum balance is very cheap computationally, so lets just store it in a function on the wallet. This also means that we don't have to change anything in the future if we would decide to load more signers // assets at runtime

tools/explorer/pkg/stellar/stellar.go
```diff
@@ -73,10 +74,18 @@ func New(seed, network string, signers []string) (*Wallet, error) {
 		log.Warn().Msg("to enable escrow account recovery, provide atleast 3 signers")
 	}
 
+	// mimimumBalance can be calculated as following
+	// Minimum Balance = (2 + # of entries) * base reserve
+	// entries is the amount of operations are required to setup the account
+	// we have 3 trustline operations for assets, and an equally number of operations for
+	// the lenght of the signers
+	minimumBalance := float64((2 + 3 + len(signers))) * 0.5
```
LeeSmet: the amount of trustlines is currently 2, as we only support 2 assets right now, but will be 3 in the near future, so this comment is wrong at the moment. Regardless, the amount of supported assets is known here since we have the map. So we can just replace the static "3" here with `len(w.assets)` (after moving to a function). This means we won't have to touch this when a new asset is added, and allows us to have a different amount of assets when we are running on testnet or mainnet

LeeSmet (changes requested):
I don't like the usage of floats, since we are working with a currency. A floating point error going down will cause the escrow setup to fail. This is easily sovled by working with uint64's, in stropes. We know that 1 XLM is 1e7 stropes. So we can replace this calculation to (2 + len(w.assets) + len(w.signers)) * (1e7 / 2 == 5e6). Once we plug in the string, we can then use a *big.Rat to format it (as is done in the tfuser provision command) without risking a floating point error anywhere

tools/explorer/pkg/stellar/stellar.go
DylanVerstraete: Actually, a rounding error cannot occur here. Since signers is always an even number?

tools/explorer/pkg/stellar/stellar.go
DylanVerstraete: Actually, a rounding error cannot occur here. Since signers is always an even number?

tools/explorer/pkg/stellar/stellar.go
```diff
@@ -597,6 +599,13 @@ func (w *Wallet) GetNetworkPassPhrase() string {
 	}
 }
 
+// getMinumumBalance calculates minimum balance for an escrow account
+// following formula is used: minimum Balance = (2 + # of entries) * base reserve
+// entries is the amount of operations are required to setup the account
+func (w *Wallet) getMinumumBalance() int {
+	return (2 + len(w.assets) + len(w.signers)) * (1e7 / 2)
```
LeeSmet: lets make this `1e7` a constant (and set it to 10000000 written out so it for sure is never a float), so it is immediatly clear what this is supposed to mean

LeeSmet (changes requested):
looks good, 1 more comment

LeeSmet (approved):

PR #733, opened 2020-04-16
restructure stellar code to reduce calls to horizon

### LINKED ISSUE #690: explorer: restructure stellar code a bit to reduce horizon calls
There are 3 places where we (sometimes) do redundant calls to the escrow:

- In the refund method, we get the source account, then the balance. If we invert this, we save a call if the balance is 0, as then we just exit.
- In the payout farmers method, we get the balance as a sanity check and to provide a nice error message. But in the flow this method is only called after a successful deployment, which only happens after the escrow confirms it has enough balance, so this check is obsolete. If it would fail, it would indicate a serious problem elsewhere. And the tx submit would fail, so we still have a (more cryptic) error.
- when getting the balance, due to paging, we keep loading pages until a page is empty. We can save a call by explicitly setting the page size and verifying that the returned page is equal to this size. If it is smaller, it was the last page.

DylanVerstraete:
closes #690 

Increased the transactions per page when we fetch the balance to 200, this will reduce the number of calls to fetch balance on a heavily used escrow address.

### DIFF
--- a/tools/explorer/pkg/stellar/stellar.go
+++ b/tools/explorer/pkg/stellar/stellar.go
@@ -292,6 +292,7 @@ func (w *Wallet) GetBalance(address string, id schema.ID, asset Asset) (xdr.Int6
 	txReq := horizonclient.TransactionRequest{
 		ForAccount: address,
 		Cursor:     cursor,
+		Limit:      200,
 	}
 
 	log.Info().Str("address", address).Msg("fetching balance for address")
@@ -384,10 +385,7 @@ func (w *Wallet) Refund(encryptedSeed string, id schema.ID, asset Asset) error {
 	if err != nil {
 		return errors.Wrap(err, "could not get keypair from encrypted seed")
 	}
-	sourceAccount, err := w.GetAccountDetails(keypair.Address())
-	if err != nil {
-		return errors.Wrap(err, "failed to get source account")
-	}
+
 	amount, funders, err := w.GetBalance(keypair.Address(), id, asset)
 	if err != nil {
 		return errors.Wrap(err, "failed to get balance")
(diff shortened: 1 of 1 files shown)


tools/explorer/pkg/stellar/stellar.go
```diff
@@ -292,6 +292,7 @@ func (w *Wallet) GetBalance(address string, id schema.ID, asset Asset) (xdr.Int6
 	txReq := horizonclient.TransactionRequest{
 		ForAccount: address,
 		Cursor:     cursor,
+		Limit:      200,
```
LeeSmet: lets make this a constant, because you will also need to change the condition of the for loop below. rather than checking if the txes are empty, you'll need to check if the amount of txes is smaller than the limit

LeeSmet (changes requested):

tools/explorer/pkg/stellar/stellar.go
```diff
@@ -304,7 +307,7 @@ func (w *Wallet) GetBalance(address string, id schema.ID, asset Asset) (xdr.Int6
 	}
 
 	donors := make(map[string]struct{})
-	for len(txes.Embedded.Records) != 0 {
+	for len(txes.Embedded.Records) < stellarPageLimit {
```
LeeSmet: Comming to think of it, this check won't work if its put here. In here we need an unconditional for loop (otherwise we won't process txes from the last page if its not completely full), and the check needs to go near the end, just before we load more txes. if the check passes, we break out. Something like that should work I think

LeeSmet (changes requested):

LeeSmet (approved):

PR #734, opened 2020-04-16
Loading nodes and farms over time in the frontend

DylanVerstraete:


tools/explorer/mw/action.go
```diff
@@ -28,6 +28,7 @@ func AsHandlerFunc(a Action) http.HandlerFunc {
 		object, result := a(r)
 
 		w.Header().Set("Content-Type", "application/json")
+		w.Header().Set("access-control-expose-headers", "pages")
```
zaibon, line 31: What is this for ?

zaibon (changes requested):
LGTM so far. We need to also support pagination for the farm and we should be good to go

tools/explorer/mw/action.go
DylanVerstraete: This enables axios to pick up the pages field in the response headers of the api calls.

zaibon (approved):

Issue #706, opened 2020-04-12
3bot name should always be lowercase


xmonader:
3bot names will be forced on jumpscale side to be lowercase and it's better to align that with the explorer to force lowercase names in the registration

PR #735, opened 2020-04-16
Validate email is lowercase improve status code return by update user

grimpy:
See: #706


zaibon (approved):
LGTM

PR #736, opened 2020-04-17
Adds automatic and manual refresh

DylanVerstraete:
- Page will now refresh every 10 minutes of the user stays on the page
- Page will refresh is the user has not visited the page since 10 minutes
- Button for manual refresh
- Refactored code for reusability

### DIFF
--- a/tools/explorer/frontend/src/App.vue
+++ b/tools/explorer/frontend/src/App.vue
@@ -140,7 +140,6 @@ export default {
       clearInterval(this.refreshInterval)
     }
 
-
     // instead of refreshing every 10 minutes in the background
     // we do following:
     // when the user enters the page again and 10 minutes have passed since the first visit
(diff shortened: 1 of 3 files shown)


tools/explorer/frontend/src/App.vue
```diff
@@ -93,8 +101,8 @@
 </template>
 
 <script>
-import { mapGetters } from 'vuex'
-
+import { mapGetters, mapActions } from 'vuex'
+/* eslint-disable */
```
zaibon: is this needed ?

zaibon (approved):
very cool :-)

tools/explorer/frontend/src/App.vue
DylanVerstraete: ah no, oops! :D

zaibon:
moved to https://github.com/threefoldtech/tfexplorer/pull/3

PR #738, opened 2020-04-17
Implement container logger into shim binary plugin

### LINKED ISSUE #718: provisiond: stuck on a 0-db namespace decommission
```
2020-04-14T14:41:26Z info start decommissioning reservation duration=23h59m0s expired=true id=459135-2 tag= to-delete=true type=zdb
2020-04-14T14:41:26Z debug decommiion
2020-04-14T14:41:26Z debug exits
2020-04-14T14:41:26Z debug decomission func
2020-04-14T14:41:26Z debug decomission zdb storage find
2020-04-14T14:41:26Z debug decomission ensure container
2020-04-14T14:41:26Z debug decomission zdb connection
2020-04-14T14:41:26Z debug connect to unix:///var/run/zdb_zdb42231853-ecc2-481d-bc31-178c2667eab5/zdb.sock
2020-04-14T14:41:26Z debug decomission zdb conncet
2020-04-14T14:41:26Z debug zdb client connect
2020-04-14T14:41:26Z debug zdb new pool
2020-04-14T14:41:26Z debug zdb try ping
2020-04-14T14:41:26Z debug zdb dial
2020-04-14T14:41:26Z debug zdb dial end
```

maxux:
In order to solve #718 we will move container logging feature from `contd` to `shim`. They provide a way to use external library/binary to handle logging.

This pull request move logging functions from `contd` to a new `shim-zoslog` binary decidated to logging purpose.

When this pull request will be ready and approved, restarting `contd` won't cut logs anymore and fifo will never be not read, except if something really bad happen to the new `shim-zoslog` which still need to be tested :)


tools/zoslog/main.go
```diff
@@ -0,0 +1,110 @@
...
+	ConfigDir = "config"
+)
+
+func main() {
+	logging.Run(runlog)
+}
+
+func runlog(ctx context.Context, config *logging.Config, ready func() error) error {
+	// initializing container logger
+	cfgfile := path.Join(RootDir, ConfigDir, config.Namespace, fmt.Sprintf("%s-logs.json", config.ID))
```
zaibon:
```suggestion
	cfgfile := filepath.Join(RootDir, ConfigDir, config.Namespace, fmt.Sprintf("%s-logs.json", config.ID))
```

tools/zoslog/main.go
```diff
@@ -0,0 +1,110 @@
...
+
+	// create local default logs directory
+	local := path.Join(RootDir, LogsDir, config.Namespace)
+	if err = os.MkdirAll(local, 0755); err != nil {
+		return err
+	}
+
+	// hardcode local logfile
+	filepath := path.Join(local, fmt.Sprintf("%s.log", config.ID))
+	fileout, fileerr, err := logger.NewFile(filepath, filepath)
```
zaibon: do we still want to logs to disk ? I'm not sure

tools/zoslog/main.go
```diff
@@ -0,0 +1,110 @@
...
+
+	s := bufio.NewScanner(r)
+	for s.Scan() {
+		if s.Err() != nil {
+			return
+		}
+
+		// send each line to each backend
+		for _, x := range writers {
+			_, err := x.Write([]byte(s.Text() + "\n"))
```
zaibon:
```suggestion
			_, err := x.Write(s.Bytes())
```
do you need to add the `\n` ?

zaibon (changes requested):
Couple of remarks on the code itself.
I also would like we decide whether we keep logging to disk at all or not.
Normally nobody would ever be able to read those logs, so I'm not sure this is worth it to create them at all.
If we decide not to log to always log to disk, then the code needs to be adapted to only start the logging-shim when the user asked for it.

I'm also a bit concerned regarding the memory usage of those shim. Maybe we could measure how much capacity they use and count that in the resource unit computation of the reservation.

Another idea to limit the amount of memory for for those daemon is to only have one per user per node. But this requries a bit more management

tools/zoslog/main.go
maxux: Yes, otherwise your local log file (for example) won't get any new lines and all lines will be concatenated on a single line

tools/zoslog/main.go
maxux: It's not enabled by default now (see line 67)

zaibon (approved):

PR #739, opened 2020-04-19
make provision package re-usable by others

zaibon:
The idea here is to convert the pkg/provision package into something that can be re-used by other system.
The provision package now exposes a bunch of interface that are needed to create a Engine type.
The engine is now fully configurable:
- type of primitives
- encode/decode functions to convert from/to schema types and internal types
- local cache
- statistics

The fist use case of this is the creation of the tf-gateway. This new component has new primitives and I wanted to be able to reuse the provision engine for it.

Once we have these change stabalize that should allow us to add new primitives in a very easy way.

To avoid circular dependencies I also had to extract the explorer from this repo.
The dependency graph is now:
```
zos <- depends on - explorer
tfgateway <- depens on - explorer and zos
```


cmds/capacityd/main.go
```diff
@@ -49,31 +48,32 @@ func cap(ctx context.Context, client zbus.Client) {
...
-		log.Fatal().Err(err).Msgf("failed to read virtualized state")
-	}
-
-	ru := directory.ResourceAmount{
-		Cru: resources.CRU,
-		Mru: float64(resources.MRU),
-		Hru: float64(resources.HRU),
-		Sru: float64(resources.SRU),
-	}
+	// dmi, err := r.DMI()
```
muhamadazmy: Is this intentional to comment out this code? or will be restored

muhamadazmy (approved):

PR #743, opened 2020-04-21
WIP adds workloads types builders

DylanVerstraete:


zaibon (commented):
A good pattern here to create these kind of builder is to use functional options. 
The idea is to create a constructor method for the struct that takes the required arguments and a list of optional "option function".  This allows the API to be very flexible and expressive for the user. 

```
package main

import "fmt"

type Workload struct {
	required string
	foo      string
	bar      uint
}

type Opt func(*Workload) error

func WithFoo(f string) Opt {
	return func(w *Workload) error {
		w.foo = f
		return nil
	}
}

func WithBar(b uint) Opt {
	return func(w *Workload) error {
		w.bar = b
		return nil
	}
}

func BuildWorkload(required string, opts ...Opt) (*Workload, error) {
	w := Workload{
		required: required,
	}
	for _, f := range opts {
		if err := f(&w); err != nil {
			return nil, err
		}
	}
	return &w, nil
}

func main() {
	w, err := BuildWorkload(
		"mandatory",
		WithFoo("hello"),
	)
	if err != nil {
		panic(err)
	}

	fmt.Printf("%+v", w)
}
```

edit: a nice post about it: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

tools/tfuser/cmds_provision.go
```diff
@@ -20,99 +16,114 @@ import (
...
-		return errors.Wrap(err, "failed to load zdb reservation schema")
-	}
+		zdbBuilder, err := builders.LoadZdbBuilder(f)
+		if err != nil {
+			return errors.Wrap(err, "failed to load the zdb builder")
+		}
 
-	if config.SecretEnv == nil {
-		config.SecretEnv = make(map[string]string)
+		zdbBuilders = append(zdbBuilders, *zdbBuilder)
```
muhamadazmy: Why do u put this in this list first, and not call `reservation.WithZdb` directly ?

DylanVerstraete:
> A good pattern here to create these kind of builder is to use functional options.
> The idea is to create a constructor method for the struct that takes the required arguments and a list of optional "option function". This allows the API to be very flexible and expressive for the user.
> .....

Whilst I like what you suggest, it is in my opinion to late for this. It will require alot of work and refactor. Also I think with the builder pattern that is implemented its nice that the editor suggests what kind of methods there are on a builder. In your suggestion the user will have to go look in the sourcecode to see what methods are there.

zaibon:
> Also I think with the builder pattern that is implemented its nice that the editor suggests what kind of methods there are on a builder. In your suggestion the user will have to go look in the sourcecode to see what methods are there.

That is not true. Now, the user will need to code dive in the code and deduct what field he needs to populate or not. There is no way for the user to know what builder methods he needs to call to have a working object. 

> it is in my opinion to late for this.

? Nothing depend on this, it's not even merged yet.

muhamadazmy:
@zaibon The idea is that the NewXBuilder() will create an instance of the builder with sane (default) values. so  `NewXBuilder().WithY(y).WithZ(z)` will set (or override) some fields. Then once u do `.Build()` validation of missing required field (container flist) (and or processing other fields like encryption of env vars), then u get a valid workload. That's the whole point. So it's all available on the Builder. 

With the Options pattern, it is (imho) a lot of code actually, and then u have to look into docs to find which options are valid for what builder.

zaibon:
You don't have to look into the core with the options pattern since each workloads have its own options type. Which means you get error at build time. Now runtime like with the current implementation 

The current implementation doesn't give any hint to the user about which With fonction he needs to call.

muhamadazmy:
@zaibon I seriously don't get how current code doesn't show u what functions u can use with a builder !
```go
struct MyBuilder{}

func (b *MyBuilder) WithValue(value Value) {}
```
I mean look at this. All editors will immediatily show u what u can do with your builder.

zaibon:
I'm not saying you cannot see what are the possible function. I'm saying there is no way for the user to know what `With` function to use or not. 
Example:
```go
// Nothing tells me this code is not valid, I have to run test to realize I'm missing some field...
b:= NewZDBBuilder()
zdb := b.Build()
```

With my proposal
```go
//  In this case you are sure AT BUILD time, that this code is valid.
b := NewZDBBuilder(
      "nodeID", 
      10,
    // here I can pass optional configuration
     WithPassword("passworld"),
)
zdb,err := b.build()
```

muhamadazmy:
I don't see how this pattern makes it better, but now instead of just knowing which one to use or not, he has to find the ones he needs in a huge list of "With*" functions, where most of them are not valid option for your builder!
Because u will have With* that only work with your ZDB builder, and others that works with your container, etc... and all of them are exposed publicly on your library. Instead of just getting a limited With functions that are only valid for your context.

DylanVerstraete:
> I'm not saying you cannot see what are the possible function. I'm saying there is no way for the user to know what `With` function to use or not.
> ....

Instead we can keep current implementation and make the required values a parameter in the NewXBuilder()

example:
```
func NewVolumeBuilder(size int64, type workloads.DiskType) *VolumeBuilder {
	return &VolumeBuilder{
		Volume: workloads.Volume{
			Size: size,
			Type: type,
		},
	}
}
```

and after that the user can already add more optional config 
example:

```
func (v *VolumeBuilder) WithSomethingElse(somethingElse string) *VolumeBuilder {
	v.Volume.somethingElse = somethingElse
	return v
}
```

zaibon:
Seems like a good compromise 👍

zaibon:
moved to https://github.com/threefoldtech/tfexplorer/pull/10

PR #761, opened 2020-05-04
Fix attaching volume container

### LINKED ISSUE #759: failed to create container with attached volume in a single reservation
https://explorer.testnet.grid.tf/explorer/reservations/459798

DylanVerstraete:


pkg/provision/primitives/converter.go
```diff
@@ -42,6 +43,9 @@ func ContainerToProvisionType(c workloads.Container) (Container, string, error)
 	}
 
 	for i, mount := range c.Volumes {
+		if strings.HasPrefix(mount.VolumeId, "-") {
+			mount.VolumeId = reservationID + mount.VolumeId
```
zaibon, line 47: The format needs to be `reservationID-workloadsID`
So 
```suggestion
			mount.VolumeId = fmt.Sprintf("%s-%s", reservationID, mount.VolumeId)
```

pkg/provision/primitives/converter_test.go
```diff
@@ -148,7 +148,7 @@ func TestTfgridReservationContainer1_ToProvisionType(t *testing.T) {
 				NetworkConnection: tt.fields.NetworkConnection,
 				StatsAggregator:   tt.fields.StatsAggregator,
 			}
-			got, _, err := ContainerToProvisionType(c)
+			got, _, err := ContainerToProvisionType(c, "")
```
zaibon: can we add a test case that test the new code too  ?

zaibon (changes requested):

pkg/provision/primitives/converter.go
DylanVerstraete: this is not needed since it passes the check if it has a prefix. So just concatinating will work?

zaibon: right, my bad :-P

zaibon (approved):

PR #765, opened 2020-05-06
make sure we always at least poll once on the latest unprocessed id

zaibon:


muhamadazmy (commented):
I think this is not enough. If u going to pull the same last reservation over and over then you have to add a check that this id doesn't match the last reservation id u received. Also i am not sure yet how this race happen yet. Since u always poll for the reservation after the last one

PR #772, opened 2020-05-08
adds reporting of ndmz pub4 and pub6 ifaces, #632

### LINKED ISSUE #632: Zero OS does not barf if node is only connected via IPV4
Farmers like GEC connect nodes on ipv4 addresses only. they are listed as node in the explorer but can not be connected to for capacity usage. can we implement a  mechanism which doe snot allow ipv4 only connection in combination listing? thx

DylanVerstraete:


cmds/networkd/addr_watcher.go
```diff
@@ -96,6 +106,53 @@ func (w WatchedLinks) Forever(ctx context.Context) error {
 	}
 }
 
+func getNdmzInterfaces() ([]types.IfaceInfo, error) {
+	var output []types.IfaceInfo
+
+	f := func(_ ns.NetNS) error {
```
zaibon, line 112: nice kinda like the pattern you use for this.

zaibon (approved):
LGTM

PR #796, opened 2020-05-20
set uptime on boot

### LINKED ISSUE #786: Always send uptime when updating node info in the explorer
Right now it sees uptime is not send unless the dedicated uptime function is used. This causes the explorer data to be out of sync when another piece of info is updated (like capacity or nics). Since the explorer does need to update its own `Updated` field from when the node was last seen for these requests, a node should always send its uptime on all requests which modify its data on the explorer, so this value stays in sync. Related to threefoldtech/tfexplorer/issues/73

DylanVerstraete:


zaibon (changes requested):
That's one way to do it. But I fell it is a bit overkill to use zbus to get the info from capacityd while the code to read the uptime is very small. I fell this time we can live with a bit of duplication and just have a method to read the uptime directly in identityd.

zaibon (approved):

PR #801, opened 2020-05-25
report hostname to bcdb, #784

### LINKED ISSUE #784: report hostname of the the node to the explorer

DylanVerstraete:


cmds/identityd/main.go
```diff
@@ -454,18 +454,25 @@ func bcdbClient() (client.Directory, error) {
...
 	log.Info().Str("version", version).Msg("start registration of the node")
+	var err error
 
 	v1ID, _ := network.NodeIDv1()
 
 	publicKeyHex := hex.EncodeToString(base58.Decode(nodeID.Identity()))
 
+	hostName, err := os.Hostname()
+	if err != nil {
+		return err
```
muhamadazmy: Maybe not return here, instead set hotname to `<unknown>` or something. It's better than not having the node registered on the explorer

DylanVerstraete: good catch

muhamadazmy (approved):

PR #811, opened 2020-06-02
Include more clean up to the storage type conversion

muhamadazmy:


pkg/flist/flist.go
```diff
@@ -160,6 +160,11 @@ func (f *flistModule) mount(name, url, storage string, opts pkg.MountOptions) (s
 	sublog := log.With().Str("url", url).Str("storage", storage).Logger()
 	sublog.Info().Msg("request to mount flist")
 
+	if len(opts.Type) == 0 {
+		// sanity check in case type is not set always use hdd
+		opts.Type = pkg.HDDDevice
```
zaibon: I would return an error instead. At this point we should have the information, not having it is mostly sign of a bad request from the user.

pkg/provision/primitives/zdb.go
```diff
@@ -141,7 +141,9 @@ func (p *Provisioner) createZdbContainer(ctx context.Context, allocation pkg.All
 	rootFS, err := flist.Mount(zdbFlistURL, "", pkg.MountOptions{
 		Limit:    10,
 		ReadOnly: false,
+		Type:     pkg.HDDDevice,
```
zaibon: Do you remember why we didn't set this to readonly ? Feel like this is the way to go so we don't have to allocate the subvolume for the 0-db other then the one for the actual data.

zaibon (changes requested):
I think this is a good step to the right direction. Added a couple of remarks to it.

pkg/provision/primitives/zdb.go
muhamadazmy: I remember we had issues when the rootfs of zdb was readonly, but that was all what I remember. BTW this line is the actual cause of the issue.

pkg/flist/flist.go
muhamadazmy: actually this is an internal call. the user does not have a direct access to flist.Mount.

muhamadazmy: But in that case, since this is optional (for read/write flists) I should move this check when it's needed.

pkg/provision/primitives/zdb.go
zaibon: Then I think we need to have more logic. Cause now if there is no HDD, then we cannot deploy 0-db at all.
So I think we should try to deploy on any type of disks

zaibon (changes requested):

pkg/provision/primitives/zdb.go
muhamadazmy: I agree, but this is not related to this issue. We need to create a new issue (also to retry read-only rootfs) but this now cause minio setup on devent to fail.

pkg/provision/primitives/zdb.go
```diff
@@ -138,12 +138,26 @@ func (p *Provisioner) createZdbContainer(ctx context.Context, allocation pkg.All
 	hw := ifaceutil.HardwareAddrFromInputBytes([]byte(allocation.VolumeID))
 
 	slog.Debug().Str("flist", zdbFlistURL).Msg("mounting flist")
-	rootFS, err := flist.Mount(zdbFlistURL, "", pkg.MountOptions{
-		Limit:    10,
-		ReadOnly: false,
-	})
+	var err error
```
zaibon, line 141: :+1:

zaibon (approved):

PR #813, opened 2020-06-03
Retry to update if it fails

muhamadazmy:
While the code is simple, and due to the bug it fixes we can't grantee that all nodes will manage to actually get the fix.

### DIFF
--- a/cmds/identityd/main.go
+++ b/cmds/identityd/main.go
@@ -44,9 +44,8 @@ const (
 )
 
 const (
-	module          = "identityd"
-	seedName        = "seed.txt"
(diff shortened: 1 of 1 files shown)


zaibon (approved):

zaibon (changes requested):
Instead of having 20 tries, shoulnd't we just set the MaxElapsedTime on the ExponatialBackoff object ?

muhamadazmy:
@zaibon I didn't know that it exists :). What do u suggest this value to be, an hour ?

zaibon:
That seems reasonable

PR #830, opened 2020-06-17
Add zufs (0-fs) bins extra

maxux:


bins/packages/golang/golang.sh
```diff
@@ -0,0 +1,32 @@
+GOLANG_VERSION="1.14.2"
```
zaibon, line 1: Instead of doing this. You can just install go using an action. The same we use in test.yml workflow

zaibon (changes requested):

zaibon (approved):

PR #835, opened 2020-06-19
Update zdb containers when they are outdated

### LINKED ISSUE #780: force all zdb container to restart in order to enable shim-logs
It seems we still have some nodes in the field that have container running without shim-logs. We should adapt contd to force it to restart all container still running without shim-logs to avoid having the container stuck

maxux:
When `provisiond` starts, it will check if any running zdb container are outdated. If they are, it will restart theses containers using latest version.

Some changes occures:
- The zdb url is now hardcoded and doesn't follow a symlink anymore
- When provisiond starts, it will check for running zdb
- When provisiond starts, it will loads the flist expected hash from the hub
- Container module contains a new function to list containers based on a namespace
- Flist module contains a way to grab md5 from url directly
- Flist module contains a function to returns flist hash from a running g8ufs

This fixes #780, when provisiond will be updated on nodes, if there is a 0-db update (and there will be one, we can force it), all zdb will restart and new shim will be applied.

### DIFF
--- a/pkg/provision/primitives/zdb.go
+++ b/pkg/provision/primitives/zdb.go
@@ -403,11 +403,16 @@ func (p *Provisioner) upgradeRunningZdb(ctx context.Context) error {
 	// Listing running zdb containers
 	containers, err := contmod.List(zdbContainerNS)
 	if err != nil {
+		log.Error().Err(err).Msg("could not load containers list")
 		return err
 	}
 
 	// fetching extected hash
-	expected := "abcd"
+	expected, err := flistmod.FlistHash(zdbFlistURL)
+	if err != nil {
+		log.Error().Err(err).Msg("could not load expected flist hash")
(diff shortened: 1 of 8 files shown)


pkg/provision/primitives/zdb.go
```diff
@@ -477,7 +492,16 @@ func (p *Provisioner) upgradeRunningZdb(ctx context.Context) error {
 			}
 
 			// restarting zdb
-			_, err = p.ensureZdbContainer(ctx, allocation, zdbmode)
+
+			// mount the new flist
+			rootfs, err := p.zdbRootFS()
```
zaibon, line 502: Don't you should also unmount the previous flist version ?

zaibon (changes requested):

zaibon (approved):

PR #837, opened 2020-06-22
Master volume type

### LINKED ISSUE #832: HDD disks should only be used by 0-DB
To make sure we get best performance how of HDD, only one 0-DB should be able to access it.

This means volume and container root filesystem can only be deployed on SSD.

Todo:
- [x] provisiond refuse to create a volume or a container root filesytem subvolume on HDD
- [x] explorer should return an error if a reservation with a volume on HDD is received
- [x] chat flow should not propose HDD for root filesystem or volume anymore (also no SSD option for ZDB)
  - [x] Volume/Container should always be SSD
  - [x] ZDB should always be HDD
- ~~zdb should only support HDD namespace type (disable the option to select SSD)~~

DylanVerstraete:


pkg/flist.go
```diff
@@ -28,6 +30,17 @@ type MountOptions struct {
 	Type DeviceType
 }
 
+// Valid checks that mount options are valid
+func (m *MountOptions) Valid() error {
+	// we disable HDD support in flist creation. root fs of containers
+	// should always be SSD
+	if m.Type != SSDDevice {
```
zaibon: I don't think this belong at this level. Flistd should keep this capability. The reason is we want to be able to still mount flist using any type of disk for internal usage (upgrade, ...).

This check should be at provisiond instead.

pkg/provision/primitives/volume.go
```diff
@@ -36,6 +43,9 @@ func (p *Provisioner) volumeProvisionImpl(ctx context.Context, reservation *prov
 		return VolumeResult{}, err
 	}
 
+	if config.Type != pkg.SSDDevice {
```
zaibon, line 46: Actually since now we only support one type, should we keep the `Type` field at all ?

pkg/provision/primitives/container.go
```diff
@@ -98,6 +98,10 @@ func (p *Provisioner) containerProvisionImpl(ctx context.Context, reservation *p
 		return ContainerResult{}, err
 	}
 
+	if config.Capacity.DiskType == pkg.HDDDevice {
```
zaibon, line 101: Actually since now we only support one type, should we keep the DiskType field at all ?

pkg/provision/primitives/zdb.go
```diff
@@ -69,6 +69,10 @@ func (p *Provisioner) zdbProvisionImpl(ctx context.Context, reservation *provisi
 		return ZDBResult{}, errors.Wrap(err, "failed to decrypt namespace password")
 	}
 
+	if config.DiskType != pkg.HDDDevice {
```
zaibon: This is wrong. We still allow people to use 0-db on top of SSD if they want.
What we want to limit is the usage of HDD for anything else the 0-DB

zaibon (changes requested):

pkg/provision/primitives/volume.go
muhamadazmy: I was thinking the same, but then i thought may be we will need to revert that and support multiple types later again ?

pkg/provision/primitives/container.go
muhamadazmy: Same as above

pkg/provision/primitives/zdb.go
muhamadazmy: I understood that zdb can ONLY be HDD

muhamadazmy: Okay, a bit misunderstanding ... we still need to support ssd and hdd for zdb

pkg/provision/primitives/volume.go
zaibon: could be. And I guess it would make more sense from a compatibility point of view to not remove the field

pkg/provision/primitives/zdb.go
DylanVerstraete: So we drop this check?

zaibon (approved):

PR #841, opened 2020-06-23
Centralize environment dependent settings into environment package

### LINKED ISSUE #827: Centralize environment dependent settings into environment package
The `environment` package is already there to set settings based on environment settings and custom variable.

However, we still have some random location on the code where this should be moved.

https://github.com/threefoldtech/zos/blob/master/cmds/identityd/main.go#L234-L245
This should be set on environment package.

https://github.com/threefoldtech/zos/blob/master/pkg/flist/flist.go#L27
I think theses kind of settings should also be set on environment.

There is probably some other files too, I just saw theses two files by looking for something.

If we have all settings in the same place, it would be easier to know where to look to set custom settings, only change one place and later improve how to maybe custom that on runtime if needed.

DylanVerstraete:
Some other variables might be missing, @zaibon do you know where I missed some?

### DIFF
--- a/pkg/environment/environment.go
+++ b/pkg/environment/environment.go
@@ -18,6 +18,9 @@ type Environment struct {
 	BcdbURL      string
 	BcdbPassword string
 
+	FlistURL string
+	BinRepo  string
+
 	FarmerID pkg.FarmID
 	Orphan   bool
 
@@ -58,6 +61,8 @@ var (
 	envDev = Environment{
 		RunningMode: RunningDev,
 		BcdbURL:     "https://explorer.devnet.grid.tf/explorer",
+		FlistURL:    "zdb://hub.grid.tf:9900",
+		BinRepo:     "tf-zos-bins.dev",
 		// ProvisionTimeout:  60,
 		// ProvisionInterval: 10,
 	}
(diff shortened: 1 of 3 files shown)


zaibon:
From the top of my head like this I don't think of any other no.

pkg/environment/environment.go
```diff
@@ -147,6 +156,10 @@ func getEnvironmentFromParams(params kernel.Params) (Environment, error) {
 		env.BcdbPassword = e
 	}
 
+	if e := os.Getenv("ZOS_FLIST_URL"); e != "" {
+		env.FlistURL = e
```
maxux, line 151: That's nice ! But if adding an environment variable for flist, could add one for BinRepo too ? :)

zaibon: Actually, how is this intended to be used ? There is no way to set a variable in a zinit service.

maxux: We have others environment variable aswell defined. In the past, we talked about a way to define environment variable easily for the whole system, that was never implemented but still could be. It's easier IMO than editing service file to add custom flags (like debug)

zaibon: alright, @maxux let's open a follow up issue  for this then.

PR #852, opened 2020-06-27
ensure we always skip public peer from the same segment

zaibon:
this is to ensure that the node doesn't only connect to other node from
the same farm and have at least on other public peer connected to
yggdrasil network


DylanVerstraete (approved):
Looks good codewise :)

PR #858, opened 2020-06-30
Feature/capacity pool

zaibon:
Relates to https://github.com/threefoldtech/tfexplorer/pull/115

Todo:
- [x] adapt reservation source to handle new workload types
- [x] Allow for new workloads to be link to existing provisioned primitives and remove expiration
- [ ] ensure not duplicate network reservation are in cache to avoid network being decommissioned at some point


muhamadazmy (approved):
This looks good to me except for one note:
- We need to mark some support code for older reservation type as "deprecated" with a dead line on when this support is gonna be dropped, because currently supporting multiple formats/versions of reservation will increase the tech dept over time and will be harder to maintain the code. We will not know the purpose of some code in the future.

zaibon:
@muhamadazmy can you mark in the code where you see support for multiple format ?
Cause I don't really see what part of the code you are talking about. The node should only support one single format.

PR #859, opened 2020-06-30
Disk shutdown

### LINKED ISSUE #856: Shutdown disks after boot
We want to shutdown disks after boot in order to save power.

DylanVerstraete:


bins/packages/hd-idle/hd-idle.sh
```diff
@@ -0,0 +1,35 @@
...
+
+dependencies_hd-idle() {
+    . ${PKGDIR}/../golang/golang.sh
+    build_golang
+    HD_IDLE_HOME="${GOPATH}/src/github.com/adelolmo"
+}
+
+download_hd-idle() {
+    echo "[+] downloading and installing hd-idle"
+    go get github.com/adelolmo/hd-idle
```
maxux: We are supposed to ship specific version and not latest version automatically, right @zaibon ?
You should download a specific version from github, not using `go get` like this.

bins/packages/storcli/storcli.sh
```diff
@@ -0,0 +1,32 @@
+dependencies_storcli() {
+    apt-get update
+    apt-get install wget
```
maxux: `wget` is already installed by basesystem, which is supposed to be installed first for every packages.
You can skip this part.

bins/packages/storcli/storcli.sh
```diff
@@ -0,0 +1,32 @@
...
+
+    echo "[+] updating soruces list"
+    wget -O - http://archive.thomas-krenn.com/tk-archive.gpg.pub | apt-key add -
+    cd /etc/apt/sources.list.d
```
maxux: Maybe use `wget -O` to avoid switching directory

bins/packages/storcli/storcli.sh
```diff
@@ -0,0 +1,32 @@
...
+    wget http://archive.thomas-krenn.com/tk-main-xenial.list
+    wget http://archive.thomas-krenn.com/tk-optional-xenial.list
+    cd
+    apt-get update
+
+    echo "[+] installing storcli"
+    apt-get install storcli
```
maxux: Again, using this way, you don't specify a version, you just use latest version they provide.

maxux (changes requested):

bins/packages/storcli/storcli.sh
```diff
@@ -0,0 +1,32 @@
...
+
+build_storcli() {
+    pushd "${DISTDIR}"
+
+    dependencies_storcli
+    install_storcli
+
+    popd
+}
+
```
maxux: You also never use `github_name` function which is needed by workflow to know package name/version specified later for the flist (used to specify the version/change the version without changing the workflow)

bins/packages/hd-idle/hd-idle.sh
zaibon: right specific version are better so we control exactly what run. Otherwise we can have different version running on different node and this makes a mess. 
But with go mod, now you can control the exact version of what you build, so something like `go get github.com/adelolmo/hd-idle@v1.2.3`

DylanVerstraete: @zaibon looks like that is not possible =/

```
➜  test go get github.com/adelolmo/hd-idle@v1.9     
go get github.com/adelolmo/hd-idle@v1.9: no matching versions for query "v1.9"
```

While the latest release is 1.9 https://github.com/adelolmo/hd-idle/tree/v1.9

bins/packages/hd-idle/hd-idle.sh
```diff
@@ -9,7 +9,12 @@ dependencies_hd-idle() {
 
 download_hd-idle() {
     echo "[+] downloading and installing hd-idle"
-    go get github.com/adelolmo/hd-idle
+    go get ${HD_IDLE_REPOSITORY}@${HD_IDLE_VERSION}
+}
+
+prepare_hd-idle() {
+    echo "[+] prepare hd-idle"
+    github_name "hd-idle-${HD_IDLE_VERSION}"
```
maxux: I would recommend to use `${HD_IDLE_VERSION:0:8}` to get a short id (otherwise filename will be large)

bins/packages/storcli/storcli.sh
```diff
@@ -1,17 +1,18 @@
+STORCLI_VERSION="1.21.06"
+
 dependencies_storcli() {
     apt-get update
```
maxux: This one is not needed, you re-do it after wget later

bins/packages/storcli/storcli.sh
```diff
@@ -1,17 +1,18 @@
...
 
     echo "[+] updating soruces list"
     wget -O - http://archive.thomas-krenn.com/tk-archive.gpg.pub | apt-key add -
-    cd /etc/apt/sources.list.d
-    wget http://archive.thomas-krenn.com/tk-main-xenial.list
-    wget http://archive.thomas-krenn.com/tk-optional-xenial.list
+
+    wget -O /etc/apt/sources.list.d/tk-main-xenial.list http://archive.thomas-krenn.com/tk-main-xenial.list
+    wget -O /etc/apt/sources.list.d/tk-optional-xenial.list http://archive.thomas-krenn.com/tk-optional-xenial.list
     cd
```
maxux: You don't change directory before, this could even do weird stuff with the `pushd` from `build_storcli`

bins/packages/storcli/storcli.sh
```diff
@@ -1,17 +1,18 @@
...
     apt-get update
 
     echo "[+] installing storcli"
-    apt-get install storcli
+    apt-get install storcli=${STORCLI_VERSION}
```
maxux: This could break if package is not available on the ubuntu repository anymore.

You should download it from: https://docs.broadcom.com/docs-and-downloads/raid-controllers/raid-controllers-common-files/007.1316.0000.0000_Unified_StorCLI_PUL.zip

This is the last version, more up-to-date than Ubuntu version (`1.21.06 Nov 18, 2016` vs `007.1316.0000.0000 Mar 12, 2020`)

Extract the zip, extract the `Unified_storcli_all_os/Linux/storcli-007.1316.0000.0000-1.noarch.rpm` and take the binary from there.

bins/packages/storcli/storcli.sh
```diff
@@ -1,17 +1,18 @@
+STORCLI_VERSION="1.21.06"
+
 dependencies_storcli() {
     apt-get update
-    apt-get install wget
 
     echo "[+] updating soruces list"
     wget -O - http://archive.thomas-krenn.com/tk-archive.gpg.pub | apt-key add -
```
maxux: Did you tested your script ? Because it doesn't works on a fresh Ubuntu 18.04 docker. Which mean it won't work neither on the github workflow. The `apt-key` command fails, you need to install `gnupg` (`apt-get install gnupg`) first.

maxux (changes requested):

muhamadazmy:
Please fix failing tests

pkg/storage/storage.go
```diff
@@ -119,13 +120,16 @@ func (s *storageModule) BrokenDevices() []pkg.BrokenDevice {
...
 func (s *storageModule) Dump() {
-	log.Debug().Int("volumes", len(s.volumes)).Msg("dumping volumes")
+	log.Debug().Int("volumes", len(s.pools)).Msg("dumping volumes")
 
-	for i := range s.volumes {
-		devices := s.volumes[i].Devices()
-		for id := range devices {
-			d := devices[id]
-			log.Debug().Str("path", d.Path).Str("label", d.Label).Str("type", string(d.DiskType)).Send()
+	for _, volume := range s.pools {
```
muhamadazmy: this should be also renamed to a `pool`

pkg/storage/storage.go
```diff
@@ -266,28 +275,56 @@ func (s *storageModule) initialize(policy pkg.StoragePolicy) error {
...
+
 	return s.ensureCache()
 }
 
+func (s *storageModule) shutdownDevice(devicePath string) error {
+	cmd := exec.Command("hdparm", "-y", devicePath)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
+	return cmd.Start()
```
muhamadazmy: Start does not wait for the command to finish execution. So your code will only print errors to stdout, but if the command fail to shutdown the device you will NOT receive an error back from this call. Instead you need to use cmd.Run() which will wait for the command to finish execution, and then return you an error if the command existed with a non 0 exit code.

```suggestion
	return cmd.Run()
```

pkg/storage/storage.go
```diff
@@ -325,15 +362,45 @@ func (s *storageModule) CreateFilesystem(name string, size uint64, poolType pkg.
...
+
+		// After releasing a filesystem by name, there is a chance there is going to be
+		// no reserved space on a volume, in this case we want to shutdown the disk in order to save power
+		volumes, err := pool.Volumes()
+		if err != nil {
+			log.Error().Err(err).Msgf("Failed to retrieve subvolumes on pool %s", pool.Name())
+		}
+		log.Debug().Msgf("Pool %s is has: %d subvolumes", pool.Name(), len(volumes))
+
+		if len(volumes) == 0 {
```
muhamadazmy: general comment:
I personally prefer negated checks (specially in loops) to reduce the amount of code indentation (and hence increase readability). 
for example i would do 
```go
if len(volumes) > 0 {
    continue
}
//rest of code can be put (un-indendent) here
```  
> You don't have to apply this code comment.

pkg/storage/storage.go
```diff
@@ -325,15 +362,45 @@ func (s *storageModule) CreateFilesystem(name string, size uint64, poolType pkg.
...
+			err = pool.UnMount()
+			if err != nil {
+				log.Error().Err(err).Msgf("Failed to unmount volume %s", pool.Name())
+				return err
+			}
+			for _, device := range pool.Devices() {
```
muhamadazmy: I think it's better if you introduce a `shutdownPool` method that takes a pool, and shutdown all pool devices. so you don't have this code everywhere.

pkg/storage/storage.go
```diff
@@ -483,39 +615,17 @@ func (s *storageModule) createSubvol(size uint64, name string, poolType pkg.Devi
 			continue
 		}
 
-		candidates = append(candidates, Candidate{
+		candidates = append(candidates, candidate{
```
muhamadazmy, line 603: This doesn't sound right. You should only add the disk to the candidates list ONLY if the disk has enough space for the requested size. So if the disk doesn't have enough space for that subvolume, it shouldn't be considered a candidate.

Also in the older code, the candidates where sorted so the one with the most disk space available is at the head of the candidate list so it gets to be tried first.

muhamadazmy (changes requested):

pkg/storage/filesystem/btrfs.go
```diff
@@ -457,6 +458,23 @@ func (p *btrfsPool) Maintenance() error {
...
 }
 
+func (p *btrfsPool) Shutdown() error {
+	for _, device := range p.Devices() {
+		log.Info().Msgf("Shutting down disk %s ...", device.Path)
+		cmd := exec.Command("hdparm", "-y", device.Path)
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+
+		err := cmd.Start()
```
muhamadazmy: This should be `cmd.Run()` please check the other (probably outdated) comment

pkg/storage/storage.go
DylanVerstraete: There is a Shutdown method on a pool now, this makes it easier

pkg/storage/storage.go
DylanVerstraete: This is already implemented? https://github.com/threefoldtech/zos/blob/93ff52177cd6cfc3292cef81e181cf9e3ce088d7/pkg/storage/storage.go#L603-L611

pkg/storage/filesystem/btrfs.go
```diff
@@ -220,6 +221,24 @@ func (p *btrfsPool) Mount() (string, error) {
 	return mnt, p.utils.QGroupEnable(ctx, mnt)
 }
 
+// Mount mounts the pool in it's default mount location under /mnt/name
```
zaibon:
```suggestion
// MountWithoutScan mounts the pool in it's default mount location under /mnt/name
```

pkg/storage/filesystem/btrfs.go
```diff
@@ -457,6 +476,23 @@ func (p *btrfsPool) Maintenance() error {
 	return nil
 }
 
+func (p *btrfsPool) Shutdown() error {
+	for _, device := range p.Devices() {
+		log.Info().Msgf("Shutting down disk %s ...", device.Path)
+		cmd := exec.Command("hdparm", "-y", device.Path)
+		cmd.Stdout = os.Stdout
```
zaibon: You don't seem to use threse streams, do you really need to link them to stdout and stderr ?

pkg/storage/filesystem/filesystem.go
```diff
@@ -37,6 +37,8 @@ type Pool interface {
 	Mounted() (string, bool)
 	// Mount the pool, the mountpoint is returned
 	Mount() (string, error)
+	// Mount the pool, the mountpoint is returned
```
zaibon: adapt the comment please.

pkg/storage/storage.go
```diff
@@ -87,13 +87,18 @@ func (s *storageModule) Total(kind pkg.DeviceType) (uint64, error) {
 	defer s.mu.RUnlock()
 	var total uint64
 
-	for idx := range s.volumes {
+	for _, pool := range s.pools {
+		// if pool is not mounted, we can't check total amount of storage
+		if _, mounted := pool.Mounted(); !mounted {
```
zaibon: I have a problem with this. The Total method is used by capacity to detect how much HRU and SRU is available on the node:
- https://github.com/threefoldtech/zos/blob/79a0e1df3d70b3233026c138cf12d67252b32946/pkg/capacity/resource.go#L27
- https://github.com/threefoldtech/zos/blob/79a0e1df3d70b3233026c138cf12d67252b32946/pkg/capacity/resource.go#L36

So if we skip disks here, we won't count the right amount of storage for each disk. Which in turn will false the minting.

pkg/storage/storage.go
```diff
@@ -159,24 +167,12 @@ func (s *storageModule) initialize(policy pkg.StoragePolicy) error {
...
-			continue
-		}
-		_, err = volume.Mount()
+	for _, pool := range existingPools {
+		_, err := pool.Mount()
 		if err != nil {
-			s.brokenPools = append(s.brokenPools, pkg.BrokenPool{Label: volume.Name(), Err: err})
-			log.Warn().Msgf("Failed to mount volume %v", volume.Name())
-			continue
+			return err
```
zaibon: Why don't we keep track of the broken pools anymore ? I don't think you can get rid of this.

pkg/storage/storage.go
```diff
@@ -159,24 +167,12 @@ func (s *storageModule) initialize(policy pkg.StoragePolicy) error {
...
-			// volume is already mounted, skip mounting it again, make sure it is
-			// in the list of available pools. Since we are in the initialize method,
-			// we can safely assume the pool has not been added before, so no need
-			// to check for duplicate entries.
-			s.volumes = append(s.volumes, volume)
-			continue
-		}
-		_, err = volume.Mount()
+	for _, pool := range existingPools {
+		_, err := pool.Mount()
```
zaibon, line 180: Is Mount now idempotent and check if it is already mounted automatically ?

pkg/storage/storage.go
```diff
@@ -266,28 +262,61 @@ func (s *storageModule) initialize(policy pkg.StoragePolicy) error {
...
+
+	return nil
+}
+
+func (s *storageModule) shutdownUnusedPools() error {
+	for _, pool := range s.pools {
+		if _, mounted := pool.Mounted(); mounted {
+			volumes, err := pool.Volumes()
+			if err != nil {
+				log.Error().Err(err).Msgf("Failed to retrieve subvolumes on pool %s", pool.Name())
```
zaibon, line 306: is that ok that you just log the error here and not return it ?
cause you do use volume line 299 while if err is not null, I guess volumes is undefined

pkg/storage/storage.go
```diff
@@ -266,28 +262,61 @@ func (s *storageModule) initialize(policy pkg.StoragePolicy) error {
...
+			log.Error().Err(err).Msgf("Error shutting down pool %s", pool.Name())
+		}
+	}
+	return nil
 }
 
 func (s *storageModule) Maintenance() error {
-	for _, pool := range s.volumes {
+	for _, pool := range s.pools {
+		// no need to run maintenance on unmounted pools
```
zaibon: Does maintenance is then called whenever we mount the pool again ?

pkg/storage/storage.go
```diff
@@ -325,17 +354,31 @@ func (s *storageModule) CreateFilesystem(name string, size uint64, poolType pkg.
...
+				err = pool.RemoveVolume(fs.Name())
+				if err != nil {
+					return err
+				}
 			}
 		}
+
+		// After releasing a filesystem by name, there is a chance there is going to be
+		// no reserved space on a volume, in this case we want to shutdown the disk in order to save power
+		if err := s.shutdownUnusedPools(); err != nil {
```
zaibon: do you have to call `shutdownUnusedPools` which loop over everything here ? Can't you just check if `filesystems` had a length of 1 and if this is the case, then we know there are no more subvolume and we can just call `pool.Shutdown` directly ?

pkg/storage/storage.go
```diff
@@ -446,20 +489,80 @@ func (s *storageModule) createSubvol(size uint64, name string, poolType pkg.Devi
 		return nil, pkg.ErrInvalidDeviceType{DeviceType: poolType}
 	}
 
-	type Candidate struct {
-		Pool      filesystem.Pool
-		Available uint64
+	// Look for candidates in mounted pools first
+	candidates, err := s.checkForCandidates(size, poolType, true)
+	if err != nil {
+		log.Error().Err(err).Msgf("failed to search conditation on unmounted pools")
```
zaibon: I think you should return an error here.

pkg/storage/storage.go
```diff
@@ -446,20 +489,80 @@ func (s *storageModule) createSubvol(size uint64, name string, poolType pkg.Devi
...
+	}
+
+	log.Debug().Msgf("Found %d candidates in mounted pools", len(candidates))
+
+	// If no candidates or found in mounted pools, we check the unmounted pools and get the first one that fits
+	if len(candidates) == 0 {
+		log.Debug().Msg("Checking unmounted pools")
+		candidates, err = s.checkForCandidates(size, poolType, false)
+		if err != nil {
+			log.Error().Err(err).Msgf("failed to search conditation on unmounted pools")
```
zaibon: I think you should return an error here.

pkg/storage/storage.go
```diff
@@ -446,20 +489,80 @@ func (s *storageModule) createSubvol(size uint64, name string, poolType pkg.Devi
...
+}
+
+func (s *storageModule) checkForCandidates(size uint64, poolType pkg.DeviceType, mounted bool) ([]candidate, error) {
+	var candidates []candidate
+	for _, pool := range s.pools {
+		_, poolIsMounted := pool.Mounted()
+		if mounted != poolIsMounted {
+			continue
+		}
+		log.Debug().Msgf("checking pool %s for space", pool.Name())
```
zaibon: let's move this log line after the check of type, so we don't print that we check on a certain pool while we actually jut skip it.

pkg/storage/storage.go
```diff
@@ -446,20 +489,80 @@ func (s *storageModule) createSubvol(size uint64, name string, poolType pkg.Devi
...
 		if pool.Type() != poolType {
 			continue
 		}
 
+		if !poolIsMounted && !mounted {
+			log.Debug().Msgf("Mounting pool %s...", pool.Name())
+			// if the pool is not mounted, and we are looking for not mounted pools, mount it first
+			_, err := pool.MountWithoutScan()
+			if err != nil {
+				log.Error().Err(err).Msgf("failed to mount pool %s", pool.Name())
```
zaibon, line 561: quite sure you must return an error here too, or at least skip this pool, cause after you call `pool.Usage` which I'm sure will crash on a unmounted pool.

pkg/storage/storage.go
```diff
@@ -483,39 +586,17 @@ func (s *storageModule) createSubvol(size uint64, name string, poolType pkg.Devi
 			continue
 		}
 
-		candidates = append(candidates, Candidate{
+		candidates = append(candidates, candidate{
 			Pool:      pool,
 			Available: usage.Size - (reserved + size), // available after new subvolume
```
zaibon: Why is Available is compute with the new reserved size substracted ? We want to know how much storage is available **before** we reserve the new sub-volume.

pkg/storage/storage.go
```diff
@@ -483,39 +586,17 @@ func (s *storageModule) createSubvol(size uint64, name string, poolType pkg.Devi
...
-		if err != nil {
-			log.Error().Err(err).Str("pool", candidate.Pool.Name()).Msg("failed to create new filesystem")
-			continue
-		}
-		if err = volume.Limit(size); err != nil {
-			candidate.Pool.RemoveVolume(volume.Name()) // try to recover
-			log.Error().Err(err).Str("volume", volume.Path()).Msg("failed to set volume size limit")
-			continue
+		// if we are looking for not mounted pools, break here
+		if !mounted {
```
zaibon, line 609: What's the point of this early return ?
Also what happens whit the previously unmounted pool that got mounted here, but had a too small available space ? Shouldn't they be unmounted and shutdown again ?

pkg/storage/zdb.go
```diff
@@ -98,34 +101,90 @@ func (s *storageModule) Allocate(nsID string, diskType pkg.DeviceType, size uint
...
 
-	var candidates []Candidate
-	// okay, so this is a new allocation
-	// we search all the pools for a zdb subvolume that can
-	// hold the new namespace
-	for _, pool := range s.volumes {
-		// skip pool with wrong disk type
-		if pool.Type() != diskType {
+	// check for candidates in mounted pools first
+	candidates, err := s.checkForZDBCandidates(size, diskType, true, targetMode)
```
zaibon: is it normal we do not look on the unmounted pool if there are not candidate withint the mounted pool ?

zaibon (changes requested):

pkg/storage/storage.go
DylanVerstraete: This Total method seems to be executed sometimes, but when the disks are spun down it's impossible to check the available space. We should come up with a solution for this.. I think we can atleast hold the the total space of a disk in memory when we first start storaged.

pkg/storage/storage.go
DylanVerstraete: I don't understand your question. but yeah mount checks if it is already mounted, if so, it returns the mount path

pkg/storage/storage.go
DylanVerstraete: No, but maybe we should!

pkg/storage/storage.go
DylanVerstraete: Yes, good point. The early return was to make sure as we found the first candidate it would be immediatly returned.

pkg/storage/zdb.go
DylanVerstraete: Because when it cannot find candidates in mounted pools, then it calls the storagemodule to create a subvol. It's on line 133

pkg/provision/primitives/zdb.go
```diff
@@ -378,6 +379,58 @@ func (p *Provisioner) zdbDecommission(ctx context.Context, reservation *provisio
...
+
+		if err := flist.Umount(rootFS); err != nil {
+			return errors.Wrapf(err, "failed to unmount flist at %s", rootFS)
+		}
+
+	} else {
+		log.Error().Err(err).Str("container", string(containerID)).Msg("failed to inspect container for decomission")
+	}
+
+	// TODO: delete network?
```
zaibon, line 432: @DylanVerstraete did you confirm that leaving the network namespace is working and that it gets re-used when a new 0-db container is restarted on the same disk ?

DylanVerstraete: How can I check?

pkg/storage/filesystem/btrfs.go
```diff
@@ -457,6 +484,21 @@ func (p *btrfsPool) Maintenance() error {
 	return nil
 }
 
+func (p *btrfsPool) Shutdown() error {
+	for _, device := range p.Devices() {
+		log.Info().Msgf("Shutting down disk %s ...", device.Path)
+		cmd := exec.Command("hdparm", "-y", device.Path)
```
zaibon, line 490: @DylanVerstraete @delandtj What happens if the nodes has some disk controller that cannot be controlled with hdparm ?

pkg/storage/filesystem/filesystem.go
```diff
@@ -47,13 +50,6 @@ type Pool interface {
 	Type() pkg.DeviceType
 	// Reserved is reserved size of the devices in bytes
 	Reserved() (uint64, error)
-	// Maintenance is a routine that is called at boot
```
zaibon, line 50: Why is this remove from the interface?

zaibon (approved):

pkg/storage/filesystem/filesystem.go
DylanVerstraete: It became a private function

pkg/storage/filesystem/btrfs.go
DylanVerstraete: Then this disk wont be shutdown, this needs to be investigated indeed

pkg/provision/primitives/zdb.go
DylanVerstraete: Yes

PR #869, opened 2020-07-28
flistd cleanup

### LINKED ISSUE #864: flistd should watch the 0-fs processes and cleanup after they stops
At the moment, flistd just fires the 0-fs process in daemon mode and forgets about it.
This is a problem when a 0-fs crashes and leave the mountpoint dirty. flistd should watch the 0-fs processes and clean up behind it if needed.

muhamadazmy:
Fixes #864

- We make sure that mounted flists names are  tracked 
- We can't use pid files because they are deleted if the g8ufs process exited gracefully
- We do the tracking by creating a symlink to the pid file
- A go routine is running (every minute) that does the following
  - list all tracking symlinks
  - if link target is missing (graceful shutdown) the clean up is run anyway just to be sure
  - if link target is there, the PID is loaded
  - we check the PID that is actually a g8ufs process
  - If process exists, nothing is done
  - Otherwise we run the clean up
- The clean up process includes
  - force unmount the target location
  - drop the storage volume used for backend
  - delete all pid, tracking and log files.

### DIFF
--- a/pkg/flist/cleanup.go
+++ b/pkg/flist/cleanup.go
@@ -6,7 +6,6 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"strings"
 	"syscall"
 	"time"
 
@@ -37,18 +36,15 @@ func (f *flistModule) listMounts() (map[string]int64, error) {
 			continue
 		}
 
-		if !strings.HasSuffix(info.Name(), ".pid") {
-			// also what is that doing here!
-			continue
-		}
-
 		path := filepath.Join(f.run, info.Name())
 		pid, err := f.getPid(path)
-		if err != nil {
+		if os.IsNotExist(err) {
+			pid = -1
+		} else if err != nil {
 			log.Error().Err(err).Str("pid-file", path).Msg("invalid pid file")
 		}
 
-		data[strings.TrimSuffix(info.Name(), ".pid")] = pid
+		data[info.Name()] = pid
 	}
 
 	return data, nil
@@ -76,6 +72,16 @@ func (f *flistModule) cleanupMount(name string) error {
 		log.Warn().Err(err).Str("subvolume", name).Msg("failed to clean up subvolume")
 	}
 
+	for _, file := range []string{
+		filepath.Join(f.pid, fmt.Sprintf("%s.pid", name)),
+		filepath.Join(f.run, name),
+		filepath.Join(f.log, fmt.Sprintf("%s.log", name)),
+	} {
+		if err := os.Remove(file); err != nil {
+			log.Warn().Err(err).Str("file", file).Msg("failed to delete pid file")
+		}
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


zaibon:
I'm not fully certain this is the right way to do it. I'm concerned about the fact that this will always delete the subvolume for any stop 0-fs process. While in case of a crash, we do NOT want to loose the data that is stored in the backend. Only when the container using the flist is explicitly deleted, then we can also delete the subvolume.

muhamadazmy:
@zaibon Yeah, I think you are right, I will drop this part and just make sure that the ro layer is unmounted cleanly

muhamadazmy:
@zaibon applied some changes based on your comment. Please take another look :)

pkg/flist/cleanup.go
```diff
@@ -0,0 +1,140 @@
...
+		log.Warn().Err(err).Str("path", path).Msg("fail to unmount flist")
+	}
+
+	volume, err := f.storage.Path(name)
+	if err != nil {
+		log.Warn().Err(err).Str("subvolume", name).Msg("subvolume does not exist")
+		return nil
+	}
+
+	ro := filepath.Join(volume, "ro")
```
zaibon, line 87: we only need to umount the ro layer ? what about the others? Or do we assume overlayfs always clean up properly ?

pkg/flist/flist.go
```diff
@@ -197,13 +202,22 @@ func (f *flistModule) mount(name, url, storage string, opts pkg.MountOptions) (s
...
+	}
 
-		args = append(args, "-backend", path)
+	if len(backend) != 0 {
+		args = append(args, "-backend", backend)
+		// in case of an error (mount is never fully completed)
+		// we need to deallocate the filesystem
+		defer func() {
+			if err != nil {
+				f.storage.ReleaseFilesystem(name)
```
zaibon, line 224: You still cannot do that here. Cause it might be you are trying to mount an flist after a reboot for which there is actual data on the backend. So even here I don't think we can release the filesystem.

muhamadazmy: I think with some code reorganization we can grantee this is a new allocation that was never get used. There is already a check that this is already a valid mount, that I think should be moved a bit higher.

maxux:
The workflow to check if mount is ready and everything is okay seems quite hackish.

Why not implement in 0-fs to fork when everything is ready. So you can just start the process, wait for it to ends. When it ended, you check if everything was good or if an error occurred.

If 0-fs cleanup itself when closing it's stuff, why cleanup should happen in the back ? Except if 0-fs segfault, and even, we can umount gracefully on segfault. No ?

It's still useful to cleanup logs, etc. btw, yes

pkg/flist/cleanup.go
muhamadazmy: There are no others, there is only the "mountpoint" which is fuse, which is unmounted on line 77. then only the fuse is mounted under the "ro". the other directories "wd", and "rw" are just folders with data in them

muhamadazmy:
@maxux It's indeed hackish, but this is not new code "waitingForPid" and checking for the "mount ready" in logs. So I will create a new issue to fix this, but it's not related to this PR.

The code is idd related to unexpected failures of the mount that was not initiated by the flitstd itself.

pkg/flist/cleanup.go
zaibon: I don't think that's true. overlayfs is also mounted by 0-fs. So there are 2 mounted filesystem:
- 0-fs which provide the ro layer
- overlays which user the ro and mount on `mountpoint`

What about overlayfs ? we don't deal with it ?

pkg/flist/flist.go
zaibon: yep let's do that.

pkg/flist/cleanup.go
muhamadazmy: Overlayfs is only mounted if you mount the fs in read-write mode. because that's it's use case, if you start the fs in read-only mode it only mount the fuse directly under your target mountpoint.

muhamadazmy: I think I get your question now, the `ro` directory is the only mounted path under the backend, the overlay fs is mounted under the mount target that has been unmounted already. 
In some cases (ro mount of 0fs) the fuse is mounted directly under the target, hence nothing to do under the volume (it will not exist already)

muhamadazmy:
Please make sure this is only merged AFTER releasing this 0fs PR  https://github.com/threefoldtech/0-fs/pull/36

zaibon (approved):

PR #883, opened 2020-08-05
Adds cleanup of provisiond unused resources

### LINKED ISSUE #709: provisiond: self cleanup
In certain situation (machine reboot, etc...) node fails to gracefully clean up some persisted resources (this usually subvolumes, or a zdb namespace)

Currently we only store a cached version of active reservations, but this is not enough because a reservation does not reflect actual resources allocated on the node. Basically, there is no way the you can map from a reservation to actual allocated volume, or namespace, etc...

We still can do some intensive reverse lookup to link a resource to a reservation. But this will probably be not generic and very intense.

Suggested solution:
- the resource provision should return the result of the reservation (As already done now) plus a set of allocated reserouces
- the provisiond stores the cached resource as it now, plus the allocated resource.
- every few hours, provisiond pause receiving new reservation, and start the housekeeping.
  - List all active reservations, and list all allocated resources. It does substraction and it will end up with a list of stall resources that doesn't link back to a reservation
  - It does a force clean up of the resource
- resume normal operation

### LINKED ISSUE #879: Btrfs subvolumes are not cleaned up
Since we want to add the disk shutdown feature we need to make sure there are no subvolumes on a disk before shutting it down. But right now we cannot do that because some subvolumes are never cleaned up.

For example if you reserve a zdb it creates a subvolumes on a disk, but when this reservation is expired or canceled this subvolumes is never deleted. Hence we can't shutdown the disk if it was the only subvolume on there..

For volume reservations it works fine, subvolumes are cleaned up after they are expired or canceled. I wonder if we can do the same for zdb's.

Also we need to find a way to clean up existing subvolumes on existing nodes..

DylanVerstraete:


pkg/provision/cleanup.go
```diff
@@ -0,0 +1,230 @@
...
+		} `json:"root"`
+		Mounts []struct {
+			Source string `json:"source"`
+		} `json:"mounts"`
+	} `json:"Spec"`
+}
+
+// we declare this method as a variable so we can
+// mock it in testing.
+var zdbConnection = func(id string) zdb.Client {
```
muhamadazmy: I don't believe this hack is needed anymore (I mean declaring zdbConnection as a var not a func)

pkg/provision/cleanup.go
```diff
@@ -0,0 +1,230 @@
...
+		for _, cnt := range lines(must(execute("ctr", "-n", ns, "container", "ls", "-q"))) {
+			bytes := must(execute("ctr", "-n", ns, "container", "info", cnt))
+			var container container
+			if err := json.Unmarshal(bytes, &container); err != nil {
+				panic(err)
+			}
+
+			zdbCl := zdbConnection(container.ID)
+			defer zdbCl.Close()
+			if err := zdbCl.Connect(); err != nil {
```
muhamadazmy: I can see you list all containers in ALL namespaces, so I am sure not all of them are ZDB containers. no ?

pkg/provision/cleanup.go
```diff
@@ -0,0 +1,230 @@
...
+			}
+
+			ns, err := zdbCl.Namespaces()
+			if err != nil {
+				return errors.Wrap(err, "failed to retrieve zdb namespaces")
+			}
+
+			toSave[filepath.Base(container.Spec.Root.Path)] = struct{}{}
+			for _, mnt := range container.Spec.Mounts {
+				if strings.HasPrefix(mnt.Source, "/mnt/") {
```
muhamadazmy: Personally I don't really like deep indented code. I would instead do 
```go
if !string.HasPrefix(...) {
   continue
}
```

pkg/provision/cleanup.go
muhamadazmy: If you want to process ALL zdb containers only list only containers from the `zdb` namespace

pkg/provision/cleanup.go
```diff
@@ -0,0 +1,230 @@
...
+	socket := fmt.Sprintf("unix://%s/zdb.sock", socketDir(id))
+	return zdb.New(socket)
+}
+
+func socketDir(containerID string) string {
+	return fmt.Sprintf("/var/run/zdb_%s", containerID)
+}
+
+// CleanupResources cleans up unused resources
+func CleanupResources() error {
```
muhamadazmy: This is a very big function, can we split the functionality into smaller functions, this will also help understanding the logic if u use descriptive function names.

pkg/provision/cleanup.go
```diff
@@ -0,0 +1,230 @@
...
+				if err := utils.QGroupDestroy(ctx, qgroup.ID, path); err != nil {
+					log.Err(err).Msgf("failed to delete qgroup: '%s'", qgroup.ID)
+				}
+				continue
+			}
+
+			// we only handle volumes that are 256MiB or 10MiB
+			if qgroup.MaxRfer != 268435456 && qgroup.MaxRfer != 10485760 {
+				// if the subvolume is a zdb and has 0 maxRfer, don't skip here. It might need to be deleted
+				if !strings.HasPrefix(subvol.Path, "zdb") && qgroup.MaxRfer != 0 {
```
muhamadazmy: zdb volumes never have size limit as far as I remember. I don't also get what you skip zdb subvolumes that has size limit ?

pkg/provision/cleanup.go
```diff
@@ -0,0 +1,230 @@
...
+				log.Err(err).Msgf("failed to delete qgroup: '%s'", qgroup.ID)
+			}
+		}
+	}
+
+	return nil
+}
+
+func deleteZdbContainer(containerID pkg.ContainerID) error {
+	zbusCl, err := zbus.NewRedisClient("unix:///var/run/redis.sock")
```
muhamadazmy: this should never be hardcoded, already provisiond accepts this as argument

pkg/provision/cleanup.go
```diff
@@ -0,0 +1,230 @@
...
+	if err != nil {
+		log.Fatal().Err(err).Msg("fail to connect to message broker server")
+	}
+
+	container := stubs.NewContainerModuleStub(zbusCl)
+	flist := stubs.NewFlisterStub(zbusCl)
+
+	info, err := container.Inspect("zdb", containerID)
+	if err == nil {
```
muhamadazmy: Again, this is a very weird (and imho ugly) a better way to rewrite this is 
```go
if err != nil { // shouldn't this check that it's a "not exist error" 
    // log
    return nil 
}

// rest of the logic that u have in the `err == nil` 
```

pkg/provision/cleanup.go
```diff
@@ -0,0 +1,230 @@
...
+func deleteZdbContainer(containerID pkg.ContainerID) error {
+	zbusCl, err := zbus.NewRedisClient("unix:///var/run/redis.sock")
+	if err != nil {
+		log.Fatal().Err(err).Msg("fail to connect to message broker server")
+	}
+
+	container := stubs.NewContainerModuleStub(zbusCl)
+	flist := stubs.NewFlisterStub(zbusCl)
+
+	info, err := container.Inspect("zdb", containerID)
```
muhamadazmy: May be you should add an `Exists` method to the container API. because inspect can return other errors as well.

pkg/provision/cleanup.go
```diff
@@ -0,0 +1,230 @@
...
+
+			zdbCl := zdbConnection(container.ID)
+			defer zdbCl.Close()
+			if err := zdbCl.Connect(); err != nil {
+				return errors.Wrapf(err, "failed to connect to 0-db: %s", container.ID)
+			}
+
+			ns, err := zdbCl.Namespaces()
+			if err != nil {
+				return errors.Wrap(err, "failed to retrieve zdb namespaces")
```
muhamadazmy: General notice about early returns here. This is after all a clean up method, so it should do its best to bring the system to a consistent and clean state, right? 
It means it should be able to "skip" some non fatal issues. For example if a zdb instance is not responding this shouldn't stop the system from checking other instances.

And next time this clean up will run it will recheck this lingering one

pkg/provision/engine.go
```diff
@@ -83,6 +85,25 @@ func (e *Engine) Run(ctx context.Context) error {
 
 	cReservation := e.source.Reservations(ctx)
 
+	after := time.After(5 * time.Minute)
```
muhamadazmy: This entire section logic is wrong. We need to make sure the clean up procedure pauses provisioning it means it has to be triggered inside the for/select loop otherwise the clean up will be running while the provisining is also "not blocked" .

As a solution i suggest the following (pseudo)
```go
cleanUp := make(chan struct{})
c := cron.New() 
c.AddFunc("@midnight", func() {
   cleanUp <- struct{}{}
})

go func() {
   <- time.After(5*time.Minute)
   cleanUP <- struct{}{}
}()

// in the loop you use 
// case <- cleanUp:
   // call CleanupResources() function
```

muhamadazmy (changes requested):

pkg/provision/cleanup.go
DylanVerstraete: Yeah that might be nice, but this should not be added in this PR

pkg/provision/cleanup.go
```diff
@@ -0,0 +1,250 @@
...
+					log.Err(err).Msgf("failed to delete qgroup: '%s'", qgroup.ID)
+				}
+				continue
+			}
+
+			// if the subvolume is a zdb and has 0 maxRfer, don't skip here. It might need to be deleted
+			zdbVolume := strings.HasPrefix(subvol.Path, "zdb") && qgroup.MaxRfer != 0
+
+			// we only handle volumes that are 256MiB or 10MiB
+			if qgroup.MaxRfer != 268435456 && qgroup.MaxRfer != 10485760 && !zdbVolume {
```
zaibon: Why do we only check volumes that have these specific size ?

pkg/provision/cleanup.go
```diff
@@ -0,0 +1,250 @@
...
+			}
+
+			toSave[filepath.Base(container.Spec.Root.Path)] = struct{}{}
+			for _, mnt := range container.Spec.Mounts {
+				if !strings.HasPrefix(mnt.Source, "/mnt/") {
+					continue
+				}
+
+				if len(ns) == 1 && namespaces[0] == "default" {
+					err := deleteZdbContainer(msgBrokerCon, pkg.ContainerID(container.ID))
```
zaibon: Do you still have to do that now that zdb decommission code already takes care of that ?

pkg/provision/cleanup.go
```diff
@@ -0,0 +1,250 @@
...
+
+// checks running containers for subvolumes that might need to be saved because they are used
+// and subvolumes that might need to be deleted because they have no attached container anymore
+func checkContainers(msgBrokerCon string) (map[string]struct{}, map[string]struct{}, error) {
+	toSave := make(map[string]struct{})
+	toDelete := make(map[string]struct{})
+
+	for _, ns := range lines(must(execute("ctr", "namespace", "ls", "-q"))) {
+		log.Info().Msgf("Checking namespace %s", ns)
+		for _, cnt := range lines(must(execute("ctr", "-n", ns, "container", "ls", "-q"))) {
```
zaibon: Is there any reason you use ctr CLI instead of talking directly to containerd?

pkg/provision/engine.go
```diff
@@ -109,20 +136,31 @@ func (e *Engine) Run(ctx context.Context) error {
 				slog.Info().Msg("start decommissioning reservation")
 				if err := e.decommission(ctx, reservation); err != nil {
 					log.Error().Err(err).Msgf("failed to decommission reservation %s", reservation.ID)
+					canCleanup = true
```
zaibon: so if I understand, as soon as one reservation is processed, we allow the clean up to run ?
I think this approach is still a bit dangerous, cause in the case of a reboot, nothing ensure you that all the workloads that should be provisioned actually are. So to me we need to have a way to explicitly know that we did process all the workloads before we allow cleanup to run.

zaibon (changes requested):
Added some comments.

Also, is the cleanup of VMs and networks is going to be part of this PR or not ?

pkg/provision/cleanup.go
DylanVerstraete: Yeah, to clean up running containers that have no zdb volumes so we can shutdown the disks afterwards.

pkg/provision/cleanup.go
DylanVerstraete: No not really, this is how azmy did it. Should I refactor this?

pkg/provision/engine.go
DylanVerstraete: How can we know provisiond is done with redeploying after a reboot?

pkg/provision/cleanup.go
muhamadazmy: I did it in a one time cleaner script that we need to run to clean up some nodes. I think now we can indeed use containerd directly.

pkg/provision/engine.go
muhamadazmy: I don't think "canCleanup" is needed at all. IF you are in the select statement only once case can run at a time. I don't see why you still have this after chaning to use channels

pkg/provision/engine.go
```diff
@@ -109,20 +136,31 @@ func (e *Engine) Run(ctx context.Context) error {
...
 
 			if err := e.updateStats(); err != nil {
 				log.Error().Err(err).Msg("failed to updated the capacity counters")
 			}
+			canCleanup = true
 
+		case <-cleanUp:
+			log.Info().Msg("start cleaning up resources")
+			if canCleanup {
+				if err := CleanupResources(e.msgBrokerCon); err != nil {
```
muhamadazmy: CleanupResources doesn't have to be public

pkg/provision/cleanup.go
muhamadazmy: I can't remember fully remember but the 250MB size is the default "container" rootfs volume so we sure it's created for that purpose. the 10MB check is similar because we use this size for zdb container rootfs.

We were trying to be as defensive as possible with the clean up

muhamadazmy: May be we don't need this anymore. If we make sure the volume is not used by a container or an active volume reservation

pkg/provision/engine.go
```diff
@@ -94,6 +120,7 @@ func (e *Engine) Run(ctx context.Context) error {
 				log.Info().Msg("reservation source is emptied. stopping engine")
 				return nil
 			}
+			canCleanup = false
```
muhamadazmy: the `canCleanup` flag is not needed anymore

pkg/provision/cleanup.go
DylanVerstraete: refactored :)

pkg/provision/cleanup.go
DylanVerstraete: I would keep it anyway, as you said. Try to be defensive

zaibon: This code was extracted from a script that has a very different purpose. We are now trying to get something the cleans up everything that should not be there.

pkg/provision/cleanup.go
zaibon: But you already do that on the zdb code now, right? So why doing it here again?

pkg/provision/engine.go
zaibon: > How can we know provisiond is done with redeploying after a reboot?

One way to do it would be to be sure that last reservation processed is also the most recent one existing on the explorer.

zaibon: > I don't think "canCleanup" is needed at all. IF you are in the select statement only once case can run at a time. I don't see why you still have this after chaning to use channels

Channel does solve the problem I'm rising. You process one reservation at a time. In the case the provisioning of all the required reservations would take more then 5 minutes. We could start deleting primitives that actually should be kept.

LeeSmet (approved):

zaibon:
I did a test with this code on one of my node and I saw this log line: 
`[+] provisiond: 2020-08-28T15:59:30Z info deleting subvolume zos-cache`
Which worries me a lot. I think we forgot to protect this special subvolume.

zaibon (changes requested):
I went over the code again and I think I've remark another problem. From what I understand the clean up code does remove any subvolume that is not used by any running container.
2 Problems with this approach:
- the 5 minute delay before running the cleanup does not guarantee all containers that should be running are actually already provisioned. This trigger should not be based on time, but on the fact that provisiond knows he has processed all workloads that are ready to be deployed
- Second problem is there could be some volume that are still reserved, but just not mounted in any container. Those volume also cannot be deleted. 

So it seems the cleanup should actually base it's decision on the workloads detail from the explorer instead of what's running on the node. Unless I miss the intent

DylanVerstraete:
https://github.com/threefoldtech/tfexplorer/pull/179 needs to be merged first

pkg/provision/cleanup.go
```diff
@@ -67,50 +75,81 @@ func cleanupResources(msgBrokerCon string) error {
 
 		for _, subvol := range subvols {
 			log.Info().Msgf("checking subvol %s", subvol.Path)
-			_, ok := qgroups[fmt.Sprintf("0/%d", subvol.ID)]
+			// Don't delete zos-cache!
+			if subvol.Path == "zos-cache" {
```
zaibon: don't we have a constant somewhere for this `zos-cache` value ? probably in the `storage` package

pkg/provision/cleanup.go
```diff
@@ -67,50 +75,81 @@ func cleanupResources(msgBrokerCon string) error {
...
 		}
 	}
 
 	return nil
 }
 
+func checkReservationToDelete(path string, client *client.Client) bool {
+	log.Info().Msgf("checking explorer for reservation: %s", path)
+	reservation, err := client.Workloads.NodeWorkloadGet(path)
+	if err != nil {
```
zaibon, line 67: you need to check that the error return is indeed a not found error and not any other type of error (network connection or something else)

pkg/provision/cleanup.go
```diff
@@ -67,50 +75,81 @@ func cleanupResources(msgBrokerCon string) error {
...
 				// delete the subvolume
-				log.Info().Msgf("deleting subvolume %s", subvol.Path)
-				// if err := utils.SubvolumeRemove(ctx, filepath.Join(path, subvol.Path)); err != nil {
-				// 	log.Err(err).Msgf("failed to delete subvol '%s'", subvol.Path)
-				// }
-				// if err := utils.QGroupDestroy(ctx, qgroup.ID, path); err != nil {
-				// 	log.Err(err).Msgf("failed to delete qgroup: '%s'", qgroup.ID)
-				// }
+				delete := checkReservationToDelete(subvol.Path, client)
+				if delete {
```
zaibon: I don't get this block. It seems we check if we need to delete it, and if we do, then we skip instead of deleting it.
Then line 117 you recheck the same volume with the same function.
Why not moving the order a bit:
1. check if the volume need to be kept, if yes, continue
2. check if the volume needs to be deleted, if yes, check the explorer
this remove the double call to the explorer and we handle the most probable case first.

pkg/provision/cleanup.go
```diff
@@ -0,0 +1,270 @@
...
+				}
+
+				zdbNamespaces, err = zdbCl.Namespaces()
+				if err != nil {
+					log.Err(err).Msg("failed to retrieve zdb namespaces")
+					continue
+				}
+			}
+
+			spec, _ := ctr.Spec(ctx)
```
zaibon:
```suggestion
			spec, err := ctr.Spec(ctx)
```

pkg/provision/cleanup.go
```diff
@@ -0,0 +1,270 @@
...
+		ctx := namespaces.WithNamespace(context.Background(), ns)
+		crts, err := client.Containers(ctx, "")
+		if err != nil {
+			log.Err(err).Msgf("failed to list containers")
+			return nil, nil, err
+		}
+
+		var zdbNamespaces []string
+
+		for _, ctr := range crts {
```
zaibon: In this loop, why do we never return any error and just continue instead ?

pkg/provision/cleanup.go
```diff
@@ -0,0 +1,270 @@
...
+					toSave[filepath.Base(mnt.Source)] = struct{}{}
+				}
+			}
+		}
+	}
+
+	return toSave, toDelete, nil
+}
+
+func deleteZdbContainer(msgBrokerCon string, containerID pkg.ContainerID) error {
```
zaibon: to avoid code duplication, can't we reuse the [deleteZdbContainer](https://github.com/threefoldtech/zos/blob/7727b34f2d71c8ac0cee5bb97de11461bceb136a/pkg/provision/primitives/zdb.go#L418) function ? 
I can already see that the implementation here is not the same

pkg/provision/source.go
```diff
@@ -30,6 +30,13 @@ type ReservationPoller interface {
 	Poll(nodeID pkg.Identifier, from uint64) (reservations []*Reservation, lastID uint64, err error)
 }
 
+// ReservationFromSource wraps a reservation type and has
+// a boolean to indicate it is the last reservation
+type ReservationFromSource struct {
```
zaibon: I would rename this type to something a bit more explicit:
```suggestion
type ProvisionJob struct {
```

pkg/provision/source.go
```diff
@@ -30,6 +30,13 @@ type ReservationPoller interface {
...
+	*Reservation
```
zaibon: I don't think you need to have a pointer here. You already use a pointer of the ReservationFromSource in the channel.

pkg/provision/source.go
```diff
@@ -83,8 +90,15 @@ func (s *pollSource) Reservations(ctx context.Context) <-chan *Reservation {
...
 				return
 			default:
-				for _, r := range res {
-					ch <- r
+				for idx, r := range res {
+					reservation := ReservationFromSource{
+						r,
+						false,
+					}
+					if idx == len(res)-1 {
```
zaibon: I don't think this condition is correct. 
Isn't it better to do it like `if lastID == reservation.ID` ?
You also need to make sure that you only set the `last` field to true only once and once only for the full life duration of provisiond. The other time it will be trigger by the cron

zaibon (changes requested):

pkg/provision/cleanup.go
DylanVerstraete: We need the following:

1: check if the volume need to be kept, if yes, continue
2: check if the volume needs to be deleted, if yes, check the explorer
3: check if the volume is any other volume that is not present in the toSave and toDelete map. Since we create and fill these maps based on containers, but I have seen some cases where there would be other volumes as well which are not linked to containers

pkg/provision/cleanup.go
DylanVerstraete: If we return an error in this loop we will stop trying to contact any other container and this will prevent us from doing a partially cleanup. I don't know what is the best choice here.

pkg/provision/cleanup.go
zaibon: But shouldn't you also list all the subvolumes present on the node ? How can you find volume not attached with container otherwise ?

pkg/provision/cleanup.go
zaibon: ok so partial cleanup is the goal. That make sense then. I guess it's indeed better to clean as much as we can even if we hit some error along the way.

pkg/provision/cleanup.go
DylanVerstraete: That is what is happening right now, we list all volumes under `/mnt` and check if it belongs there or not.

pkg/provision/source.go
DylanVerstraete: LastID is an int and `reservation.ID` is a string. We cannot match on this.

DylanVerstraete: Length of an array - 1 points to the last element in that array no?

DylanVerstraete: > You also need to make sure that you only set the `last` field to true only once and once only for the full life duration of provisiond. The other time it will be trigger by the cron

I don't understand what you mean by this.

pkg/provision/cleanup.go
DylanVerstraete: If I reuse this code then I face endless import cycles.
```
home/dylan/Projects/zos/bin/provisiond
package github.com/threefoldtech/zos/cmds/provisiond
	imports github.com/threefoldtech/zos/pkg/provision
	imports github.com/threefoldtech/zos/pkg/provision/primitives
	imports github.com/threefoldtech/zos/pkg/provision: import cycle not allowed

```

Let's just copy the new code from the zdb primitives ?

pkg/provision/cleanup.go
DylanVerstraete: Cool, tested it and it seems to work! :)

pkg/provision/source.go
zaibon: > Length of an array - 1 points to the last element in that array no? 
it does but nothing tell you the explorer send you all the workloads in one request. So unless you have process the reservation that has the lastID reported by the explorer, you have not reach the end.

>LastID is an int and reservation.ID is a string. We cannot match on this.

I think you can: it's generated like this https://github.com/threefoldtech/zos/blob/7727b34f2d71c8ac0cee5bb97de11461bceb136a/pkg/provision/primitives/converter.go#L252

pkg/provision/engine.go
```diff
@@ -81,6 +87,23 @@ func (e *Engine) Run(ctx context.Context) error {
 
 	cReservation := e.source.Reservations(ctx)
 
+	// Indicates wether provisiond cleaned up resources after
+	// booting and provisioning all reservations
+	triggeredCleanupOnBoot := false
```
zaibon: I would actually have put this into the source.go file instead of here. But I guess it doesn't really matter.

zaibon (changes requested):

pkg/provision/engine.go
```diff
@@ -121,6 +144,19 @@ func (e *Engine) Run(ctx context.Context) error {
 				log.Error().Err(err).Msg("failed to updated the capacity counters")
 			}
 
+			if reservation.last && !triggeredCleanupOnBoot {
+				// Trigger cleanup by sending a struct onto the channel
+				cleanUp <- struct{}{}
```
zaibon: I wonder if there is not a chance of dead lock here. If the `cleanUp` channel is already full, then this line will block and prevent the execution from reading the`cleanUp` channel to allow to continue here => dead lock

zaibon (changes requested):

pkg/provision/engine.go
```diff
@@ -151,7 +151,7 @@ func (e *Engine) Run(ctx context.Context) error {
 				triggeredCleanupOnBoot = true
 			}
 
-		case <-cleanUp:
+		case v := <-cleanUp:
```
zaibon: I don't think you need to assign the value, it's an empty struct.
```suggestion
		case  <-cleanUp:
```

zaibon (changes requested):

pkg/provision/cleanup.go
```diff
@@ -0,0 +1,234 @@
...
+	"github.com/threefoldtech/zos/pkg/zdb"
+	"golang.org/x/net/context"
+)
+
+const (
+	containerdSock = "/run/containerd/containerd.sock"
+)
+
+// we declare this method as a variable so we can
+// mock it in testing.
```
muhamadazmy: Wrong comment! it's obviously not true anymore

pkg/provision/cleanup.go
```diff
@@ -0,0 +1,234 @@
...
+		log.Info().Msgf("subvolume with path %s has next action to delete", path)
+		return true
+	}
+
+	return false
+}
+
+// checks running containers for subvolumes that might need to be saved because they are used
+// and subvolumes that might need to be deleted because they have no attached container anymore
+func checkContainers(msgBrokerCon string) (map[string]struct{}, map[string]struct{}, error) {
```
muhamadazmy: I would instead pass the connection to zbus and not the connection url. Hence i don't have to create a new connection pool every time. Possible leakage.

pkg/provision/cleanup.go
```diff
@@ -0,0 +1,234 @@
...
+		return nil, nil, err
+	}
+
+	client, err := containerd.New(containerdSock)
+	if err != nil {
+		log.Err(err).Msgf("failed to create containerd connection")
+		return nil, nil, err
+	}
+
+	ns, err := client.NamespaceService().List(context.Background())
```
muhamadazmy: create one context `ctx := context.Background()` and then pass it to all methods that needs it. Or even better, change the function signature `checkContainers` to accept a context object.

pkg/provision/cleanup.go
```diff
@@ -0,0 +1,234 @@
...
+			return nil, nil, err
+		}
+
+		var zdbNamespaces []string
+
+		for _, ctr := range crts {
+			log.Info().Msgf("container ID %s", ctr.ID())
+			if ns == "zdb" {
+				zdbCl := initZdbConnection(ctr.ID())
+				defer zdbCl.Close()
```
muhamadazmy: defer can't be done in a loop (check https://blog.learngoprogramming.com/gotchas-of-defer-in-go-1-8d070894cb01) 
instead, move the logic of this `if` statment to a function and call it here. The function then can have the loop.

pkg/provision/cleanup.go
```diff
@@ -0,0 +1,234 @@
...
+				continue
+			}
+
+			toSave[filepath.Base(spec.Root.Path)] = struct{}{}
+			for _, mnt := range spec.Mounts {
+				if !strings.HasPrefix(mnt.Source, "/mnt/") {
+					continue
+				}
+
+				if len(ns) == 1 && zdbNamespaces[0] == "default" {
```
muhamadazmy: This will cause a PANIC if `ns != "zdb"` because zdbNamespaces will be of length zero! The reason this might have worked for you is that you defined the zdbNamespaces in an outer scope 
- If first ns in the loop was `zdb` then you will fill in the zdbNamespaces
- next iteration on other namespaces, zdbNamespace will still have a value.

Solution, the `var zdbNamespaces should be moved inside the `containers loop` not the namespaces loop to it's ONLY filled for containers of the right namespace.

muhamadazmy: I think now that you wanted to actually check the `len` of the zdbNamespaces. But you still need to make sure to move the variable declaration to the right context.

pkg/provision/cleanup.go
```diff
@@ -0,0 +1,234 @@
...
+					err := common.DeleteZdbContainer(pkg.ContainerID(ctr.ID()), zbus)
+					if err != nil {
+						log.Err(err).Msgf("failed to delete zdb container %s", ctr.ID())
+						continue
+					}
+					toDelete[filepath.Base(mnt.Source)] = struct{}{}
+				} else {
```
muhamadazmy, line 145: I don't see where you mark volumes that belongs to containers that are NOT zdb here. shouldn't these go to the `toSave` map ?

pkg/provision/cleanup.go
```diff
@@ -0,0 +1,234 @@
...
+			}
+
+			qgroup, ok := qgroups[fmt.Sprintf("0/%d", subvol.ID)]
+			if !ok {
+				log.Info().Msgf("skipping volume '%s' has no assigned quota", subvol.Path)
+				continue
+			}
+
+			// Now, is this subvol in one of the toSave ?
+			if _, ok := toSave[filepath.Base(subvol.Path)]; ok {
```
muhamadazmy: Why do we even need the toSave map? isn't it better to only fill a list of the `toDelete` and then just check if the volume is in that map or not ? I mean a volume can either be `to-save` or `to-delete` hence a one map is enough

pkg/provision/engine.go
muhamadazmy: Why didn't you name it `triggerCleanup = true` instead! this is a very weird naming that makes it confusing. and then you have to do the negation in the `if` check. Instead it can be `if reservation.last && triggerCleanup` then u change the value to false.

pkg/provision/source.go
```diff
@@ -30,6 +30,13 @@ type ReservationPoller interface {
 	Poll(nodeID pkg.Identifier, from uint64) (reservations []*Reservation, lastID uint64, err error)
 }
 
+// ReservationJob wraps a reservation type and has
+// a boolean to indicate it is the last reservation
+type ReservationJob struct {
+	Reservation
+	last bool
```
muhamadazmy, line 37: Why this is private.

muhamadazmy (changes requested):

pkg/provision/engine.go
DylanVerstraete: It's the past tense of `trigger`. It means if provisiond HAS triggered cleanup before. We only want to trigger cleanup at boot.

pkg/provision/cleanup.go
DylanVerstraete: Well it cannot panic actually, if the namespace list fails then we continue to the next element. But indeed we can move this variable declaration

pkg/provision/source.go
DylanVerstraete: Why should it be public?

LeeSmet (changes requested):
I don't really like the general structure of the code here. One of the benefits of working with the modules is that we can define some strict responsibliities for every module, which should be handled by that and only that module, and the module should not care about anything else. This PR causes a lot of abstraction loss. Mainly in the provision/cleanup.go file.

Here, for example, we look at all pools by checking `/mnt`. While this is probably correct at this time, it is the sole responsibility of `storaged` to define _where_ it mounts the pools. Even the fact that the pool is mounted can be considered an implementation detail. It would be perfectly fine for `storaged` to mount pools in `/storaged` in the future, which will then break this code. Likewise, this file pulls in dependencies on `containerd` code. The fact that we use containerd is actually an implementation detail of the container module. In other words, we are duplicating a lot of logic here, which should ideally be isolated inside a module.

I think it would be better to split up the cleanup flows to the individual modules, so that every module still maintains full and exclusive ownership of the primitives to matter. This includes searching for stuff to delete.

zaibon:
@LeeSmet while I agree we could move some logic from the cleanup back to the different module. Only provisiond knows if one primitive can be safely deleted or not. So the drive of this logic must remain in provisiond. 
But like you said, we should probably move some of the logic into its own module to be more future proof.

So based on what you said we could create some new zbus method on the different modules:
- storaged: Method to list all filesystems
- the `checkContainers` function can be rewritten by using the contd module. It already exposes a `List` method that does exactly the same

LeeSmet:
I'm fine with provisiond driving the cleanup. After all, thats part of its responsibility. The thing about storaged is just an example I mentioned because it is probably the most obvious one: in the cleanup flow, we go to a hardcoded path, which is decided by storaged, then we check for btrfs subvolumes and destory qgroups. Qgroups are an implementation detail of btrfs, subvols are an implementation detail of btrfs, and btrfs is an implementation detail of storaged. All other code will still work if start using zfs filesystems and instead of subvolumes/qgroups just make exact sized partitions. But this code will definitely break then.

So to me, all daemons should indeed expose some methods which allow you to trigger a cleanup for a resource identified by a reservation. This way, all implementation details are nicely isolated inside the module, and only the primitives are shared among the daemons (and thus part of the interface).

Ultimately, I think this method should aim to reuse the regular decomission flows as much as possible as well. So we add something which allows daemons to verify they have a lingering resource for a given reservation, and if they find one, the regular decomission method is called.

pkg/flist/flist_test.go
```diff
@@ -42,6 +42,12 @@ func (s *StorageMock) ReleaseFilesystem(name string) error {
 	return args.Error(0)
 }
 
+// ReleaseFilesystem releases filesystem mock
```
LeeSmet: Copy pasted comment?

pkg/storage.go
```diff
@@ -112,6 +112,26 @@ type StoragePolicy struct {
...
 }
 
+// Usage struct
+type Usage struct {
+	Size uint64
+	Used uint64
+}
+
+// Filesystem represents a storage space that can be used as a filesystem
+type Filesystem struct {
```
LeeSmet, line 122: Why do we suddenly have this type? Afaik in the past, a volume was identified just by its mountpoint. Why is this suddenly needed

pkg/provision/common/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package common hold logic that is used by both the provision and primitive package
+// it purpose is mainly to avoid circular dependencies
+package common
```
LeeSmet, line 3: seeing the evolution of this PR it feels to me like this package would no longer be needed?

pkg/provision/common/zdb.go
```diff
@@ -0,0 +1,38 @@
...
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/threefoldtech/zbus"
+	"github.com/threefoldtech/zos/pkg"
+	"github.com/threefoldtech/zos/pkg/stubs"
+)
+
+// DeleteZdbContainer removes a 0-DB container and cleanup all related resources (container, flist, network)
+func DeleteZdbContainer(containerID pkg.ContainerID, zbus zbus.Client) error {
```
LeeSmet, line 13: Any particular reason this function was extracted? it feels to me like it was in the proper place to begin with

pkg/provision/cleanup.go
```diff
@@ -0,0 +1,178 @@
...
+
+	fss, err := storaged.ListFilesystems()
+	if err != nil {
+		return err
+	}
+
+	for _, fs := range fss {
+		log.Info().Msgf("checking subvol %s", fs.Path)
+		// Don't delete zos-cache!
+		if fs.Name == storage.CacheLabel {
```
LeeSmet: I don't really know how the cache is treated right now in other modules. But seeing this I wonder if it would not be better to treat it as a special case, and not return it when listing filesystems. Instead if some module needs to have access to the cache (or a subdir of the cache), it might be better to expose a "GetCache" method on storaged imo. The way the cache is handled should also be an implementation detail of storaged (imo).

LeeSmet (changes requested):

pkg/provision/common/doc.go
zaibon: It still is, without it we still circular dependency between the primitives and provision package.

pkg/provision/common/zdb.go
zaibon: >// Package common hold logic that is used by both the provision and primitive package
// it purpose is mainly to avoid circular dependencies
package common

pkg/storage.go
zaibon: Because of the new `ListFilesystems() ([]Filesystem, error)` method

pkg/provision/engine.go
```diff
@@ -71,16 +77,25 @@ func New(opts EngineOps) *Engine {
...
 func (e *Engine) Run(ctx context.Context) error {
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
-
 	cReservation := e.source.Reservations(ctx)
 
+	// run a cron task that will fire the cleanup at midnight
+	cleanUp := make(chan struct{}, 2)
+	c := cron.New()
+	_, err := c.AddFunc("@midnight", func() {
```
LeeSmet, line 92: It feels like there is a race condition here, no?

pkg/storage.go
LeeSmet: Yes, but why does that method return the new type then and not a list of mountpoints (strings)?

pkg/provision/engine.go
zaibon: race condition on what ?

pkg/storage.go
zaibon: Because that would mean you need to do some manual inspection of the volume from wherever this method is called. While now storaged does return you pretty much all the information you would want to know on a volume directly. I find this better abstract the logic of storaged then just returning a list of string.

We could change the type of the other method to use this new type too for consistency though.

pkg/storage/storage.go
```diff
@@ -393,6 +394,42 @@ func (s *storageModule) ReleaseFilesystem(name string) error {
 	return nil
 }
 
+// ListFilesystems return all the filesystem managed by storeaged present on the nodes
+func (s *storageModule) ListFilesystems() ([]pkg.Filesystem, error) {
+	fss := make([]pkg.Filesystem, 0, 10)
+
+	for _, pool := range s.pools {
+		if _, mounted := pool.Mounted(); !mounted {
```
zaibon, line 402: Maybe we should make this configurable by an argument on `ListFilesystems` ?

LeeSmet (approved):
LGTM

zaibon (approved):

PR #897, opened 2020-08-17
Only increment network provision counter when it's created for the first time

### LINKED ISSUE #874: counter for network resource workload should only be increased when a new network is provisioned
https://github.com/threefoldtech/zos/blob/912de085a44abbf3c7a7c84991f7e3fe9e29b21c/pkg/provision/engine.go#L185-L187
Right now any time a workload is provisioned, we always increase the counter. For most workload this is ok. But for the network resource type, it can actually be an update of an existing network and thus the counter should not be increased.

So what I think should be done to solve this:
- make the counter a bit smarter or verify that is indeed a new network before calling `e.statser.Increment(r)`
- Update cache.[Sync function](https://github.com/threefoldtech/zos/blob/912de085a44abbf3c7a7c84991f7e3fe9e29b21c/pkg/provision/primitives/cache/cache.go#L100) to only count the same network once when provisiond starts

DylanVerstraete:


pkg/provision/engine.go
```diff
@@ -182,6 +182,19 @@ func (e *Engine) provision(ctx context.Context, r *Reservation) error {
 		return errors.Wrapf(err, "failed to cache reservation %s locally", r.ID)
 	}
 
+	// If an update occurs on the network we don't increment the counter
```
muhamadazmy, line 187: I don't believe you need this, you can just check the cache `e.cache` if this reservation exists (and still active) or not (before adding it to the cache) then only increment if didn't exist before.

pkg/provision/engine.go
```diff
@@ -182,6 +182,19 @@ func (e *Engine) provision(ctx context.Context, r *Reservation) error {
...
+	if r.Type == "network" {
+		nr := pkg.NetResource{}
+		if err := json.Unmarshal(r.Data, &nr); err != nil {
+			return fmt.Errorf("failed to unmarshal network from reservation: %w", err)
+		}
+
+		exists, _ := e.cache.NetworkExists(nr.Name)
```
muhamadazmy: This will ALWAYS return true. because you do the check already after the provision! no ?

pkg/provision/primitives/cache/cache.go
```diff
@@ -248,6 +249,90 @@ func (s *Fs) Exists(id string) (bool, error) {
...
+
+// incrementCounters will increment counters for all workloads
+// for network workloads it will only increment those that have a unique name
+func (s *Fs) incrementCounters(statser provision.Statser) error {
+	s.RLock()
+	defer s.RUnlock()
+
+	uniqueNetworks := make(map[string]*provision.Reservation)
+
+	infos, err := ioutil.ReadDir(s.root)
```
muhamadazmy, line 280: May be add `List` method to `Fs`.

pkg/provision/primitives/cache/cache.go
```diff
@@ -248,6 +249,90 @@ func (s *Fs) Exists(id string) (bool, error) {
...
+			return fmt.Errorf("failed get reservation: %w", err)
+		}
+
+		if r.Type == primitives.NetworkReservation {
+			nr := pkg.NetResource{}
+			if err := json.Unmarshal(r.Data, &nr); err != nil {
+				return fmt.Errorf("failed to unmarshal network from reservation: %w", err)
+			}
+
+			if _, ok := uniqueNetworks[nr.Name]; ok {
```
muhamadazmy: not true, network name is not unique. the interface/bridges names are made unique by using a mashup of name and user id.

pkg/provision/primitives/cache/cache.go
```diff
@@ -248,6 +249,90 @@ func (s *Fs) Exists(id string) (bool, error) {
...
+				return true, nil
+			}
+		}
+	}
+	return false, nil
+}
+
+// incrementCounters will increment counters for all workloads
+// for network workloads it will only increment those that have a unique name
+func (s *Fs) incrementCounters(statser provision.Statser) error {
```
muhamadazmy: I really think that there is no restriction on the network name (I probably need to refresh my memory on this part) but the network id is based on a combination of network name + user id. Otherwise no 2 users can have the same network name!

muhamadazmy (changes requested):

pkg/provision/primitives/cache/cache.go
```diff
@@ -248,6 +249,89 @@ func (s *Fs) Exists(id string) (bool, error) {
...
+		if info.IsDir() {
+			continue
+		}
+
+		r, err := s.get(info.Name())
+		if err != nil {
+			return nil, fmt.Errorf("failed get reservation: %w", err)
+		}
+
+		reservations = append(reservations, r)
```
muhamadazmy, line 296: This is wrong here, because you already created a LIST with LEN(infos), so append will just add to the end of the list
for example
```go
l := make([]int, 10)
l = append(l, 42)

assert(l[0] == 0)
assert(l[10] == 42)
```

The right way to do it is
```go
reservations := make([]int, 0, cap)
```
so you create a list with underlying capacity (cap) "len(infos) in  your case" but of length 0 (no elements) 
then you do `append` normally

pkg/provision/primitives/cache/cache.go
muhamadazmy: Why did you resolve this? the Name is still not unique. you need to use the `name+id` key instead of just the name

muhamadazmy (changes requested):

pkg/provision/primitives/cache/cache.go
```diff
@@ -248,6 +249,91 @@ func (s *Fs) Exists(id string) (bool, error) {
...
+
+	for _, r := range reservations {
+		if r.Type == primitives.NetworkReservation {
+			nr := pkg.NetResource{}
+			if err := json.Unmarshal(r.Data, &nr); err != nil {
+				return fmt.Errorf("failed to unmarshal network from reservation: %w", err)
+			}
+
+			// if the network name + user exsists in the list, we skip it.
+			// else we add it to the list
```
muhamadazmy, line 320: Hmm, is there a way to get a unique ID from the name/user parts without doing `name + user` because this is not granted to be unique even if the user is unique !
for example a network `test` for user `user` will have the same key as the network `testu` for user `ser` !

Can u please check how the unique bridge name for the network is built ? can we reuse this code ? If it's not "clean" enough to use this code (because it belongs to a different module) you can try something like 
```go
key := hash(network) + hash(user)
```

muhamadazmy: Check how it's done here

https://github.com/threefoldtech/zos/blob/master/pkg/provision/primitives/container.go#L187

muhamadazmy: I think this method already have an issue and should be fixed the way I suggested above

muhamadazmy: May be u should use this method, so once it's fixed it's fixed for all

muhamadazmy (approved):

LeeSmet (approved):

pkg/provision/engine.go
```diff
@@ -182,6 +182,19 @@ func (e *Engine) provision(ctx context.Context, r *Reservation) error {
 		return errors.Wrapf(err, "failed to cache reservation %s locally", r.ID)
 	}
 
+	// If an update occurs on the network we don't increment the counter
+	if r.Type == "network" {
```
zaibon:
```suggestion
	if r.Type == "network_resource" {
```

pkg/provision/engine.go
```diff
@@ -182,6 +182,19 @@ func (e *Engine) provision(ctx context.Context, r *Reservation) error {
...
+		nr := pkg.NetResource{}
+		if err := json.Unmarshal(r.Data, &nr); err != nil {
+			return fmt.Errorf("failed to unmarshal network from reservation: %w", err)
+		}
+
+		exists, _ := e.cache.NetworkExists(nr.Name, r.User)
```
zaibon: any reason we don't process the error ?

zaibon: You also probably want to build the networkID using the function define at https://github.com/threefoldtech/zos/blob/c1291720d92556114e5cb093fcf2eede13a7e2d4/pkg/provision/primitives/network.go#L64
and use this ID in the `NetworkExists` method

pkg/provision/interface.go
```diff
@@ -35,6 +35,7 @@ type ReservationCache interface {
 	Get(id string) (*Reservation, error)
 	Remove(id string) error
 	Exists(id string) (bool, error)
+	NetworkExists(name, user string) (bool, error)
```
zaibon: lets make it consistent with all the other method and just accept `id string`

pkg/provision/primitives/cache/cache.go
```diff
@@ -248,6 +251,93 @@ func (s *Fs) Exists(id string) (bool, error) {
...
+		}
+
+		reservations = append(reservations, r)
+	}
+	return reservations, nil
+}
+
+// incrementCounters will increment counters for all workloads
+// for network workloads it will only increment those that have a unique name
+func (s *Fs) incrementCounters(statser provision.Statser) error {
```
zaibon, line 303: It seems you are only incrementing the network counter in this function. What about all the other primitive types?

zaibon (changes requested):

pkg/provision/engine.go
DylanVerstraete: Not really because this function contains a known issue now. It's better to check if the name and user are equal no?

DylanVerstraete: Using this now. This method also addresses the issue in https://github.com/threefoldtech/zos/issues/898. The reasons it's defined in engine.go is because we get an import cycle if we define it in primitives.

https://github.com/threefoldtech/zos/blob/e729d31d50a0eda533cc359d3f58c97bbe40ee85/pkg/provision/engine.go#L398-L408

pkg/provision/interface.go
DylanVerstraete: see https://github.com/threefoldtech/zos/pull/897#discussion_r477104065

pkg/provision/engine.go
DylanVerstraete: using this now https://github.com/threefoldtech/zos/pull/897#discussion_r474521613

pkg/provision/interface.go
DylanVerstraete: Well we can't really change the input of the method since this method scans on ALL the reservations for a specific network..

pkg/provision/interface.go
```diff
@@ -35,6 +36,7 @@ type ReservationCache interface {
 	Get(id string) (*Reservation, error)
 	Remove(id string) error
 	Exists(id string) (bool, error)
+	NetworkExists(id pkg.NetID) (bool, error)
```
zaibon:
```suggestion
	NetworkExists(id string) (bool, error)
```

zaibon (changes requested):

zaibon (approved):

PR #899, opened 2020-08-19
Register dhcp client of ndmz in zinit

### LINKED ISSUE #876: networkd: register dhcp client of ndmz in zinit
While we get an IP address from the dhcp server for ndmz, once configured, the dhcp client isnt' started again.
After the Lease expiration, that IP is not registered in the DHCP pool any more, and that can create problems.
In se that is not a problem  for most DHCP servers as they verify with an arp request if a potential IP is already taken, but we don't know what dhcp servers are running in the farm.

Better to register the dhcp client of ndmz in zinit, once ndmz is configured.
Also, let's then not forget that when we reconfigure ndmz, we also need to kill the dhcp client first

DylanVerstraete:


pkg/network/dhcp/dhcp.go
```diff
@@ -38,6 +40,45 @@ func (d *Probe) Start(inf string) error {
...
+func (d *Probe) Run(inf string) error {
+	z, err := zinit.New("")
+	if err != nil {
+		log.Error().Err(err).Msg("failed to connect to zinit")
+		return err
+	}
+
+	serviceName := fmt.Sprintf("dhcp-%s", inf)
+
+	ns, err := exec.Command("ip", "netns", "identify").Output()
```
muhamadazmy, line 67: What if you are running on the 'host' namespace which is just an empty string? I think the first thing you need to do after receiving this value `ns` is to trim it, and then make sure it's not an empty string.

Then if it still have a value, you do the `ip netns exec` trick, otherwise you need to just excute the udechpc command directly.

pkg/network/dhcp/dhcp.go
```diff
@@ -38,6 +40,45 @@ func (d *Probe) Start(inf string) error {
...
+	if err != nil {
+		return errors.Wrap(err, "failed to identify namespace")
+	}
+
```
muhamadazmy, line 71: As we agreed, you need to check if another service with the same name already exists! And may be return some error (that can be checked later)

We need also to provide a way to stop that backgroup dhcp process, or check if it exists.

May be it's better if we provide 2 Prope types. The current one can be the default, but a new BackgroundPrope type can also be added that provide 
```go 
NewBackgroundPrope() *BackgroundPrope
Start(interface) error
IsRunning(interface) (bool, error)
// and
Stop(interface) error
```

pkg/network/dhcp/dhcp.go
```diff
@@ -38,6 +40,45 @@ func (d *Probe) Start(inf string) error {
...
+		log.Error().Err(err).Msg("fail to create dhcp-zos zinit service")
+		return err
+	}
+
+	if err := z.Monitor(serviceName); err != nil {
+		log.Error().Err(err).Msg("fail to start monitoring dhcp-zos zinit service")
+		return err
+	}
+
+	if err := z.Start(serviceName); err != nil {
```
muhamadazmy: zinit monitor already auto starts the service. You do Start only if you did a Stop before. You can check zinit docs for more details about the serivce live cycle.

muhamadazmy (changes requested):

pkg/network/ndmz/dualstack.go
```diff
@@ -92,10 +92,16 @@ func (d *DualStack) Create() error {
 // Delete deletes the NDMZ network namespace
 func (d *DualStack) Delete() error {
 	netNS, err := namespace.GetByName(NetNSNDMZ)
-	if err == nil {
-		if err := namespace.Delete(netNS); err != nil {
-			return errors.Wrap(err, "failed to delete ndmz network namespace")
-		}
+	if err != nil {
```
zaibon: Here you risk to return an not exist error while you are trying to delete the ndmz.
That's why the code was put inside the `err == nil` block cause you only need to do something if the namespace actually exists. If it already doesn't exists, this method is a no-op.

pkg/network/ndmz/dualstack.go
```diff
@@ -189,12 +195,24 @@ func (d *DualStack) IP6PublicIface() string {
 // waitIPS waits to receives some IP from DHCP and Router advertisement
 func waitIP4() error {
 	// run DHCP to interface public in ndmz
-	probe := dhcp.NewProbe()
+	probe, err := dhcp.NewBackgroundProbe()
```
zaibon: This is not right. This method should only try to get an IP from DHCP it should not create a service that maintain the DHCP alive. 
Only in the `Create` functions [there](https://github.com/threefoldtech/zos/blob/2f295e39c9b77f57fd6839e68d3c2f1e0f7ad50e/pkg/network/ndmz/dualstack.go#L90) and [there](https://github.com/threefoldtech/zos/blob/2f295e39c9b77f57fd6839e68d3c2f1e0f7ad50e/pkg/network/ndmz/hidden.go#L37)` we need to create the dhcp service only as a last step of the after `waitIP4` and `waitIP6`.

pkg/network/dhcp/dhcp.go
```diff
@@ -38,6 +58,62 @@ func (d *Probe) Start(inf string) error {
 	return nil
 }
 
+// Start runs the DHCP client process and registers it to zinit
+func (d *BackgroundProbe) Start(inf string) error {
```
zaibon: I'm not a big fan of how this is implemented. The tricks to detect in which network namespace we are etc seems error prone.
Why can't we just specify the network namespace if there is any in the NewBackgroundProbe function ? The `inf` argument should also be on the `NewBackgroundProbe` instead of the `Start` method. The idea is you create your object with all it's property and then call `Start`/`Stop` on it without any further argument.

zaibon (changes requested):

pkg/network/ndmz/dualstack.go
muhamadazmy: So the wait is only to test if you can get an IP on that interface?

pkg/network/dhcp/dhcp.go
muhamadazmy: - I think the reason he passes the inf as param to Start is to be aligned with the `Probe` type.
- I suggested the code to detect the active namespace so it can be used seemlesy from the `ns.Do()` function and it will configure the zinit unit file to use the correct namespace (or not), again to be aligned with the `Probe` type. Then we can have a Probe interface at some point if needed.

Of course this also can be introduced as a completely new code not related to the Probe functionality and hence have its own API

pkg/network/ndmz/dualstack.go
zaibon: Actually I just checked the code that choose the network mode (dualstack, hidden) of the node and I was wrong. We do this check already before creating the ndmz: https://github.com/threefoldtech/zos/blob/d693d0656a8cc2bd7e72f10f1ef9d6082e43cec5/cmds/networkd/main.go#L315

So I this code is alright, we can indeed start the DHCP service directly here.
I would double check the argument we pass to udhcpc though cause they were configure to test if we get an IP and not to run as a daemon.

pkg/network/dhcp/dhcp.go
zaibon: yep I just saw the Probe type was also using this pattern. Which I still think is wrong (it's my own code, beurk).

I see that we indeed already use the `Probe` type sometimes into different network namespace (ns.Do). So if we want to have these 2 types have the same interface, we can keep the auto detection of the current namespace. But I think you can do this directly instead of using `ip` command.

So to recap I would do somehting like:
```go
type Probe struct{
    Interface string
}

func NewProbe(inf string) *Probe {...}
func (d *Probe) Start() error { ... }
func (d *Probe) Stop() error { ... }
```

And use the same interface for `BackgroundProbe` type but use zinit to implement.

pkg/network/ndmz/dualstack.go
```diff
@@ -187,35 +195,47 @@ func (d *DualStack) IP6PublicIface() string {
 }
 
 // waitIPS waits to receives some IP from DHCP and Router advertisement
-func waitIP4() error {
+func waitIP4() (*dhcp.BackgroundProbe, error) {
```
muhamadazmy: Why does this has to return the Probe now ?

pkg/network/ndmz/dualstack.go
```diff
@@ -25,8 +25,9 @@ import (
 
 // DualStack implement DMZ interface using dual stack ipv4/ipv6
 type DualStack struct {
-	nodeID     string
-	ipv6Master string
+	nodeID          string
+	ipv6Master      string
+	backgroundProbe *dhcp.BackgroundProbe
```
muhamadazmy: I don't think you need to store this here, because it's a stateless object. it means you still can create another one with the same "interface" name and use it any time.

pkg/network/dhcp/dhcp.go
```diff
@@ -38,6 +60,61 @@ func (d *Probe) Start(inf string) error {
...
+	return nil
+}
+
+// IsRunning checks if a background process is running in zinit
+func (d *BackgroundProbe) IsRunning() (bool, error) {
+	serviceName := fmt.Sprintf("dhcp-%s", d.inf)
+
+	services, err := d.z.List()
+	if err != nil {
+		log.Error().Err(err).Msg("fail to create dhcp-zos zinit service")
```
muhamadazmy: Wrong log message! Also no point of logging here, just make sure to wrap your err with the correct message.

pkg/network/ndmz/dualstack.go
```diff
@@ -267,3 +284,21 @@ func waitIP6() error {
...
+
+func stopBackgroundProbe() error {
+	probe, err := dhcp.NewBackgroundProbe(DMZPub4)
+	if err != nil {
+		return err
+	}
+
+	running, err := probe.IsRunning()
+	if err != nil {
+		return nil
```
muhamadazmy: why ignoring the error? the error does not mean the service is running, no? This need to bubble up the stack

pkg/network/ndmz/dualstack.go
```diff
@@ -267,3 +284,21 @@ func waitIP6() error {
...
+	}
+
+	// this means this process is already running, stop here
+	if running {
```
muhamadazmy: This method is to 'stop' background probe, so the check should be `if !running` not `if running` !

pkg/network/ndmz/hidden.go
```diff
@@ -65,14 +65,21 @@ func (d *Hidden) Create() error {
 			return errors.Wrapf(err, "failed to enable forwarding in ndmz")
 		}
 
-		return waitIP4()
+		if err := waitIP4(); err != nil {
```
muhamadazmy: Why this change is needed, this has exactly the same side effect as before

pkg/network/dhcp/dhcp.go
```diff
@@ -38,6 +60,61 @@ func (d *Probe) Start(inf string) error {
...
+	serviceName := fmt.Sprintf("dhcp-%s", d.inf)
+
+	ns, err := exec.Command("ip", "netns", "identify").Output()
+	if err != nil {
+		return errors.Wrap(err, "failed to identify namespace")
+	}
+
+	exec := fmt.Sprintf("/sbin/udhcpc -v -f -i %s -t 20 -T 1 -s /usr/share/udhcp/simple.script", d.inf)
+
+	if strings.Trim(string(ns), " ") != "" {
```
muhamadazmy: Use `strings.TrimSpaces()` instead. it will also cover utf-8 spaces, tabs and new lines

muhamadazmy: it's actually better if you do `ns = strings.TrimSpaces(ns)` and then use the cleaned up ns because you still using `ns` in code below

muhamadazmy (changes requested):

zaibon (approved):

PR #911, opened 2020-08-27
fix broken counters provisiond, closes #874

### LINKED ISSUE #874: counter for network resource workload should only be increased when a new network is provisioned

DylanVerstraete:
We were calling `Sync` twice apparently.. Also we looped over reservervations whilst incrementing counter, this incrementing method would also loop over reservations..

### DIFF
--- a/pkg/provision/engine.go
+++ b/pkg/provision/engine.go
@@ -76,10 +76,6 @@ func New(opts EngineOps) *Engine {
 
 // Run starts reader reservation from the Source and handle them
 func (e *Engine) Run(ctx context.Context) error {
-	if err := e.cache.Sync(e.statser); err != nil {
-		return fmt.Errorf("failed to synchronize statser: %w", err)
-	}
-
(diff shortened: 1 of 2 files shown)


zaibon (changes requested):
Nice catch ! :+1:

zaibon (approved):

PR #917, opened 2020-09-02
Handle zdb namespace and macvlan cleanly 

### LINKED ISSUE #916: zdb: deploy zdb gives error 'address already in use'
While deploying the workload https://explorer.testnet.grid.tf/api/v1/reservations/workloads/472974 it failed with 

`failed to ensure zdb containe running: failed to prepare zdb network: address already in use`

`~ # ip netns|grep zdb

zdb-ns-201a81e7
zdb-ns-ee1fd86b
zdb-ns-119049e0
zdb-ns-652bde38
zdb-ns-1f1e6307
zdb-ns-5801d929
zdb-ns-fc7ff3d6
zdb-ns-c96a5865
zdb-ns-19520456
zdb-ns-7a7ea7d4
`

Question: Why we create random namespaces names if we have a predicted hw address?

Suggestion: Since the hw address is computed from the volume path, we can create a predicted namespace, and then be sure that only one namespace is created by zdb. Then check that the macvlan is already create, or create a new one. 

> I can't really find out why we recieve the 'address already in use' error

muhamadazmy:
This is to avoid creating multiple namespaces for the same
zdb deployment, which can result in address already in use error

Fixes #916

### DIFF
--- a/pkg/network/namespace/namespace.go
+++ b/pkg/network/namespace/namespace.go
@@ -159,7 +159,7 @@ func Exists(name string) bool {
 
 	if !mounted {
 		//the file shouldn't be there
-		os.Remove(nsPath)
+		_ = os.Remove(nsPath)
 	}
 
 	return mounted
--- a/pkg/network/namespace/namespace.go
+++ b/pkg/network/namespace/namespace.go
@@ -1,6 +1,7 @@
 package namespace
 
 import (
+	"bufio"
 	"fmt"
 	"os"
 	"path"
@@ -11,6 +12,7 @@ import (
 
 	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/pkg/errors"
+	"github.com/rs/zerolog/log"
 	"golang.org/x/sys/unix"
 )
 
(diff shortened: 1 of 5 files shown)


DylanVerstraete (approved):

pkg/network/namespace/namespace.go
```diff
@@ -146,11 +148,55 @@ func Delete(ns ns.NetNS) error {
...
+	}
+	mounted, err := isNamespaceMounted(name)
+	if err != nil {
+		log.Error().Err(err).Msg("failed to check if namespace is mounted")
+		return false
+	}
+
+	if !mounted {
+		//the file shouldn't be there
+		os.Remove(nsPath)
```
zaibon: make it explicit you don't want to handle the error
```suggestion
		_ = os.Remove(nsPath)
```

pkg/network/networker.go
```diff
@@ -266,10 +267,10 @@ func (n *networker) Leave(networkdID pkg.NetID, containerID string) error {
 // ZDBPrepare sends a macvlan interface into the
 // network namespace of a ZDB container
 func (n networker) ZDBPrepare(hw net.HardwareAddr) (string, error) {
-	netNSName, err := ifaceutil.RandomName("zdb-ns-")
-	if err != nil {
-		return "", err
-	}
+	netNSName := fmt.Sprintf(
```
zaibon, line 270: does this naming convention always ensure the namespace name will be smaller then 16 characters ?

zaibon (changes requested):

pkg/network/networker.go
muhamadazmy: No, but we can drop few characters from the prefix I think `zdb-` is enough, no. Also I don't believe there is a limitation on the name of the namespace. This works perfectly well on my dev node.

zaibon: @delandtj aren't network namespace name also limited in length like network interface are ?

zaibon (approved):

PR #925, opened 2020-09-08
WIP: clean up network on container destroy

muhamadazmy:


pkg/network/networker.go
```diff
@@ -317,6 +317,27 @@ func (n networker) ZDBPrepare(hw net.HardwareAddr) (string, error) {
 	return netNSName, n.createMacVlan(ZDBIface, hw, ips, routes, netNs)
 }
 
+// ZDBDestroy is the opposite of ZDPrepare, it makes sure network setup done
+// for zdb is rewind. ns param is the namespace return by the ZDBPrepare
+func (n networker) ZDBDestroy(ns string) error {
+	if !strings.HasPrefix(ns, "zdb-ns-") {
```
zaibon: Let's make this prefix a constant so we can reuse it in both `ZDBPrepare` and `ZDBDestroy` methods

pkg/provision/primitives/zdb.go
```diff
@@ -421,28 +421,24 @@ func (p *Provisioner) deleteZdbContainer(containerID pkg.ContainerID) error {
...
 
-	} else {
-		log.Error().Err(err).Str("container", string(containerID)).Msg("failed to inspect container for decomission")
+	network := stubs.NewNetworkerStub(p.zbus)
+	if err := network.ZDBDestroy(info.Network.Namespace); err != nil {
+		return errors.Wrapf(err, "failed to destroy zdb network namespace")
 	}
 
-	// TODO: delete network?
+	if err := flist.Umount(info.RootFS); err != nil {
```
zaibon, line 439: you are missing this code:
```go
rootFS := info.RootFS
if info.Interactive {
	rootFS, err = findRootFS(info.Mounts)
	if err != nil {
		return err
	}
}
```
that is needed otherwise the `Umount` could fail in case the container was interactive mode.

zaibon (changes requested):

pkg/provision/primitives/zdb.go
muhamadazmy: Yeah, i removed it because we will never run zdb container in interactive mode, no ? the delete zdb container code was initially copied from the more generic container decomission

zaibon: Right

zaibon (approved):

PR #931, opened 2020-09-10
fix disk shutdown after decomission filesystem

### LINKED ISSUE #930: Disk shutdown after decomission of last volume turns back on after some time
So when we decomission a filesystem and it was the only one remaining on that disk we shutdown the disk. Turns out that after a few minutes this disk got turned on again. This is most likely btrfs doing some IO under the hood.. 

Proposed fix: run periodically check and see which disks are supossed to be OFF and they seem ON, turns these OFF again

DylanVerstraete:


pkg/storage/storage.go
```diff
@@ -295,6 +296,8 @@ func (s *storageModule) initialize(policy pkg.StoragePolicy) error {
 		log.Error().Err(err).Msg("Error shutting down unused pools")
 	}
 
+	s.verifyShutdown()
```
muhamadazmy: this function name is missleading, it sounds like it does a certain function and exits, while in the matter of fact it monitor the disks.

pkg/storage/storage.go
```diff
@@ -668,3 +671,65 @@ func (s *storageModule) Monitor(ctx context.Context) <-chan pkg.PoolsStats {
 
 	return ch
 }
+
+func (s *storageModule) verifyShutdown() {
```
muhamadazmy: The entire function needs re factoring to avoid this deep indentation it makes it very hard to read.

pkg/storage/storage.go
```diff
@@ -668,3 +671,65 @@ func (s *storageModule) Monitor(ctx context.Context) <-chan pkg.PoolsStats {
...
+				}
+			case <-quit:
+				ticker.Stop()
+				return
+			}
+		}
+	}()
+}
+
+func checkDiskPowerStatus(path string) (bool, error) {
```
muhamadazmy, line 722: I am not fully sure but I think there is a smarctl utility library in zos already, this might be not true

muhamadazmy (changes requested):

zaibon (changes requested):
I would like we gather a bit more data about how often and maybe why a disk turn on automatically.
My fear is that with this PR, we could potentially having disk being turn on and off every 5 minutes and I don't think this is that good for the lifetime of the disk.

DylanVerstraete:
You can now ssh to a machine and do `wget -O - http://localhost:8081/debug/vars | grep /dev` which will output the number of shutdowns for each disk.

pkg/storage/storage.go
DylanVerstraete: There is but we don't need this logic anywhere else.

pkg/storage/storage.go
```diff
@@ -668,3 +679,71 @@ func (s *storageModule) Monitor(ctx context.Context) <-chan pkg.PoolsStats {
...
+			log.Debug().Msgf("checking device: %s", device.Path)
+			on, err := checkDiskPowerStatus(device.Path)
+			if err != nil {
+				if err, ok := err.(*exec.ExitError); ok {
+					log.Err(err).Msgf("skipping device")
+					// if cmd exits with exit error the device is shutdown
+					continue
+				}
+			}
+
```
muhamadazmy, line 706: you should instead do 
```go
if !on {
  continue
} 
```

pkg/storage/storage.go
```diff
@@ -668,3 +679,71 @@ func (s *storageModule) Monitor(ctx context.Context) <-chan pkg.PoolsStats {
...
+			_, mounted := pool.Mounted()
+			if mounted {
+				continue
+			}
+
+			if on {
```
muhamadazmy: Then you don't need to do this check anymore since you won't be here if disk is off

pkg/storage/storage.go
```diff
@@ -668,3 +679,71 @@ func (s *storageModule) Monitor(ctx context.Context) <-chan pkg.PoolsStats {
...
+	}()
+}
+
+// shutdownDisks will check the disks power status.
+// If a disk is on and it is not mounted then it is not supposed to be on, turn it off
+func (s *storageModule) shutdownDisks() {
+	for _, pool := range s.pools {
+		for _, device := range pool.Devices() {
+			log.Debug().Msgf("checking device: %s", device.Path)
+			on, err := checkDiskPowerStatus(device.Path)
```
muhamadazmy, line 701: Why should a caller of this method know how it's implemented (with exec) ? A caller wouldn't know what kind of underlying error is returned. hence this method should check the error and then return defined types here.

pkg/storage/storage.go
```diff
@@ -668,3 +679,71 @@ func (s *storageModule) Monitor(ctx context.Context) <-chan pkg.PoolsStats {
...
+					return
+				}
+			}
+		}
+	}
+}
+
+func checkDiskPowerStatus(path string) (bool, error) {
+	output, err := exec.Command("smartctl", "-i", "-n", "standby", path).Output()
+	if err != nil {
```
muhamadazmy, line 724: you should instead do the check here and just return the proper "bool"/error flag
if the cmd exits with non success error if the device is down, you should do 
```
if (err is exit error) {
   return false, nil
} else {
   return false, err
}
```
for example

pkg/storage/storage.go
```diff
@@ -668,3 +679,71 @@ func (s *storageModule) Monitor(ctx context.Context) <-chan pkg.PoolsStats {
...
+	for _, pool := range s.pools {
+		for _, device := range pool.Devices() {
+			log.Debug().Msgf("checking device: %s", device.Path)
+			on, err := checkDiskPowerStatus(device.Path)
+			if err != nil {
+				if err, ok := err.(*exec.ExitError); ok {
+					log.Err(err).Msgf("skipping device")
+					// if cmd exits with exit error the device is shutdown
+					continue
+				}
```
muhamadazmy: you also never return here if you have a different error :/

cmds/storaged/main.go
```diff
@@ -65,6 +66,12 @@ func main() {
 		log.Info().Msg("shutting down")
 	})
 
+	go func() {
+		if err := http.ListenAndServe(":8080", http.DefaultServeMux); err != nil {
```
muhamadazmy: What is this?

muhamadazmy (changes requested):

cmds/storaged/main.go
DylanVerstraete: https://github.com/threefoldtech/zos/pull/931#issuecomment-691037449

muhamadazmy: i don't like this port is hard-coded specially a very "common" port like 8080

DylanVerstraete: changed it to 8081

cmds/storaged/main.go
```diff
@@ -65,6 +66,12 @@ func main() {
 		log.Info().Msg("shutting down")
 	})
 
+	go func() {
+		if err := http.ListenAndServe(":8081", http.DefaultServeMux); err != nil {
```
muhamadazmy: Are you serious ? :'D please make this configurable via the command line, and make the default something with 5 figures. try 28682 (it's my birthday)

pkg/storage/storage.go
```diff
@@ -668,3 +679,70 @@ func (s *storageModule) Monitor(ctx context.Context) <-chan pkg.PoolsStats {
...
+			}
+		}
+	}
+}
+
+func checkDiskPowerStatus(path string) (bool, error) {
+	output, err := exec.Command("smartctl", "-i", "-n", "standby", path).Output()
+	if err != nil {
+		if err, ok := err.(*exec.ExitError); ok {
+			log.Err(err).Msgf("skipping device")
```
muhamadazmy: This log doesn't make any sense here

pkg/storage/storage.go
```diff
@@ -668,3 +679,70 @@ func (s *storageModule) Monitor(ctx context.Context) <-chan pkg.PoolsStats {
...
+// shutdownDisks will check the disks power status.
+// If a disk is on and it is not mounted then it is not supposed to be on, turn it off
+func (s *storageModule) shutdownDisks() {
+	for _, pool := range s.pools {
+		for _, device := range pool.Devices() {
+			log.Debug().Msgf("checking device: %s", device.Path)
+			on, err := checkDiskPowerStatus(device.Path)
+			if err != nil {
+				log.Err(err).Msgf("error occurred while checking disk power status")
+				return
```
muhamadazmy: You aware once you return here, there is no way this routine will do it's job ever again? (unless you restart storaged) instead please log an error and try the following disk. return will exit the go routine

pkg/storage/storage.go
```diff
@@ -668,3 +679,70 @@ func (s *storageModule) Monitor(ctx context.Context) <-chan pkg.PoolsStats {
...
+			_, mounted := pool.Mounted()
+			if mounted || !on {
+				continue
+			}
+
+			log.Debug().Msgf("shutting down device %s because it is not mounted and the device is on", device.Path)
+			err = pool.Shutdown()
+			if err != nil {
+				log.Err(err).Msgf("failed to shutdown device %s", device.Path)
+				return
```
muhamadazmy: Same comment, if i can't handle one disk it doesn't mean i never try again (and what about the other ones i can shutdown?)

pkg/storage/storage.go
```diff
@@ -668,3 +679,70 @@ func (s *storageModule) Monitor(ctx context.Context) <-chan pkg.PoolsStats {
 
 	return ch
 }
+
+func (s *storageModule) periodicallyCheckDiskShutdown() {
+	ticker := time.NewTicker(5 * time.Minute)
+	quit := make(chan struct{})
```
muhamadazmy: You aware this channel is never getting any input? the right way to do it is to pass a `Context` object to this function, then the caller can call the cancel method.

muhamadazmy: If you don't need a context (or don't want one) you can simply drop the select stmt and instead have a loop like

```go
for {
 <- ticker.C
 s.shutdownDisks()
}
```

pkg/storage/storage.go
DylanVerstraete: CI fails if you don't use that error.. Also it does not like it if you replace it with _

zaibon (approved):

PR #935, opened 2020-09-14
force make filesystem on a device

DylanVerstraete:
We add the -f because some disks with an existing partition table that we know are free gave issues when trying to create the brfs filesystem.

### DIFF
--- a/pkg/storage/filesystem/btrfs.go
+++ b/pkg/storage/filesystem/btrfs.go
@@ -53,7 +53,7 @@ func NewBtrfs(manager DeviceManager) Filesystem {
 	return newBtrfs(manager, executerFunc(run))
 }
 
-func (b *btrfs) Create(ctx context.Context, name string, policy pkg.RaidProfile, devices ...*Device) (Pool, error) {
+func (b *btrfs) Create(ctx context.Context, name string, policy pkg.RaidProfile, force bool, devices ...*Device) (Pool, error) {
 	name = strings.TrimSpace(name)
 	if len(name) == 0 {
 		return nil, fmt.Errorf("invalid name")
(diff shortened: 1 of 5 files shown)


delandtj:
how do we __know__ they are free ?

DylanVerstraete:
@delandtj  If we reach this point in the code we know a disk is free because we check it before doing this.
https://github.com/threefoldtech/zos/blob/294a0b6a64338a63215c80572d9cf232614c536d/pkg/storage/storage.go#L201

pkg/storage/filesystem/btrfs.go
```diff
@@ -81,6 +81,7 @@ func (b *btrfs) Create(ctx context.Context, name string, policy pkg.RaidProfile,
 		"-L", name,
 		"-d", string(policy),
 		"-m", string(policy),
+		"-f",
```
zaibon: I would really want to not have this flag so deep here by default.
This change the behavior of all the call to `btrfs.Create` while we only need this flag during intialization when we claim the free disks.

Instead I would create a new method CreateForce or something, so its very explicit that the method is destructive. And we use this method only in the initialization code.

zaibon (changes requested):

pkg/storage/filesystem/btrfs.go
DylanVerstraete: :+1:

pkg/storage/filesystem/btrfs.go
```diff
@@ -97,6 +97,51 @@ func (b *btrfs) Create(ctx context.Context, name string, policy pkg.RaidProfile,
 	return newBtrfsPool(name, devices, &b.utils), nil
 }
 
+func (b *btrfs) CreateForce(ctx context.Context, name string, policy pkg.RaidProfile, devices ...*Device) (Pool, error) {
```
zaibon, line 60: Can we also refactor the code a bit to avoid having to copy the full function like this ?

zaibon (changes requested):

pkg/storage/filesystem/btrfs.go
zaibon: Something like:
```go
 func (b *btrfs) CreateForce(ctx context.Context, name string, policy pkg.RaidProfile, devices ...*Device) (Pool, error) { 
    return b.create(...,true)
}

 func (b *btrfs) Create(ctx context.Context, name string, policy pkg.RaidProfile, devices ...*Device) (Pool, error) { 
   return b.create(...,false)
}

 func (b *btrfs) create(ctx context.Context, name string, policy pkg.RaidProfile, devices ...*Device, force bool) (Pool, error) { 
...
}
```

zaibon (approved):

Issue #922, opened 2020-09-07
provisiond: try to provision the same workload twice in 1 second


zaibon:
```
[+] provisiond: 2020-09-07T10:24:29Z info start provisioning reservation duration=2562047h47m16.854775807s expired=false id=475694-1 tag= to-delete=false type=zdb
[+] provisiond: 2020-09-07T10:24:30Z error failed to apply provision error="failed to ensure zdb containe running: failed to prepare zdb network: address already in use" id=475694-1
[+] provisiond: 2020-09-07T10:24:30Z error failed to provision reservation 475694-1 error="failed to ensure zdb containe running: failed to prepare zdb network: address already in use"
[+] provisiond: 2020-09-07T10:24:30Z info start provisioning reservation duration=2562047h47m16.854775807s expired=false id=475694-1 tag= to-delete=false type=zdb
[+] provisiond: 2020-09-07T10:24:30Z error failed to apply provision error="failed to ensure zdb containe running: failed to prepare zdb network: address already in use" id=475694-1
[+] provisiond: 2020-09-07T10:24:30Z error failed to provision reservation 475694-1 error="failed to ensure zdb containe running: failed to prepare zdb network: address already in use"
```

zaibon:
More occurrence of the problem: https://github.com/threefoldtech/zos/issues/910#issuecomment-701332924

DylanVerstraete:
@zaibon issue still relevant?

zaibon:
yes, I've not seen any code going in to fix this.

DylanVerstraete:
Cannot reproduce though, also seems like #940 should have fixed it no?

zaibon:
I don't think so no. #940 only make it possible to re-send the same result for an already provisioned workload. Which maybe is enough but for sure doesn't change the fact that the same workload can be received in a very short period

muhamadazmy:
I will just state some facts here that I am sure you are fully aware of, but just to try logically resolve this:
- The provision engine is completely sequential. basically means that it process single reservation at a time
- Results of a reservation is cached, so next time the same reservation is received, the result will be returned immediately 
- If the same reservation is queued 2 (or more) times immediately after each other, only the 1st one is applied, but it **should** never cause a race condition due to the sequential nature of the engine.
- If the same reservation was queued 2 (or more) times it's probably caused by the explorer, hence we might be looking in the wrong place.
[...]

PR #940, opened 2020-09-14
Cache results and resend them when needed

muhamadazmy:
We trying to improve how the node handle double send of reservations better by making sure that all reservations results are cached and send back if needed.

Related to #922


muhamadazmy:
Testing in progress

zaibon (approved):

PR #946, opened 2020-09-16
networkd: disable ra on ygg interface

### LINKED ISSUE #941: netoworkd: ygg interface in container should not receive IP from SLAAC
The goal is to integrate Yggdrasil into 0-OS.
Use cases are:
- Use Yggdrasil as gobal routing network so IPv4 only nodes can still be part of the Grid
- 0-DB are always accessible over Yggdrasil (so ipv4 only nodes can still host 0-DB)
- User can ask to have an extra network interface in their container or k8s VM that has an Yggdrasil IP

Todo:
- [ ] Specification for networkd internal architecture (How and Where does yggdrasil daemon must run)
- [ ] Modification of networkd
- [ ] Testing
  - [ ] ZDB reachable over Yggdrasil
  - [ ] Container can be accessed over Yggdrasil
  - [ ] K8S VMs can be accessed over Yggdrasil

zaibon:
fixes #941

### DIFF
--- a/pkg/network/networker.go
+++ b/pkg/network/networker.go
@@ -21,6 +21,7 @@ import (
 	"github.com/threefoldtech/zos/pkg/network/yggdrasil"
 
 	"github.com/containernetworking/plugins/pkg/ns"
+	"github.com/containernetworking/plugins/pkg/utils/sysctl"
 	"github.com/vishvananda/netlink"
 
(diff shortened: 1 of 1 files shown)


delandtj (commented):
ait

muhamadazmy (approved):

PR #968, opened 2020-09-23
Include runc on containerd flist package

maxux:
Latest `initramfs` (autobuild branch) doesn't include `runc` anymore, since it's a runtime dependency linked to `containerd`. It was a mistake to keep it there and this is fixed now.

This lead to missing `runc` binary on new image and thus crash container deployment. This fix the `containerd.flist` with `runc` included.

This was tested on local VM on devnet and works out of box without local overlay :)

### DIFF
--- a/bins/packages/runc/runc.sh
+++ b/bins/packages/runc/runc.sh
@@ -1,0 +1,73 @@
+RUNC_VERSION="1.0.0-rc9"
+RUNC_CHECKSUM="e88bcb1a33e7ff0bfea495f7263826c2"
+RUNC_LINK="https://github.com/opencontainers/runc/archive/v${RUNC_VERSION}.tar.gz"
+
+dependencies_runc() {
+    apt-get install -y btrfs-tools libseccomp-dev build-essential pkg-config
+
+    if [ -z $GOPATH ]; then
+        if command -v go > /dev/null; then
+            export GOPATH=$(go env GOPATH)
+        else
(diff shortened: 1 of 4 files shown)


zaibon (changes requested):
You also need to update the workload for testing and production. Otherwise looks good to me.

bins/packages/runc/runc.sh
```diff
@@ -0,0 +1,73 @@
...
+RUNC_LINK="https://github.com/opencontainers/runc/archive/v${RUNC_VERSION}.tar.gz"
+
+dependencies_runc() {
+    apt-get install -y btrfs-tools libseccomp-dev build-essential pkg-config
+
+    if [ -z $GOPATH ]; then
+        if command -v go > /dev/null; then
+            export GOPATH=$(go env GOPATH)
+        else
+            curl -L https://dl.google.com/go/go1.13.1.linux-amd64.tar.gz > /tmp/go1.13.1.linux-amd64.tar.gz
```
zaibon, line 12: Can we use the latest version of go instead?

zaibon (changes requested):

bins/packages/runc/runc.sh
maxux: This code is not used when running on github action, we install go (1.15) first, then execute this.
This part of code is executed only when doing standalone and is mainly for debug purpose. We can update it, sure, but it doesn't have any impact on resulting flist.

zaibon: Got it.

zaibon (approved):
LGTM

PR #978, opened 2020-09-25
Rename StatsAggregator field to Stats on Container

maxux:
This update is related to https://github.com/threefoldtech/tfexplorer/pull/174

This update internal usage of `StatsAggregator` for container.
A dummy `StatsAggregator` struct still exists for the other types which needs it (but don't use it really).
A new `Stats` for containers is added next to that and is dedicated to `Container`

This prevent a misunderstanding (`StatsAggregator` field not using `StatsAggregator` type)

This update also points to a `tfexplorer` version already supporting the new schema.


zaibon (approved):
See the comment on the linked PR in tfexplroer: https://github.com/threefoldtech/tfexplorer/pull/174#discussion_r494975219 but other then that LGTM

zaibon (approved):
LGTM

PR #991, opened 2020-10-07
rate limit the logs reservation poll logs

zaibon:
the goal is to avoid having provisiond logs spamming the circular buffer
of zinit and push away usefull logs.
We allow to pint these logs once every hour, just as an easy way to make
sure provisiond is not sutck.

### DIFF
--- a/pkg/flist/cleanup.go
+++ b/pkg/flist/cleanup.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog/log"
+	"github.com/threefoldtech/zos/pkg/app"
 )
 
 // Cleaner interface, implementer of this interface
@@ -127,6 +128,8 @@ func (f *flistModule) cleanupAll() error {
 // Cleaner runs forever, checks the tracker files for filesystem processes
 // that requires cleanup
 func (f *flistModule) Cleaner(ctx context.Context, every time.Duration) {
+	log := app.SampledLogger()
+
 	for {
 		select {
 		case <-ctx.Done():
--- a/pkg/provision/source.go
+++ b/pkg/provision/source.go
@@ -11,6 +11,7 @@ import (
 	"github.com/rs/zerolog/log"
 	"github.com/threefoldtech/tfexplorer/client"
 	"github.com/threefoldtech/zos/pkg"
+	"github.com/threefoldtech/zos/pkg/app"
 )
 
 var (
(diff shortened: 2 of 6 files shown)


pkg/flist/cleanup.go
```diff
@@ -127,11 +128,14 @@ func (f *flistModule) cleanupAll() error {
 // Cleaner runs forever, checks the tracker files for filesystem processes
 // that requires cleanup
 func (f *flistModule) Cleaner(ctx context.Context, every time.Duration) {
+	loglimiter := rate.NewLimiter(rate.Every(time.Hour), 1)
```
muhamadazmy: Can this be defined globally instead, with the configured rate, but with a burst of like 10, so each hour we get max of 10 logs ?

muhamadazmy (commented):
Just one question

muhamadazmy:
Getting to think about it, can't we have some sort of special logger (may be a wrapper around the current one) that just counts how much a unique message was printed, and then just print a summary every 1 hour. Higher level logs can get printed directly (like errors)

pkg/flist/cleanup.go
zaibon: We could indeed adapt the code so I don't have to do these initialization every time. Maybe a customer logger.
But a burst of 10 means we will have 10 logs the 100 first second, then nothing for the rest of the hour, not sure this is really better.

zaibon:
turns out zerolog library already have some utility to do exactly what I'm looking for, I've refactor to use this feature instead of the manual rate limiting.

pkg/flist/cleanup.go
```diff
@@ -127,13 +128,18 @@ func (f *flistModule) cleanupAll() error {
 // Cleaner runs forever, checks the tracker files for filesystem processes
 // that requires cleanup
 func (f *flistModule) Cleaner(ctx context.Context, every time.Duration) {
+	slog := log.Sample(zerolog.LevelSampler{InfoSampler: &zerolog.BurstSampler{
```
muhamadazmy: I don't think this one is needed because the slog only sample info, but the only log in this method that uses it is already of type Error.

pkg/provision/source.go
```diff
@@ -20,6 +21,13 @@ var (
 	ErrPollEOS = fmt.Errorf("end of stream")
 )
 
+var (
+	logSampler = &zerolog.BurstSampler{
```
muhamadazmy: This can be the entire LogSampler instead of only the BurstSampler.

zaibon:
@muhamadazmy I ended up exposing a new method in the app package to get a pre-configured sampled logger. for the sampling to work properly each logger needs to have its own BurstSampler object.

muhamadazmy (approved):

PR #992, opened 2020-10-07
Add cache fs size to counters

### LINKED ISSUE #988: the storage reserved by zos itself for its cache is not taking in account in the resource reported to the explorer
When the node report it's total and used capacity to the explorer, the capacity reserved by 0-OS itself for its cache is no taken in account. This lead to the node reporting more storage available then it can actually use which then leads to problem during capacity planning.
We should take the cache size in account when reporting total capacity to the explorer.

DylanVerstraete:
We can display in the explorer UI that there is 100 GB reserverd for zos cache.

### DIFF
--- a/pkg/storage.go
+++ b/pkg/storage.go
@@ -130,6 +130,8 @@ type Filesystem struct {
 	Name string
 	// FsType of the Filesystem
 	FsType string
+	// DiskType of the Filesystem
+	DiskType string
 }
 
 // VolumeAllocater is the zbus interface of the storage module responsible
--- a/pkg/storage.go
+++ b/pkg/storage.go
@@ -131,7 +131,7 @@ type Filesystem struct {
 	// FsType of the Filesystem
 	FsType string
 	// DiskType of the Filesystem
-	DiskType string
+	DiskType DeviceType
 }
 
 // VolumeAllocater is the zbus interface of the storage module responsible
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -444,7 +444,7 @@ func (s *storageModule) ListFilesystems() ([]pkg.Filesystem, error) {
 					Size: usage.Size,
 					Used: usage.Used,
 				},
-				DiskType: string(pool.Type()),
+				DiskType: pool.Type(),
 			})
 		}
 	}
@@ -476,7 +476,7 @@ func (s *storageModule) Path(name string) (pkg.Filesystem, error) {
 						Size: usage.Size,
 						Used: usage.Used,
 					},
-					DiskType: string(pool.Type()),
+					DiskType: pool.Type(),
 				}, nil
 			}
 		}
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -423,7 +423,6 @@ func (s *storageModule) ListFilesystems() ([]pkg.Filesystem, error) {
 			return nil, err
 		}
 		for _, v := range volumes {
-
 			// Do not return "special" volumes here
 			// instead the GetCacheFS and GetVdiskFS to access them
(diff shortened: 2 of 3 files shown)


zaibon (approved):

zaibon:
LGTM.
@DylanVerstraete can you also open a PR for the UI work on the explorer please

DylanVerstraete:
@zaibon https://github.com/threefoldtech/tfexplorer/pull/193

maxux (approved):

pkg/storage.go
```diff
@@ -130,6 +130,8 @@ type Filesystem struct {
 	Name string
 	// FsType of the Filesystem
 	FsType string
+	// DiskType of the Filesystem
+	DiskType string
```
muhamadazmy: Do we really need this type, because as far as i remember, we have a VolumeType defined somewhere with probably exactly possible values.

pkg/storage/storage.go
```diff
@@ -455,27 +455,28 @@ func (s *storageModule) ListFilesystems() ([]pkg.Filesystem, error) {
 // Path return the path of the mountpoint of the named filesystem
 // if no volume with name exists, an empty path and an error is returned
 func (s *storageModule) Path(name string) (pkg.Filesystem, error) {
-	for idx := range s.pools {
```
LeeSmet, line 458: Is there a particular reason that these loops are changed from index based access to value copy access? Since there seems to be a lot of changes here which don't add anything

DylanVerstraete: No reason, was looking at the code and I don't think using indexes in for loops is very readable. That's why I changed this

pkg/storage.go
DylanVerstraete: I think we do yes, only this Filesystem is exposed with zbus

muhamadazmy: I don't get your answer

DylanVerstraete: Are you talking about this: https://github.com/threefoldtech/zos/blob/736c1119840e06aa6919935826c7abc01c8ddcf3/pkg/storage/filesystem/filesystem.go#L17 ?

pkg/storage/storage.go
LeeSmet: Thats actually highly opinionated. I for one think that this is perfectly readable. I'm not really a fan of changing code because someone thinks it looks prettier in a certain way, else you end up with a bunch of pull requests all changing code back and forth. As long as it passes the linters it should be fine.

Also, while it is not the case here, this change is not purely cosmetic. There is actually a subtle difference in looping based on indexes, or by assigning a value.

DylanVerstraete: As I aggree with you, but I also think consistency is a good thing. The most of provisiond's for loops are not using indexes since the disk shutdown feature was introduced. Ideally we want code to be consistent. But I will revert if you insist.

LeeSmet: Following that argument, it should actually have been the disk shutdown PR where the loops needed to change style to keep consistency with the ones already there. But the main thing I wanted to highlight here is that I don't really like changing code to just change code.

That being said, you can leave it as it is. Otherwise it would be another change just to change it :confused:

LeeSmet (commented):
lgtm as long as the other comment gets sorted out

pkg/storage.go
DylanVerstraete: @muhamadazmy ?

muhamadazmy: https://github.com/threefoldtech/zos/blob/736c1119840e06aa6919935826c7abc01c8ddcf3/pkg/storage.go#L47

zaibon (approved):
LGTM

PR #996, opened 2020-10-13
Contd/decomission on crashes

### LINKED ISSUE #957: contd: improve retry logic when container fails to start
2020-09-22_3494x695_scrot

muhamadazmy:
Fixes #957 

- `contd` watches for events from containerd 
- on task failure, contd keep tracks of how many failures happens in one minute. 
- if the number of failure above a certain threshold, a call to provisiond is made to delete this reservation
- provisoind updates the explorer with the failure reason, and decommission the reservation

### DIFF
--- a/pkg/container/watch.go
+++ b/pkg/container/watch.go
@@ -26,7 +26,7 @@ func (c *Module) handlerEventTaskExit(ns string, event *events.TaskExit) {
 		c.failures.Set(event.ContainerID, int(0), cache.DefaultExpiration)
 	}
 
-	if marker == permenant {
+	if marker == permanent {
 		// if the marker is permenant. it means that this container
 		// is being deleted. we don't need to take any more action here
 		// (don't try to restart or delete)
--- a/pkg/container/watch.go
+++ b/pkg/container/watch.go
@@ -8,7 +8,7 @@ import (
 	"github.com/containerd/typeurl"
 	"github.com/patrickmn/go-cache"
 	"github.com/rs/zerolog/log"
-	"github.com/threefoldtech/zos/pkg"
+	"github.com/threefoldtech/zos/pkg/stubs"
 )
 
 func (c *Module) handlerEventTaskExit(ns string, event *events.TaskExit) {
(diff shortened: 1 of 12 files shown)


muhamadazmy:
This solution works fine in case the process dies for ANY reason. BUT the problem is when contained can't 'respawn'  the process (failure in the filesystem) in that case the system never receives a 'task exit' event. hence no way to start deprovisioning.

The only possible solution i can think of is by handling the restarts ourselves (not in containerd) 
- when an entrypoint dies
 - check the failure counters
   - threshold exceeded: -> decommission
   - threeshold not exceeded: -> increment counters + restart
      - restart fails: -> (this is permanent ) decommission
      - restart succeed: continue

muhamadazmy:
I have changed the code again so the restart handling is completely in contd, we then can decide if we need to stop the task, try to restart it and if the failure is permanent.

pkg/provision/engine.go
```diff
@@ -320,6 +320,29 @@ func (e *Engine) reply(ctx context.Context, result *Result) error {
...
+		return err
+	}
+
+	ctx := context.Background()
+	result, err := e.buildResult(id, r.Type, fmt.Errorf(reason), nil)
+	if err != nil {
+		return errors.Wrapf(err, "failed to build result object for reservation: %s", id)
+	}
+
+	if err := e.reply(ctx, result); err != nil {
```
zaibon: shouldn't we decommission before sending information to the explorer ?
What if the decommission function fails ? We would be left with a workload that is impossible to remove, no ?

pkg/container/watch.go
```diff
@@ -0,0 +1,115 @@
...
+		<-time.After(restartDelay) // wait for 2 seconds
+		reason = c.start(ns, event.ContainerID)
+	} else {
+		reason = fmt.Errorf("deleting container due to so many crashes")
+	}
+
+	if reason != nil {
+		log.Debug().Msg("deleting container due to so many crashes")
+
+		stub := stubs.NewProvisionStub(c.client)
```
zaibon, line 59: maybe move this into the Module struct to avoid re-creating a new connection every time ?

pkg/provision/engine.go
muhamadazmy: Well, i switched the 2 steps multiple time and was not convinced which must come first. What if decomission fails? Should we still set the error or not?

pkg/container/watch.go
muhamadazmy: We don't create a mew connection. The connection is available on the module structure. It's the stub that is created which is just a wrapper around the connection to expose the module methods. Hence it can be deallocated cleanely after the stack is over.

muhamadazmy:
I think i still need to handle normal decommission flow, to make sure we don't try to restart the container 

**done**

zaibon (approved):
Code LGTM. 
@muhamadazmy What kind of test did you already do around this ?

muhamadazmy:
@zaibon 
- Provision a container
- Kill the entrypoint, you notice that it shows the correct number of crashes, and the container is restarted
- Kill it again quickly, and the failures count is incremented.
- Wait for a minute and try again, the count is reset to 1
- Killing multiple times, and then the container is deprovisioned, and the state on the explorer is updated to the write reason (so many failures)
- We recreate a container, and then kill it's filesystem, then kill the entrypoint (if it didn't already die)
- The container tries to restart, but hits an immediate error on start, the restart process is stopped and the container is decommissioned and the cause of error is reported back to the explorer.
- Finally, test the normal decommission flow initiated by the user, and make sure the container does not auto-restart.

PR #999, opened 2020-10-14
cleanup cached files

muhamadazmy:
Make sure cached files older than 30 days are cleaned from the cache.


zaibon (commented):
Code LGTM, but can we have a test for the new `CacheCleaner` function please.

maxux (approved):

zaibon (approved):

PR #1003, opened 2020-10-21
provisiond: skip failed reservation when syncing cache

zaibon:
when provisiond restarts, it does walk over all the reservation in the
cache to rebuild the state of the resource units statser.
We now properly skip the reservation that failed to provision

### DIFF
--- a/pkg/provision/primitives/cache/cache.go
+++ b/pkg/provision/primitives/cache/cache.go
@@ -53,7 +53,67 @@ func NewFSStore(root string) (*Fs, error) {
 
 	log.Info().Msg("restart detected, keep reservation cache intact")
 
+	if err := store.updateReservationResults(root); err != nil {
+		log.Err(err).Msgf("error while updating reservation results")
+		// Don't return error, a cache is still valid even without the results
+		return nil, nil
+	}
+
 	return store, nil
+}
+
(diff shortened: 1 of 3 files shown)


pkg/provision/primitives/cache/cache.go
```diff
@@ -115,7 +171,11 @@ func (s *Fs) Add(r *provision.Reservation) error {
 		return err
 	}
 
-	f, err := os.OpenFile(filepath.Join(s.root, r.ID), os.O_CREATE|os.O_EXCL|os.O_WRONLY, 0660)
+	flags := os.O_CREATE | os.O_WRONLY
+	if !update {
+		flags |= os.O_EXCL
+	}
```
zaibon, line 177:
```suggestion
	flags :=  os.O_WRONLY
	if !update {
		flags |= os.O_EXCL | os.O_CREATE
	}
```
Seems you only need the O_CREATE flag only when you do NOT update, right ?

muhamadazmy: Do we care? create will create it if the file does not exist, which update should still do if the file is gone for any reason.

zaibon: the previous behavior was we return an error if there was already a reservation present. But I guess you are write in both case we should create it. so ignore me :-P

PR #1007, opened 2020-10-22
create a qgroup if none exists yet

### LINKED ISSUE #1006: storaged: weird behavior when trying to mount cache
found on node qzuTJJVd5boi6Uyoco1WWnSgzTb7q8uN79AjBT9x9N3
```
[+] storaged: 2020-10-22T07:07:52Z debug Checking pools for existing cache
[+] storaged: 2020-10-22T07:07:52Z debug Found existing cache at /mnt/7e3aa84f-e21b-4b41-9a26-b5386b445e86/zos-cache
[+] storaged: 2020-10-22T07:07:52Z info set cache quota to 100 GiB
[+] storaged: 2020-10-22T07:07:52Z error failed to set cache quota error="ERROR: unable to limit requested quota group: No such file or directory\n"
[+] storaged: 2020-10-22T07:07:52Z debug Cache partition already mounted in /var/cache
[+] storaged: 2020-10-22T07:07:52Z debug Pool 7e3aa84f-e21b-4b41-9a26-b5386b445e86 has: 4 subvolumes
[+] storaged: 2020-10-22T07:07:52Z debug Pool a2de348c-74c9-4d37-b820-fdf89cabbfa4 has: 1 subvolumes
[+] storaged: 2020-10-22T07:07:52Z debug Pool 1a948a44-f5da-4a17-86fc-30ebf2f15eb5 has: 0 subvolumes
[+] storaged: 2020-10-22T07:07:52Z info Shutting down disk /dev/sdb ...
[+] storaged: 2020-10-22T07:07:53Z info Disk /dev/sdb is shutdown
[+] storaged: 2020-10-22T07:07:53Z debug Pool cedeaaaa-f76b-4abc-99bf-0766f910b1e3 has: 0 subvolumes
[+] storaged: 2020-10-22T07:07:53Z info Shutting down disk /dev/sdc ...
[+] storaged: 2020-10-22T07:07:53Z info Disk /dev/sdc is shutdown
[+] storaged: 2020-10-22T07:07:53Z info Finished initializing storage module
[+] storaged: 2020-10-22T07:07:53Z error failed to initialize virtual disk module error="device is not mounted" limited-cache=false
```

DylanVerstraete:


pkg/storage/filesystem/btrfs_utils.go
```diff
@@ -157,6 +157,12 @@ func (u *BtrfsUtil) QGroupEnable(ctx context.Context, root string) error {
 	return err
 }
 
+// QGroupCreate creates a qgroup
+func (u *BtrfsUtil) QGroupCreate(ctx context.Context, id string, path string) error {
```
LeeSmet, line 161: Why are we using this way to create a qgroup now?

pkg/storage/filesystem/btrfs_utils.go
```diff
@@ -157,6 +157,12 @@ func (u *BtrfsUtil) QGroupEnable(ctx context.Context, root string) error {
 	return err
 }
 
+// QGroupCreate creates a qgroup
```
LeeSmet, line 160: Why are we using this to create a qgroup now?

pkg/storage/filesystem/btrfs.go
```diff
@@ -591,6 +592,26 @@ func (v *btrfsVolume) Usage() (usage Usage, err error) {
...
+	// Check if there is a qgroup for before we try to limit
+	_, ok := groups[fmt.Sprintf("0/%d", info.ID)]
+	if !ok {
+		log.Debug().Msgf("Creating qgroup 0/%d, at %v", info.ID, v.Path())
+		err = v.utils.QGroupCreate(ctx, strconv.Itoa(info.ID), v.Path())
+		if err != nil {
+			return err
+		}
+	}
+
```
LeeSmet, line 614: What is the exact reason for this logic being added? In the past there was just the QGroupLimited call, which took care of everything as far as I know. So what is the underlying reason that this might be needed?

LeeSmet (changes requested):

pkg/storage/filesystem/btrfs.go
zaibon: we've identify a node which failed to set the qgroup limit on the cache at the boot of storaged. Turns out the reason is there was no qgroup for this specific subvolume.

DylanVerstraete: @LeeSmet check the linked issue

LeeSmet: Previously we never explicitly created the qgroup. I understand that this works in the scenario you described, but I want to know why this is suddenly needed. I.E why can limit not implicitly create the subgroup anymore? And more importantly, why is the subgroup gone in the first place? That seems to indicate a bug somewhere else in the code to me, which we are just covering up here.

From the linked issue, the error is `ERROR: unable to limit requested quota group: No such file or directory\n`. Are we sure that is related to qgroups, or is there a possibility tha the path that was passed happened to just be wrong for some reason?

DylanVerstraete: It could indeed indicate that there is another issue somewhere else, perhaps maybe while testing the cleanup script and the script did not account for the zos-cache.. 

Also I debugged the error that was listed in the issue and it seemed like there was nothing wrong with passing the `path`. I also tried using btrfs command myself explicitly passing the path and it also failed. So that's how I verified this issue.

LeeSmet: > Also I debugged the error that was listed in the issue and it seemed like there was nothing wrong with passing the path. I also tried using btrfs command myself explicitly passing the path and it also failed

I'm confused, exactly what did you do and what succeeded/failed?

DylanVerstraete: > Are we sure that is related to qgroups, or is there a possibility tha the path that was passed happened to just be wrong for some reason?

I verified that this is not the issue

LeeSmet: so, if I understand correctly, the situation is as follows:

1. Cache is created
2. Cache quota is enabled en set to 100GB
3. Something happens, but we have no clue what
4. Storaged restarts
5. Storaged finds the mounted cache and tries to enable quota, but fails for an unknown reason

This suggests to me that there really is an issue somewhere else we should fix. Sure if we do this, it'll allow the cache mount flow to continue. But we still have no clue what really went wrong, and more importantly, what other issues this may cause that we just haven't noticed yet.

Also, the regular flow just enables quota on a path and does not specify the qgroup ID explicitly, nor create it explicitly. So if the qgroup is just deleted, then the old flow should continue to work. That means that even if this is a rogue cleanup which forgot to account for the cache, the cleanup routine in itself still leaves something lingering behind.

muhamadazmy (approved):

pkg/storage/filesystem/btrfs.go
DylanVerstraete: I double checked the cleanup routine and I did not find anything that could cause this issue again. I suggest we close this PR until we might see this issue pop up again so we can find the cause. I fixed this one node that had this issue.

DylanVerstraete:
Closing because this was an isolated occurrence on a node that we did some manual testing on. I have fixed this node in favor of deploying this code that will affect all other nodes as well. This can be reopened if needed.

PR #1012, opened 2020-10-26
Fix decomission container is netns does not exist

### LINKED ISSUE #1011: Failed to decomission container
On devnet node: BpTAry1Na2s1J8RAHNyDsbvaBSM3FjR4gMXEKga3UPbs 

```
2020-10-26 16:16:14 | [+] provisiond: 2020-10-26T15:16:14Z error failed to decommission reservation 1920-1 error="decommissioning of reservation failed: failed to delete container network namespace: failed to Statfs \"/var/run/netns/1920-1\": no such file or directory"
2020-10-26 16:16:14 | [+] provisiond: 2020-10-26T15:16:14Z error failed to inspect container for decomission error="container \"1920-1\" in namespace \"ns1\": not found" container=1920-1
```

DylanVerstraete:
Seems like an error was returned from the `ns` package that we were not matching.

### DIFF
--- a/pkg/network/nr/container.go
+++ b/pkg/network/nr/container.go
@@ -177,17 +177,17 @@ func (nr *NetResource) Leave(containerID string) error {
 		Str("container", containerID).
 		Msg("delete container network namespace")
 
-	ns, err := namespace.GetByName(containerID)
-	if err != nil && !os.IsNotExist(err) {
+	namespc, err := namespace.GetByName(containerID)
+	if _, ok := err.(ns.NSPathNotExistErr); ok {
+		return nil
+	} else if os.IsNotExist(err) {
+		return nil
+	} else if err != nil {
 		return err
 	}
-	if os.IsNotExist(err) {
-		// nothing to do, early return
-		return nil
-	}
-	defer ns.Close()
+	defer namespc.Close()
 
-	err = namespace.Delete(ns)
+	err = namespace.Delete(namespc)
 	if err != nil {
 		return err
 	}


pkg/network/nr/container.go
```diff
@@ -177,17 +177,17 @@ func (nr *NetResource) Leave(containerID string) error {
...
 		Msg("delete container network namespace")
 
-	ns, err := namespace.GetByName(containerID)
-	if err != nil && !os.IsNotExist(err) {
-		return err
-	}
-	if os.IsNotExist(err) {
-		// nothing to do, early return
+	namespc, err := namespace.GetByName(containerID)
+	if _, ok := err.(ns.NSPathNotExistErr); ok {
```
LeeSmet, line 181: The proper way to do this these days is with `errors.As(...)`, since the error itself does not seem to be used anyhow you might even be able to get away with `errors.Is`, though I'm not sure about that

LeeSmet (changes requested):

pkg/network/nr/container.go
DylanVerstraete: Tried both those options but I did not get that to work..

LeeSmet: That's rather strange since the documentations of errors.As here (https://golang.org/pkg/errors/) literally explains that `As` should be preferred over this construct

DylanVerstraete: Ah but wait, I used a pointer while it should have been a concrete type.

LeeSmet (approved):

Issue #1008, opened 2020-10-24
Failed to deploy network in time


AhmedHanafy725:
**Explorer:** testnet
**workload:** https://explorer.testnet.grid.tf/api/v1/reservations/workloads/493452

After 10 mins waiting, it failed to deploy, seems nodes doesn't accept network reservations.

m-motawea:
doesn't respond to containers as well
https://explorer.testnet.grid.tf/api/v1/reservations/workloads/493621

zaibon:
@maxux seems provisiond is stuck on some connection to zdb. Please check the zdb on node `72CP8QPhMSpF7MbSvNR1TYZFbTnbRiuyvq5xwcoRNAib`.

maxux:
After investigation, there is a container which have a zombie process (`sync`) which cannot be killed. Container cannot be stopped and current code waits forever for container (in `contd`) to get stopped before handling others commands.

I have a hotfix which doesn't block `contd` forever by returning an error if process cannot be killed after some time, but this lead to a forever retry of decommission after that. At least, provision are not blocked anymore and node can still work, but not in good condition.

What I suggest is implementing a `fatal error without (direct) retry` flag which would notify that decommission were blocked by a system error, and we can't do anything for now and try again later...

In this case, I don't know what caused the `sync` to block, I don't see anything on `dmesg` but maybe one disk is corrupted and unless we can unblock kernel with this device, the process will never dies (except with a `reboot`).

m-motawea:
this node on testnet is not accepting workloads as well
3dAnxcykEDgKVQdTRKmktggL2MZbm3CPSdS9Tdoy4HAF

PR #1015, opened 2020-10-26
contd: do not wait forever on container delete

maxux:
Request of container delete waits forever if the process cannot be killed. The workflow tries to `SIGTERM` the running process then `SIGKILL` after some try, but it never stop trying `SIGKILL`.
    
 If process is in `D` or `Z` state, process won't be killed if stuck on kernel call (occured with `sync` on testnet). This make contd not responding because stuck on deleting.
    
This fix will make contd responding again and not being stuck forever but since an error is returned, provisiond will try to re-delete it later.

This PR add aswell more verbosity to `contd` to be less blind on what happen.
This PR is related to #1008

### DIFF
--- a/pkg/container/container.go
+++ b/pkg/container/container.go
@@ -443,12 +443,21 @@ func (c *Module) Delete(ns string, id pkg.ContainerID) error {
 			if trials <= 0 {
 				signal = syscall.SIGKILL
 			}
+
+			log.Debug().Str("id", string(id)).Int("signal", int(signal)).Msg("sending signal")
 			_ = task.Kill(ctx, signal)
 			trials--
+
 			select {
 			case <-exitC:
 				break loop
 			case <-time.After(1 * time.Second):
+			}
+
+			if trials < -5 {
+				msg := fmt.Errorf("cannot stop container, SIGTERM and SIGKILL ignored")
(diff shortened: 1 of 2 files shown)


pkg/container/container.go
```diff
@@ -413,13 +421,18 @@ func (c *Module) List(ns string) ([]pkg.ContainerID, error) {
...
 func (c *Module) Delete(ns string, id pkg.ContainerID) error {
+	log.Info().Str("id", string(id)).Str("ns", ns).Msg("delete container")
+
 	client, err := containerd.New(c.containerd)
 	if err != nil {
 		return err
 	}
 	defer client.Close()
+
+	// log.Debug().Msg("fetching namespace")
```
DylanVerstraete, line 432: is there a reason to keep this commented logs?

DylanVerstraete (changes requested):

LeeSmet (approved):

zaibon (approved):

pkg/container/container.go
maxux: I will re-enable all debug commented.
But first we should enable `-debug` flag for `contd` otherwise it will be printing debug by default and generate lot of logs

PR #1029, opened 2020-10-28
do use the proper size when doing capacity planning for 0-DB namespace

### LINKED ISSUE #1027: node report more storage reserved then total available
6m9gFPTYLRiCH7FD8yU5VWwXZdfegKUvXZgrJoS1EJfQ
2fi9ZZiBGW4G9pnrN656bMfW6x55RSoHDeMrd9pgSA8T
72CP8QPhMSpF7MbSvNR1TYZFbTnbRiuyvq5xwcoRNAib

zaibon:
fixes #1027

### DIFF
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -344,7 +344,7 @@ func (s *storageModule) CreateFilesystem(name string, size uint64, poolType pkg.
 		return pkg.Filesystem{}, fmt.Errorf("invalid volume name. zdb prefix is reserved")
 	}
 
-	fs, err := s.createSubvol(size, name, poolType)
+	fs, err := s.createSubvolWithQuota(size, name, poolType)
 	if err != nil {
 		return pkg.Filesystem{}, err
 	}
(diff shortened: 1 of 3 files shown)


muhamadazmy (approved):

pkg/storage/zdb.go
```diff
@@ -139,7 +139,7 @@ func (s *storageModule) Allocate(nsID string, diskType pkg.DeviceType, size uint
 
 		// we create the zdb instance with 0 (unlimited) because this subvolume is gonna
```
DylanVerstraete: Good catch, maybe update this comment as well then?

pkg/storage/storage.go
```diff
@@ -575,9 +575,24 @@ func (s *storageModule) ensureCache() error {
...
+	if err = volume.Limit(size); err != nil {
+		log.Error().Err(err).Str("volume", volume.Path()).Msg("failed to set volume size limit")
+		return nil, err
+	}
+
+	return volume, nil
+}
+
+// createSubvol creates a subvolume with the given name
+// if the requested disk type does not have a storage pool available, an error is returned
```
muhamadazmy: documentation update is required to state that this size is only for capacity planning (Find a pool that can support the requested size) and that it doesn't limit the size of the subvol to that size

PR #1030, opened 2020-10-28
Add checks if pools are mounted

### LINKED ISSUE #1023: error failed to initialize virtual disk module error
`  2020-10-28 12:02:11  
[+] storaged: 2020-10-28T11:02:11Z error failed to initialize virtual disk module error="Not enough space left in pools of this type ssd" limited-cache=false`

466X86s2ctLmL2Q6R7SrJ4VBMyt9h9zJDCSZhHUSQ9py

DylanVerstraete:
closes #1023

### DIFF
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -457,6 +457,9 @@ func (s *storageModule) ListFilesystems() ([]pkg.Filesystem, error) {
 // if no volume with name exists, an empty path and an error is returned
 func (s *storageModule) Path(name string) (pkg.Filesystem, error) {
 	for _, pool := range s.pools {
+		if _, mounted := pool.Mounted(); !mounted {
+			continue
+		}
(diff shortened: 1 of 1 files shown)


pkg/storage/storage.go
```diff
@@ -787,6 +790,9 @@ func (s *storageModule) periodicallyCheckDiskShutdown() {
 // If a disk is on and it is not mounted then it is not supposed to be on, turn it off
 func (s *storageModule) shutdownDisks() {
 	for _, pool := range s.pools {
+		if _, mounted := pool.Mounted(); !mounted {
+			continue
```
muhamadazmy: If the pool is not mounted, is there is a way to make sure that it's also shutdown? and shut it down if it's not.

DylanVerstraete: ah wait this should not be there indeed

muhamadazmy (approved):

PR #1040, opened 2020-10-29
WIP

### LINKED ISSUE #861: Provisiond should only keep one single instance of a network in cache
Instead of just keeping all reservation ever made for a same network in cache, only the latest one should be kept. 
When a new version come to the node, the it should replace the deprecated one from cache instead of just being stored next to it.

zaibon:


pkg/provision/engine.go
```diff
@@ -242,23 +242,8 @@ func (e *Engine) provision(ctx context.Context, r *Reservation) error {
...
-		if err != nil {
-			return errors.Wrap(err, "failed to check if network exists")
-		}
-		if exists {
-			return nil
-		}
-	}
-
+	// TODO: need to be able to decide if the increment or not...
+	// maybe this should be also moved into the cache ?
```
zaibon, line 246: This is a bit of a problem cause now we will count the number of network resource wrong again.

zaibon:
this approach won't work. Any time provisiond restart it would re-provision the old network workloads...

PR #1041, opened 2020-10-30
fix provisiond cleanup for reservation in state deleted

### LINKED ISSUE #1037: "failed to create read-write subvolume for 0-fs: Not enough space left in pools of this type ssd",
https://explorer.testnet.grid.tf/api/v1/reservations/workloads/494762

DylanVerstraete:
Now checking reservations for state `deleted` and `invalid` as well. Also this client was returning an error because the `get` method was trying to unmarshal in an interface. Using `NodeWorkloadGet` now, it's the same result.


zaibon (approved):

LeeSmet (approved):
:ship:

PR #1052, opened 2020-11-10
Zui 80x24

DylanVerstraete:


cmds/zui/prov.go
```diff
@@ -43,13 +40,44 @@ func provisionRender(client zbus.Client, grid *ui.Grid, render *Flag) error {
...
+	sysMonitor := stubs.NewSystemMonitorStub(client)
+	stream, err := sysMonitor.CPU(context.Background())
+	if err != nil {
+		return errors.Wrap(err, "failed to start cpu monitor stream")
+	}
+
+	go func() {
+		for point := range stream {
+			prov.Mutex.Lock()
+			for _, cpu := range point {
```
muhamadazmy: Each point returned from the stream has CPU stats for each CPU core on the node. so what u do now here is that u go over each cpu update the cell value and then render, which will basically show only LAST cpu usage not all of the them. 
What we need to do here is that the stream can return aggregated values (total cpu usage) + a list of individual cpus. then in ZUI can show the aggregated cpu load instead

cmds/zui/net.go
```diff
@@ -82,56 +82,65 @@ func addressRender(ctx context.Context, table *widgets.Table, client zbus.Client
...
 	if err := addressRender(ctx, addresses, client, render); err != nil {
 		return err
 	}
 
-	monitor := stubs.NewSystemMonitorStub(client)
-	stats, err := monitor.Nics(ctx)
-	if err != nil {
-		return errors.Wrap(err, "failed to start net monitor stream")
-	}
+	/*
```
muhamadazmy: please remove out all commented code if it's not needed anymore

cmds/zui/prov.go
```diff
@@ -43,13 +40,47 @@ func provisionRender(client zbus.Client, grid *ui.Grid, render *Flag) error {
...
+	if err != nil {
+		return errors.Wrap(err, "failed to start cpu monitor stream")
+	}
+
+	go func() {
+		var aggregatedValue float64
+		for point := range stream {
+			aggregatedValue = 0
+			for _, cpu := range point {
+				aggregatedValue += cpu.Percent
```
muhamadazmy: May you can't ADD percentage like that :D the system monitor uses psutils with some tools to return an avg system usage

DylanVerstraete: erm how and where then? `cpu.Percent` can be a value like for example `2` or `25`, you can add integers and get an average of them as far as I remember...

PR #1053, opened 2020-11-10
identityd drop dependency on 0-fs

muhamadazmy:
- Modify the upgrade process to use 0-fs library
- The current code uses not release 0-fs code (hence draft) 
  - We need to merge https://github.com/threefoldtech/0-fs/pull/44 first and create a release before we can merge this


cmds/Makefile
```diff
@@ -8,10 +8,13 @@ ldflags = '-w -s -X $(version).Branch=$(branch) -X $(version).Revision=$(revisio
...
 	strip $(OUT)/*
 
-.PHONY: output clean
+.PHONY: output clean identityd
 
 output:
 	mkdir -p $(OUT)
 
+identityd:
+	cd identityd && GOOS=linux go build -ldflags $(ldflags) -o $(OUT)/identityd
```
zaibon, line 17: this is needed cause you now need gco ?
isn't cgo only used by the fuse part of 0-Fs ? I guess you do not need it just to read the flist is it ?

zaibon (approved):

cmds/Makefile
muhamadazmy: No actually this is needed by sqlite !!

PR #1054, opened 2020-11-12
Check if we have enough space in the pool for a vdisk

muhamadazmy:
Also fixes a precision issue with the counters

### DIFF
--- a/pkg/storage.go
+++ b/pkg/storage.go
@@ -119,7 +119,7 @@ type Usage struct {
 	Used uint64
 }
 
-// Filesystem represents a storage space that can be used as a filesystem
+// Filesystem represents a storage space that can be used as a filesystem`
 type Filesystem struct {
 	// Filesystem ID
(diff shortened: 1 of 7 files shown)


pkg/storage.go
```diff
@@ -119,7 +119,7 @@ type Usage struct {
 	Used uint64
 }
 
-// Filesystem represents a storage space that can be used as a filesystem
+// Filesystem represents a storage space that can be used as a filesystem`
```
LeeSmet: Out of curiosity, what's the backtick doing here?

LeeSmet (approved):

zaibon (approved):
LGTM

LeeSmet:
It seems you need to regenerate the stubs though as the interface changed

pkg/storage.go
muhamadazmy: no idea.

PR #1056, opened 2020-11-12
Create chatflows workflow

AhmedHanafy725:
Add workflow to run chatflows tests with each pre-release.

### DIFF
--- a/.github/workflows/chatflows.yml
+++ b/.github/workflows/chatflows.yml
@@ -1,8 +1,6 @@
 name: Chatflows Tests
 
-on:
-  release:
-    types: [ prereleased ]
+on: workflow_dispatch
 
 jobs:
   build:
--- a/.github/workflows/chatflows.yml
+++ b/.github/workflows/chatflows.yml
@@ -1,0 +1,35 @@
+name: Chatflows Tests
+
+on:
+  release:
+    types: [ prereleased ]
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Gathering deps
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y git python3-pip python3-venv python3-setuptools tmux redis restic nginx cron wireguard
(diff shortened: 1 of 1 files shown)


zaibon:
2 things here.
1. Can you describe exactly what this workflow does and what is the impact on the network when it runs ?

2. I would like to change the trigger event of this workflow from creation of pre-release to a manual one. See: https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/manually-running-a-workflow

@muhamadazmy you thought on this ?

.github/workflows/chatflows.yml
```diff
@@ -0,0 +1,35 @@
...
+          sudo poetry install
+      - name: Run tests
+        env:
+          TNAME: ${{ secrets.TNAME }}
+          EMAIL: ${{ secrets.EMAIL }}
+          WORDS: ${{ secrets.WORDS }}
+          WALLET_SECRET: ${{ secrets.WALLET_SECRET }}
+        run: |
+          cd js-sdk
+          sudo --preserve-env=TNAME --preserve-env=EMAIL --preserve-env=WORDS --preserve-env=WALLET_SECRET \
```
muhamadazmy, line 32: i think it should be $WORDS, and $WALLET_SECRET also why not do that directly via the ${{ secrets.WORDS }} directly in the run command ?

muhamadazmy:
While this indeed useful but an automated workflow is not the right way to do it. This will run DURING the pre-release building which mans the testnet still did not get this pre-release. hence the tests are running against an older release that is currently on testnet.

The right way imho to do it is by running it manually hence it does not even have to be part of zos repo

.github/workflows/chatflows.yml
AhmedHanafy725: I can't use Github secrets while using `sudo`, so I take them to `sudo` using their names without `$`

AhmedHanafy725:
The workflow is changed to run manually.
This workflow is running most of the dashboard solutions, marketplace, and threebot deployer chatflows.

PR #1063, opened 2020-11-17
bins: add cloud-hypervisor extra package

maxux:
Needed for #1061 
https://hub.grid.tf/tf-autobuilder/cloud-hypervisor-0.11.0.flist.md


bins/packages/cloudhypervisor/cloudhypervisor.sh
```diff
@@ -0,0 +1,62 @@
+CLOUDHYPERVISOR_VERSION="0.13.0"
+CLOUDHYPERVISOR_CHECKSUM="44bd96c255b8d573a83eef2364d934ae"
+CLOUDHYPERVISOR_LINK="https://github.com/cloud-hypervisor/cloud-hypervisor/archive/v${CLOUDHYPERVISOR_VERSION}.tar.gz"
+CLOUDHYPERVISOR_RUST_TOOLCHAIN="1.48.0"
```
LeeSmet: Is there any reason we are building with this version and not latest stable (currently 1.50)?

pkg/vm/ch.go
```diff
@@ -0,0 +1,97 @@
...
+
+	tmp, err := ioutil.TempFile("", "vm-exec-*.sh")
+	if err != nil {
+		return errors.Wrap(err, "failed to create vm exec script")
+	}
+
+	log.Debug().Str("name", m.ID).Str("exec", tmp.Name()).Msg("using exec file")
+	defer func() {
+		tmp.Close()
+		//os.Remove(tmp.Name())
```
LeeSmet: commented code

LeeSmet (changes requested):

bins/packages/cloudhypervisor/cloudhypervisor.sh
muhamadazmy: no

pkg/vm/ch.go
muhamadazmy: ah, good catch

LeeSmet (approved):
:ship:

PR #1082, opened 2020-11-24
If no-keep-alive flag is set do not restart the machine

muhamadazmy:
This is needed by the provision process for the installation
to work since the machine shutsdown when it's done with the
installations

### DIFF
--- a/pkg/provision/primitives/kubernetes.go
+++ b/pkg/provision/primitives/kubernetes.go
@@ -201,6 +201,7 @@ func (p *Provisioner) kubernetesInstall(ctx context.Context, name string, cpu ui
 		InitrdImage: imagePath + "/k3os-initrd-amd64",
 		KernelArgs:  cmdline,
 		Disks:       disks,
+		NoKeepAlive: true, //machine will not restarted automatically when it exists
 	}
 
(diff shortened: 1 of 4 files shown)


LeeSmet (approved):

pkg/provision/primitives/kubernetes.go
```diff
@@ -201,6 +201,7 @@ func (p *Provisioner) kubernetesInstall(ctx context.Context, name string, cpu ui
 		InitrdImage: imagePath + "/k3os-initrd-amd64",
 		KernelArgs:  cmdline,
 		Disks:       disks,
+		NoKeepAlive: true, //machine will not restarted automatically when it exists
```
xmonader, line 204: Shouldn't it be `KeepAlive: false` instead of an inverted option? Also, I believe it's more realistic to have the default to be restartable, that's what I expect at least as a user

muhamadazmy: This is not exposed to the user or the grid. This is only internal and used by the installation process for k8s.
The flag is inverted because the zero value is false which will be the dedault expected behavior unless requested otherwise

zaibon (approved):

PR #1083, opened 2020-11-25
Make sure to delete the machine if installation fails

### LINKED ISSUE #1069: k8s VM is not deleted after workload failure and after decommission on devnet
trying public ips on devnet on node `2anfZwrzskXiUHPLTqH1veJAia8G6rW2eFLy5YFMa1MP`,
I created a VM https://explorer.devnet.grid.tf/api/v1/reservations/workloads/2083 and it failed with error `failed to start k3s VM in 5 minutes` then shortly after that tried to deploy a new VM https://explorer.devnet.grid.tf/api/v1/reservations/workloads/2084 on the same node it failed with error `kubernetes VM with the same name exists on network pub`

seems the first VM https://explorer.devnet.grid.tf/api/v1/reservations/workloads/2083 is still running although next action is 6.
even after decomission it is still not deleted and I can ssh into the VM.

related to https://github.com/threefoldtech/home/issues/925

muhamadazmy:
Fixes #1069

We always make sure to delete the vm (even if installation failed)

### DIFF
--- a/pkg/provision/primitives/kubernetes.go
+++ b/pkg/provision/primitives/kubernetes.go
@@ -202,11 +202,7 @@ func (p *Provisioner) kubernetesInstall(ctx context.Context, name string, cpu ui
 		InitrdImage: imagePath + "/k3os-initrd-amd64",
 		KernelArgs:  cmdline,
 		Disks:       disks,
-		// machine will not restarted automatically when it exists
-		// because it's part of the installation that the machine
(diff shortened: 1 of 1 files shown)


muhamadazmy:
This already have been fixed by this commit here https://github.com/threefoldtech/zos/commit/6ac1bdad030918f8e041ce454903df3c592bd68e

PR #1084, opened 2020-11-25
add memory capacity planning

### LINKED ISSUE #1079: provisiond: memory capacity planning
Current we don't do any capacity planning for workloads. We need to make sure the machine has enough memory to support the deployed workload type.

Proposal:
- Provision daemon checks workloads that requires memory resources
- Check coutners against available memory
- If resource memory is more than available free memory, then refuse the provision and return the reason

DylanVerstraete:


cmds/provisiond/main.go
```diff
@@ -139,6 +145,7 @@ func main() {
 		Statser:        statser,
 		ZbusCl:         zbusCl,
 		Janitor:        provision.NewJanitor(zbusCl, puller),
+		MemStats:       memStats,
```
muhamadazmy: You don't have to pass it, I think the engine can get that internally and keep it.

pkg/provision/engine.go
```diff
@@ -88,6 +92,7 @@ func New(opts EngineOps) *Engine {
 		statser:        opts.Statser,
 		zbusCl:         opts.ZbusCl,
 		janitor:        opts.Janitor,
+		memStats:       opts.MemStats,
```
muhamadazmy: I will probably only keep the `Total` memory instead of the entire struct (while it's not a problem) . But what we definitely need is to not allocate the full memory. ZOS needs at least 2GB of memory for it's services and rootfs.

So my suggestion is, engine can keep a single value of allowed memory which is `total - const` where the const is 2G

pkg/provision/primitives/counters.go
```diff
@@ -184,6 +186,35 @@ type resourceUnits struct {
...
+	switch r.Type {
+	case ContainerReservation:
+		requestedUnits, err := processContainer(r)
+		if err != nil {
+			return err
+		}
+
+		// If current MRU + requested MRU exceeds total, return an error
+		if uint64(c.CurrentUnits().Mru)+requestedUnits.HRU > memStats.Total {
+			return errors.New("reservation requested memory is to high")
```
muhamadazmy: This is a misleading error message, it means that this memory size is not supported by the system while in the matter of fact it's only a node capacity issue. May be change the message to `not enough free resources to support this memory size' or something similar

pkg/provision/primitives/counters.go
```diff
@@ -184,6 +186,35 @@ type resourceUnits struct {
...
+		}
+	case KubernetesReservation:
+		requestedUnits, err := processKubernetes(r)
+		if err != nil {
+			return err
+		}
+
+		// If current MRU + requested MRU exceeds total, return an error
+		if uint64(c.CurrentUnits().Mru)+requestedUnits.HRU > memStats.Total {
```
muhamadazmy: this code is identical to the one for containers, instead can we have the switch set the "requestedUnits" value, and then handle the calculations after

muhamadazmy (changes requested):

pkg/provision/primitives/counters.go
```diff
@@ -184,6 +186,35 @@ type resourceUnits struct {
 	CRU uint64 `json:"cru,omitempty"`
 }
 
+// CheckMemoryRequirements checks memory requirements for a reservation and compares it to whats in the counters
+// and what is the total memory on this node
+func (c *Counters) CheckMemoryRequirements(r *provision.Reservation, memStats *mem.VirtualMemoryStat) error {
```
zaibon: This method doesn't use anything from the `Counters` object. So we could just extract it from there.

DylanVerstraete: Yes it does use the `c.CurrentUnits()`, also this function uses the private functions in order to calculate the requested HRU by a workload. notice `processVolume` and `processKubernetes`

DylanVerstraete: It was placed here to avoid import cycles

pkg/provision/engine.go
```diff
@@ -95,11 +99,17 @@ func New(opts EngineOps) *Engine {
 func (e *Engine) Run(ctx context.Context) error {
 	cReservation := e.source.Reservations(ctx)
 
+	memStats, err := mem.VirtualMemory()
+	if err != nil {
+		log.Fatal().Err(err).Msg("failed retrieve memory stats")
```
muhamadazmy: please return the error, don't do log.Fatal here. it's not up to this method to decide.
Also this should have been done in the NewMethod, not the run.

DylanVerstraete:
Are we sure we want to substract 2 GIB from the available memory? If yes we need to mark this in the explorer as well! 

This will also make the nodes have less 'usable' RAM.

pkg/provision/engine.go
```diff
@@ -77,17 +81,23 @@ type EngineOps struct {
 // one reservation at a time. On error, the engine will log the error. and
 // continue to next reservation.
 func New(opts EngineOps) *Engine {
+	memStats, err := mem.VirtualMemory()
+	if err != nil {
```
muhamadazmy, line 85: If this fail, the code at line `100` will panic with nil pointer deference error. May be it's better if New returns an error also then the caller can decide to exit the program

muhamadazmy (changes requested):
I am working on a refactor for the provisiond engine that will make this much easier/cleaner to implement but for now we can merge this after addressing the review requests

muhamadazmy (approved):

PR #1085, opened 2020-11-25
Vm pub ip

### LINKED ISSUE #1046: Enable IPV6 in K8S VM

### LINKED ISSUE #1057: PublicIP: investigate how to configure the IP inside the k3s node
- VirtIO-FS might be a good option to import static config to the vm
- Other hackish methods 
  - Modify the vm disk after installation and before the 2nd boot to inject static config

### LINKED ISSUE #1069: k8s VM is not deleted after workload failure and after decommission on devnet

LeeSmet:
Before merging this, a new explorer version should be deployed on devnet, where the reservation version is incremeted to 2.

Nodes will need to be rebooted to get the new network setup which allows public IP's. I'd prefer this does not happen immediatly, since that will not be the case in the field, and we should test the code to see if the backward compatibility works as expected. To also be able to test the new stuff, it might be best to reboot half of devnet (the "upper part in 10.6.0.{11-14}, since one of those has been manually patched to have the new stuff.

Yggdrasil IP in containers and vms is not yet present in this PR though support for it *should* have been prepared.

Fixes #1069 
Close #1057 
Close #1046

### DIFF
--- a/pkg/network/ndmz/dualstack.go
+++ b/pkg/network/ndmz/dualstack.go
@@ -257,6 +257,12 @@ func (d *DualStack) SupportsPubIPv4() bool {
 	return d.hasPubBridge
 }
 
+// EnsureRoutable implements DMZ interface
+func (d *DualStack) EnsureRoutable() error {
+	// dualstack nodes have a dedicated physical interface so this is a no op
+	return nil
+}
+
 // waitIP4 waits to receives some IPv4 from DHCP
 // it returns the pid of the dhcp process or an error
 func waitIP4() error {
--- a/pkg/network/ndmz/dualstack.go
+++ b/pkg/network/ndmz/dualstack.go
@@ -60,9 +60,6 @@ func (d *DualStack) Create(ctx context.Context) error {
 	// The second option is used by default, and the first one is now legacy.
 	// However to not break existing containers, we keep the old one if networkd
 	// is restarted but the node is not. In this case, ndmz will already be present.
-	// TODO: properly check if we can switch from a phys iface to a bridge in
-	// between, probably by enumerating every device in every namespace and verifying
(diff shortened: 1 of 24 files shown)


pkg/network/ndmz/dualstack.go
```diff
@@ -201,6 +259,17 @@ func (d *DualStack) IP6PublicIface() string {
 	return d.ipv6Master
 }
 
+// SupportsPubIPv4 implements DMZ interface
+func (d *DualStack) SupportsPubIPv4() bool {
+	return d.hasPubBridge
+}
+
+// EnsureRoutable implements DMZ interface
+func (d *DualStack) EnsureRoutable() error {
```
zaibon: This statement is not really true. You can be dualstack and also only have one physical interface where all the traffic goes and that is linked to the zos bridge.
The type of ndmz is not related to the fact that there are 2 interfaces for traffic (management and public).
You could perfectly be a hidden node that has 2 physical interfaces

zaibon (changes requested):

PR #1096, opened 2020-12-01
Set static mac on public vm iface

LeeSmet:
Signed-off-by: Lee Smet <lee.smet@hotmail.com>


LeeSmet:
Don't merge yet, will need some changes due to #1098

PR #1114, opened 2020-12-06
Implement host firewall for public vm interfaces

### LINKED ISSUE #1102: Proper host filtering and firewalling for vms with public IP's
follow up of #1098 
CC @delandtj for what actually needs to happen

LeeSmet:
Closes #1102

### DIFF
--- a/cmds/networkd/nft.go
+++ b/cmds/networkd/nft.go
@@ -11,7 +11,7 @@ import (
 
 func ensureHostFw(ctx context.Context) error {
 	log.Info().Msg("ensuring existing host nft rules")
-	cmd := exec.CommandContext(ctx, "/bin/sh", "-c", "nft list ruleset")
+	cmd := exec.CommandContext(ctx, "nft", "list", "ruleset")
 	out, err := cmd.Output()
 	if err != nil {
 		return errors.Wrap(err, "could not load existing nft rules")
--- a/cmds/networkd/nft.go
+++ b/cmds/networkd/nft.go
@@ -1,0 +1,42 @@
+package main
+
+import (
+	"context"
+	"os/exec"
+
+	"github.com/pkg/errors"
+)
+
+func ensureHostFw(ctx context.Context) error {
+	cmd := exec.CommandContext(ctx, "/bin/sh", "-c", "nft list ruleset")
+	out, err := cmd.Output()
+	if err != nil {
+		return errors.Wrap(err, "could not load existing nft rules")
+	}
+
+	// there are already rules in place, return
+	if string(out) == "" {
+		return nil
+	}
+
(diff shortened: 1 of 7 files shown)


cmds/networkd/nft.go
```diff
@@ -0,0 +1,47 @@
...
+	"os/exec"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog/log"
+)
+
+func ensureHostFw(ctx context.Context) error {
+	log.Info().Msg("ensuring existing host nft rules")
+	cmd := exec.CommandContext(ctx, "/bin/sh", "-c", "nft list ruleset")
```
muhamadazmy:
```suggestion
	cmd := exec.CommandContext(ctx, "nft", "list", "ruleset")
```

pkg/provision/primitives/public_ip.go
```diff
@@ -26,11 +34,122 @@ func (p *Provisioner) publicIPProvision(ctx context.Context, reservation *provis
...
+	}
 	return network.DisconnectPubTap(reservation.ID)
 }
+
+func filterName(reservationID string) string {
+	return fmt.Sprintf("r-%s", reservationID)
+}
+
+func setupFilters(ctx context.Context, fName string, iface string, ip string, ipv6 string, mac string) error {
+	cmd := exec.CommandContext(ctx, "/bin/sh", "-c",
```
muhamadazmy, line 81: Would it be nicer if we use go `text/template` to generate a template and write it directly to `nft` stdin, then just execute `nft -f -`

pkg/provision/primitives/public_ip.go
```diff
@@ -26,11 +34,122 @@ func (p *Provisioner) publicIPProvision(ctx context.Context, reservation *provis
...
+nft 'add rule bridge filter %[1]s ip6 version 6 udp sport 547 drop'
+nft 'add rule bridge filter %[1]s ip version 4 udp sport 67 drop'`, fName, iface, ip, mac, ipv6))
+
+	if err := cmd.Run(); err != nil {
+		return errors.Wrap(err, "could not setup firewall rules for public ip")
+	}
+	return nil
+}
+
+func teardownFilters(ctx context.Context, fName string) error {
```
muhamadazmy, line 107: same comment as above

pkg/provision/primitives/public_ip.go
LeeSmet: Not sure. Keep in mind that we are not just setting a template, but actually applying rules to an existing nft setup. I personally don't know nft enough to know if this is possible

muhamadazmy (approved):

PR #1122, opened 2020-12-14
fix: continue when fetching device info fails

### LINKED ISSUE #1121: Failed to read smartctl information from disks error="exit status 2"
```

2020-12-14 09:36:10 | [+] capacityd: 2020-12-14T08:36:10Z fatal failed to read smartctl information from disks error="exit status 2"
-- | --
  |   | 2020-12-14 09:36:10 | [+] storaged: 2020-12-14T08:36:10Z debug pool usage device=/dev/sdb size=11999601754112
  |   | 2020-12-14 09:36:10 | [+] storaged: 2020-12-14T08:36:10Z debug pool usage device=/dev/sdd size=3000034656256
  |   | 2020-12-14 09:36:10 | [+] capacityd: 2020-12-14T08:36:10Z info read DMI info
  |   | 2020-12-14 09:36:10 | [+] capacityd: 2020-12-14T08:36:10Z info resource units found CRU=32 HRU=25145 MRU=126 SRU=0
  |   | 2020-12-14 09:36:10 | [+] storaged: 2020-12-14T08:36:10Z error Failed to mount pool ddedf812-4468-4ffa-ae3f-d0d2822225c7 error="invalid argument"
  |   | 2020-12-14 09:36:09 | [+] storaged: 2020-12-14T08:36:09Z debug pool usage device=/dev/sda size=11999601754112
  |   | 2020-12-14 09:36:09 | [+] capacityd: 2020-12-14T08:36:09Z info inspect hardware resources

```

on node: `2E9DRBNHaMXh2zswQsbxLppZi17r9w2qQfsSwsKEL7so`

DylanVerstraete:


LeeSmet (changes requested):
There is a logic problem here. At the start of the loop, you create a new **local** `brokenPools` slice. Then when a pools is found which is bad, you overwrite the existing `brokenPools` slice on the struct, with the appendage of the local version and the broken pool. This means the local slice will never have any elements. Furthermore, because of this, the struct lvl `brokenPools` slice will always only have the latest broken pool to be found. If multiple pools are bad, all but the last bad one will no longer be tracked after this function is executed (untill the cache is filled again, if ever). It seems there is actually no good reason to create a local slice here, you can just append the broken pool to the struct lvl slice, since that is ultimately what we are looking for anyway

muhamadazmy (approved):

muhamadazmy (approved):

PR #1125, opened 2020-12-16
add new k8s size

DylanVerstraete:


pkg/provision/primitives/kubernetes.go
```diff
@@ -509,6 +509,9 @@ func vmSize(size uint8) (cpu uint8, memory uint64, storage uint64, err error) {
 	case 17:
 		//4 vCpu, 8 GiB RAM, 50 GB disk
 		return 4, 8 * 1024, 50 * 1024, nil
+	case 18:
+		//1 vCpu, 1 GiB RAM, 25 GB disk
+		return 1, 1 * 1024, 25 * 1024, nil
 	}
 
 	return 0, 0, 0, fmt.Errorf("unsupported vm size %d, only size 1 and 2 are supported", size)
```
muhamadazmy: Please change error message because obviously we have more than 2 sizes :D

muhamadazmy (changes requested):
This size still need to be added here as well https://github.com/threefoldtech/zos/blob/master/pkg/provision/primitives/counters.go#L271

muhamadazmy (approved):

PR #1188, opened 2021-02-12
container: allow elevated privileged container for specific flist

maxux:
Allow zdbfs flist to be started with elevated privileges (SYS_ADMIN and /dev/fuse available).
The flist url will change in the futur.


pkg/provision/primitives/container.go
```diff
@@ -249,6 +253,13 @@ func (p *Provisioner) containerProvisionImpl(ctx context.Context, reservation *p
 		return ContainerResult{}, err
 	}
 
+	var elevated = false
+
+	if config.FList == FListElevated {
```
xmonader: I think it's a quite useful feature and can see it used more in the future for other things, I guess It might be a good idea to have the elevated information on the flist object itself on the hub, and based on its permissions it'd reflect here as elevated.

maxux: We don't want anybody to get theses elevated privileged, only trusted flist.
EDIT: after better understanding of the comment, this was approved. Matching is made on trusted repository and not single flist.

maxux:
Tested locally, elevated are well restricted to https://hub.grid.tf/tf-elevated repository. :)

muhamadazmy (approved):

PR #1201, opened 2021-02-25
Support ygg adderesses in containers

muhamadazmy:


pkg/network/networker.go
```diff
@@ -285,7 +286,7 @@ func (n networker) ZDBPrepare(hw net.HardwareAddr) (string, error) {
 
 	}
 
-	return netNSName, n.createMacVlan(ZDBIface, hw, ips, routes, netNs)
+	return netNSName, n.createMacVlan(ZDBIface, types.YggBridge, hw, ips, routes, netNs)
```
LeeSmet: The yggdrasil macvlan for zdbs should be installed next to the already existing macvlan, not replace it

LeeSmet (changes requested):

pkg/network/networker.go
muhamadazmy: Ah right, so zdb will has both as well. right. This means we need another interface in the zdb container ?

LeeSmet: indeed, everything with ygg IP will have a dedicated interface for that. Later when a vm can have an ygg IP, it will also create a dedicated tap iface on the ygg bridge next to the existing ones. That way traffic flow is nicely separated.

pkg/network/networker.go
```diff
@@ -240,7 +243,10 @@ func (n *networker) Leave(networkdID pkg.NetID, containerID string) error {
 
 // ZDBPrepare sends a macvlan interface into the
 // network namespace of a ZDB container
-func (n networker) ZDBPrepare(hw net.HardwareAddr) (string, error) {
+func (n networker) ZDBPrepare(id string) (string, error) {
+
+	hw := ifaceutil.HardwareAddrFromInputBytes([]byte("pub:" + id))
```
LeeSmet, line 249: I think the "pub:" prefix should not be here, so the input is the same as it was previously for this iface. Otherwise, if a node has a zdb running and reboots after this update, it will generate a new mac address, resulting in a different ip6 from slaac, which will mean the ipv6 originally returned in the result (which is probably used) is not valid anymore

LeeSmet (changes requested):

muhamadazmy:
I can drop the prefix, but we already are not backward compatible with
zos2. So this is all for fresh nodes

LeeSmet:
Hmm yes, if we only use it for grid3 that makes sense. Looks good, but CI errors

PR #1214, opened 2021-03-09
Set the correct container cpu limit

### LINKED ISSUE #1212: CPU Core resource allocation is off
Did a test with a number of different sized container as we are working with a partner who would like to test the performance of a container with low memory. All containers are provisioned with the "Ubuntu" solution in the online 3bot on testnet.

1. container with 1 core and 1GB of RAM
2. container with 2 cores and 1GB of RAM
3. container with 4 cores and 1GB of RAM

the results are things that are off:
- when I provision a single core I can max out all 24 cores (100% utilisation) (all 24 corews)
- when I provision multiple cores I can max out all 24 cores to (1/number_of_cores). 2cores -> max out all @ 50%. 4 cores -> max out all cores @ 25%
- when I provision more cores I can still use all cores, but I get a lower maximum percentage of all cores 

I guess desired behavior is:
- when provisioned a single core - you get 100% of a single core
- when you provision multiple cores you get access to multiple cores @ 100 % or get access to multiple of the provisioned cores at less than 100% max usage totaling to number of provisioned cores * 100%.

Partner is going to test (for a blockchain solution, proof of work type) how much performance they get renting cores.

- In order to test I have installed sysbench (parallel thread stress testing tools).
- Each container is installed on a physically different server
- pushing to launch 24 threads on each of these containers.

Result single core provisioning when the command is ```sysbench cpu --threads=24 run```
screenshot_20210309_081216

24 cores * 100% utlilisation = 24 times what I paid for.

Result double core provisioning when the command is ```sysbench cpu --threads=24 run```
screenshot_20210309_081342
24 cores * 50%  / 2 cores = 1200 / 2 = 6 times what I paid for

Result four  core provisioning when the command is ```sysbench cpu --threads=24 run```
screenshot_20210309_081501

24 cores * 25% utilization / 4  = 600% / 4 = 1.5 time what I paid for

DylanVerstraete:
This PR fixes the issue where containers were using all the cpu they could find on the host. 

It has the following implications:

- load will be divided over all the cpus on the host, but cannot exceed the total reserved cpu for that container.
- monitoring a container cpu will be harder because the load is divided over all cpus

### DIFF
--- a/pkg/container/opts.go
+++ b/pkg/container/opts.go
@@ -107,7 +107,7 @@ func WithCPUCount(cru uint) oci.SpecOpts {
 			return fmt.Errorf("asked %d CRU while only %d are available", cru, totalCPU)
 		}
 
-		quota, period := cruToLimit(cru, totalCPU)
+		quota, period := cruToLimit(cru)
 
 		if s.Linux.Resources == nil {
 			s.Linux.Resources = &specs.LinuxResources{}
@@ -123,8 +123,9 @@ func WithCPUCount(cru uint) oci.SpecOpts {
 	}
 }
 
-func cruToLimit(cru uint, totalCPU int) (quota int64, period uint64) {
-	quota = 100000
-	period = uint64(quota) * uint64(cru)
+func cruToLimit(cru uint) (quota int64, period uint64) {
+	period = 100000
+	quota = int64(period * uint64(cru))
+
 	return quota, period
 }
--- a/pkg/container/opts.go
+++ b/pkg/container/opts.go
@@ -3,7 +3,6 @@ package container
 import (
 	"context"
 	"fmt"
-	"math"
 
 	"path"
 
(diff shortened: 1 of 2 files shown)


LeeSmet (approved):

pkg/container/opts.go
```diff
@@ -124,15 +123,9 @@ func WithCPUCount(cru uint) oci.SpecOpts {
...
-		total    = float64(totalCPU)
-		p        float64
-	)
-	quota = int64(1000000) // 1 sec
-	p = required / total
-	p *= float64(quota)
-	p = math.Ceil(p)
-	return quota, uint64(p)
+func cruToLimit(cru uint) (quota int64, period uint64) {
+	period = 100000
```
muhamadazmy:
```suggestion
	const period = 100000
```

PR #1223, opened 2021-03-22
use macvtap instead of tap for public Ips

muhamadazmy:


pkg/network/macvtap/tap.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	var ok bool
+	tap, ok = link.(*netlink.Macvtap)
+	if !ok {
+		// right now, the netlink lib returns a `*GenericLink` for the tap interface,
+		// so  assign properties to a blank tap
+		gl, ok := link.(*netlink.GenericLink)
+		if !ok {
+			return nil, fmt.Errorf("link %s should be of type tuntap", name)
```
LeeSmet:
```suggestion
			return nil, fmt.Errorf("link %s should be of type macvtap", name)
```

pkg/network/networker.go
```diff
@@ -501,6 +503,7 @@ func (n *networker) DisconnectPubTap(pubIPReservationID string) error {
 		return errors.Wrap(err, "could not load tap device")
 	}
 
+	// TODO: this will not work with a macvtap,
```
LeeSmet: I assume this is fixed?

pkg/provision/primitives/public_ip.go
```diff
@@ -66,7 +66,7 @@ func (p *Provisioner) publicIPProvisionImpl(ctx context.Context, reservation *pr
 
 	tapName := fmt.Sprintf("p-%s", reservation.ID) // TODO: clean this up, needs to come form networkd
 	fName := filterName(reservation.ID)
-	mac := ifaceutil.HardwareAddrFromInputBytes(config.IP.IP.To4())
+	mac := ifaceutil.HardwareAddrFromInputBytes([]byte(reservation.ID))
```
LeeSmet, line 69: whats the reason for this change?

LeeSmet (changes requested):

pkg/network/networker.go
muhamadazmy: No, actually it's not fixed, because you can't unset the master of a macvlan. If u try that with command line it does not fail but doesn't change anything.
I am not sure what we can do here instead to disconnect the device. I thought may be change the txqueue back to 0

pkg/provision/primitives/public_ip.go
muhamadazmy: The reason for this is that the mac need to be set also on the macvtap device on host. networkd only take the reservation ID to create the macvtap device. so instead of changing the zbus interface instead I changed the mac to use the reservation ID instead. this way 
- the macvtap has the correct mac
- the CH is passed the correct mac
- the vm gets (of course) the same mac
if the mac on the host does not match the one in the vm it will not work

pkg/network/networker.go
LeeSmet: right, I think honestly this is by design since the macvtap is essentially looks like a macvlan in the upper part. This function should just cut the network connection between the 2 ends of the interface somehow so a vm no longer has network access after calling this

LeeSmet (approved):

PR #1227, opened 2021-03-26
zdb: update flist url to latest version in release mode

maxux:
Update zdb flist, contains lot of fixes and features, including `fd` leaks.
Switch to development to `release` mode, which will improve performance a lot, we don't really monitor logs anyway.


LeeSmet (approved):
:ship:

Issue #1189, opened 2021-02-15
ZOS - vm: add a virtual machine primitives with a list of predefined images


muhamadazmy:
Create a VirtualMachine primitives that has a support of a predefined list of images.

PR #1261, opened 2021-04-29
Generic vms

OmarElawady:
Changes:

- Add generic VMs workload
- Add validation for ssh keys and plain secrets in k8s (if the plain secret contains white spaces or new lines, part of it is ignored or it might crash the VM)
- Pull out some of the common code of vms and k8s in vmutils

Notes:

Not tested on public ips yet.

Related issues:
https://github.com/threefoldtech/zos/issues/1189


pkg/provision/primitives/converter.go
```diff
@@ -208,6 +208,25 @@ func K8SToProvisionType(w workloads.Workloader) (Kubernetes, string, error) {
 	return k8s, k.NodeId, nil
 }
 
+// K8SToProvisionType converts type to internal provision type
```
xmonader: VMToProvisionType

pkg/provision/primitives/kubernetes.go
```diff
@@ -229,7 +229,11 @@ func (p *Provisioner) kubernetesInstall(ctx context.Context, name string, cpu ui
 		cmdline = fmt.Sprintf("%s k3os.server_url=https://%s:6443", cmdline, ipstring)
 	}
 	for _, key := range cfg.SSHKeys {
-		cmdline = fmt.Sprintf("%s ssh_authorized_keys=\"%s\"", cmdline, key)
+		trimmed := strings.TrimSpace(key)
+		if strings.ContainsAny(trimmed, "\"\n") {
```
xmonader, line 217: can't I add multiple sshkeys in the same step to the VM? giving multiple people access in the same time

pkg/provision/primitives/vm.go
```diff
@@ -0,0 +1,277 @@
...
+	Name string `json:"name"`
+}
+
+type VMInfo struct {
+	Initrd    string
+	Kernel    string
+	ImagePath string
+}
+
+const VMREPO = "https://hub.grid.tf/omar0.3bot/"
```
xmonader: let's update to `tf-official-vms`

pkg/provision/primitives/vmutils.go
```diff
@@ -0,0 +1,264 @@
...
+		return 2, 4 * 1024, 50 * 1024, nil
+	case 17:
+		//4 vCpu, 8 GiB RAM, 50 GB disk
+		return 4, 8 * 1024, 50 * 1024, nil
+	case 18:
+		//1 vCpu, 1 GiB RAM, 25 GB disk
+		return 1, 1 * 1024, 25 * 1024, nil
+	}
+
+	return 0, 0, 0, fmt.Errorf("unsupported vm size %d, only size 1 to 18 are supported", vm.GetSize())
```
xmonader: -1, and 1 to 18 are supported

pkg/provision/primitives/kubernetes.go
OmarElawady: ssh keys are passed as a list so you can add multiple keys by providing them as items. Making them one item separated by new line will fail because of how it's parsed inside the vm. This is an [example](https://github.com/emmericp/initramfs-tools/blob/14a89aba6bd21f6e8af8177e2343a3b7a9bb804b/init#L75), so a new line would make the ssh key separated between two steps.

pkg/provision/primitives/vmutils.go
OmarElawady: -1 is a flag that a custom size is provided not a predefined size. So I feel it might confuse the user (that he can provide -1 and everything will work as if he provided any other number). Do you think it should be mentioned in the message?

xmonader: I'm actually not sure, it's a valid value for the size, but probably @muhamadazmy can decide on this

pkg/provision/primitives/counters.go
```diff
@@ -156,6 +161,9 @@ func (c *Counters) Decrement(r *provision.Reservation) error {
 		c.zdbs.Decrement(1)
 		u, err = processZdb(r)
 	case KubernetesReservation:
+		c.k8s.Decrement(1)
+		u, err = processKubernetes(r)
+	case VirtualMachineReservation:
 		c.vms.Decrement(1)
 		u, err = processKubernetes(r)
```
muhamadazmy: This should be `processVM` not `processKubernetes`

pkg/provision/primitives/counters.go
```diff
@@ -203,6 +211,12 @@ func (c *Counters) CheckMemoryRequirements(r *provision.Reservation, totalMemAva
 		if err != nil {
 			return err
 		}
+
+	case VirtualMachineReservation:
+		requestedUnits, err = processKubernetes(r)
```
muhamadazmy: This is probably should be processVM also, right ?

pkg/provision/primitives/kubernetes.go
```diff
@@ -124,7 +122,9 @@ func (p *Provisioner) kubernetesProvisionImpl(ctx context.Context, reservation *
 	if err != nil {
 		return result, errors.Wrap(err, "failed to decrypt namespace password")
 	}
-
+	if strings.ContainsAny(config.PlainClusterSecret, " \t\r\n\f") {
```
muhamadazmy: I will move all validatation to `config.Validate()` method that is called the moment the config is unmarshalled

pkg/provision/primitives/kubernetes.go
muhamadazmy, line 217: Validation of user input should have been done by a `config.Validate()` method instead.

pkg/provision/primitives/kubernetes.go
```diff
@@ -189,7 +189,7 @@ func (p *Provisioner) kubernetesProvisionImpl(ctx context.Context, reservation *
 	}
 
 	var netInfo pkg.VMNetworkInfo
-	netInfo, err = p.buildNetworkInfo(ctx, reservation.Version, reservation.User, iface, pubIface, config)
+	netInfo, err = p.buildNetworkInfo(ctx, reservation.Version, reservation.User, iface, pubIface, config.IP, config.PublicIP, config.NetworkID)
```
muhamadazmy: Why expansion of arguments list?

pkg/provision/primitives/order.go
```diff
@@ -31,5 +35,6 @@ var ProvisionOrder = map[provision.ReservationType]int{
 	VolumeReservation:          4,
 	ContainerReservation:       5,
 	KubernetesReservation:      6,
-	PublicIPReservation:        7,
+	VirtualMachineReservation:  7,
+	PublicIPReservation:        8,
```
muhamadazmy: I really think public-ip should be before both VM and K8s actually, this might have been a bug.

pkg/provision/primitives/vm.go
```diff
@@ -0,0 +1,283 @@
...
+
+	// check if the network tap already exists
+	// if it does, it's most likely that a vm with the same network id and node id already exists
+	// this will cause the reservation to fail
+	exists, err := network.TapExists(netID)
+	if err != nil {
+		return result, errors.Wrap(err, "could not check if tap device exists")
+	}
+
+	if exists {
```
muhamadazmy, line 91: For now this is fine, it may be made sense with K8s to avoid having multiple clusters on the same machine. moving on to version 0.5.0 we may be should allow multiple VMs to run on the same node with the same network.

pkg/provision/primitives/vm.go
```diff
@@ -0,0 +1,283 @@
...
+		// attempt to delete the vm, should the process still be lingering
+		vm.Delete(reservation.ID)
+	}
+
+	return result, err
+}
+
+func (p *Provisioner) prepareVMFS(ctx context.Context, imageInfo VMInfo, diskPath string) error {
+
+	var cmd *exec.Cmd
```
muhamadazmy: This section can be done programatically by first opening the files (with correct flags, including RW mode and NO truncate) then use `io.Copy` to copy the image.

muhamadazmy: Plus calling dd with NO `bs` will make it copy data byte by byte which is very slow.

pkg/provision/primitives/vm.go
```diff
@@ -0,0 +1,283 @@
...
+	if err := cmd.Run(); err != nil {
+		return errors.Wrap(err, "failed to copy the ubuntu raw image over the disk")
+	}
+	dname, err := ioutil.TempDir("", "btrfs-resize")
+	if err != nil {
+		return errors.Wrap(err, "couldn't create a temp dir to mount the btrfs fs to resize it")
+	}
+	defer os.RemoveAll(dname)
+
+	cmd = exec.CommandContext(ctx, "mount", diskPath, dname)
```
muhamadazmy: No need to execute command, instead use the `syscall.Mount` function.

pkg/provision/primitives/vm.go
```diff
@@ -0,0 +1,283 @@
...
+
+	if err := cmd.Run(); err != nil {
+		return errors.Wrap(err, "couldn't mount the created disk to a tmp dir")
+	}
+
+	defer func() {
+		cmd = exec.CommandContext(ctx, "umount", dname)
```
muhamadazmy: Same for unmount

pkg/provision/primitives/vm.go
```diff
@@ -0,0 +1,283 @@
...
+	err = p.vmRun(ctx, reservation.ID, cpu, memory, diskPath, imageInfo, cmdline, netInfo)
+	if err != nil {
+		// attempt to delete the vm, should the process still be lingering
+		vm.Delete(reservation.ID)
+	}
+
+	return result, err
+}
+
+func (p *Provisioner) prepareVMFS(ctx context.Context, imageInfo VMInfo, diskPath string) error {
```
muhamadazmy: I really think this entire process should be handled by  `storaged` daemon. In the call to Allocate()  for the Vm it can accept an optional (source image path)
If provided storaged can then make sure to copy the image correctly.
Reasos:
- Storaged already have tons of utilities to work with a btrfs disk
- The logic here has an issue the way u call prepareVMS since u call it everytime the provisiond daemon tries to run this reservations. this happens on node reboots. which means each time the node restarts (or the vm is reprovisioned for any other reason) the disk will get completely wiped out and reset to initial state. but it should only be called once
- this logic still can be fixed here in provision, but will be eaiser to implement in storaged since u already only call Allocate on first creation

pkg/provision/primitives/converter.go
```diff
@@ -208,6 +208,25 @@ func K8SToProvisionType(w workloads.Workloader) (Kubernetes, string, error) {
...
 }
 
+// VMToProvisionType converts type to internal provision type
+func VMToProvisionType(w workloads.Workloader) (VM, string, error) {
+	k, ok := w.(*workloads.VirtualMachine)
+	if !ok {
+		return VM{}, "", fmt.Errorf("failed to convert kubernetes workload, wrong format")
+	}
+
+	vm := VM{
```
muhamadazmy, line 220: Since this is so similar to K8s strcut i suggest that you change Kubernetes struct to be like
```go
type Kubernetes struct{
   VM
   ...
    // extra k8s specific properties
}
```

pkg/provision/primitives/vmutils.go
```diff
@@ -0,0 +1,264 @@
...
+	SRU float64 `json:"sru"`
+}
+
+func (p *Provisioner) vmDecomission(ctx context.Context, reservation *provision.Reservation) error {
+	var (
+		storage = stubs.NewVDiskModuleStub(p.zbus)
+		network = stubs.NewNetworkerStub(p.zbus)
+		vm      = stubs.NewVMModuleStub(p.zbus)
+
+		cfg Kubernetes
```
muhamadazmy: should be `cfg VM` since that is the base for both structs.

pkg/provision/primitives/vmutils.go
muhamadazmy: The custom size is filled up by the explorer (user can't provide that) so if a user asked for size `-1` the explorer will fill the custom size in the reservation to max that this node can support

muhamadazmy (changes requested):
First, thank you for diving in layers of code that is trying to be backward compatible with older version (luckily we break this in version 0.5.0). Please check my comments, also take a look at why the CI is not happy.

muhamadazmy:
Can we also implement this for version `0.5.0` things are much cleaner on this version. will be easier to implement. this version is available on branch `master-3`. i can explain the differences

pkg/provision/primitives/vm.go
OmarElawady: I couldn't use the mount syscall since I'm mounting a btrfs from a file not a device. The mount command takes care of this.

pkg/provision/primitives/kubernetes.go
OmarElawady: Since they were separate types, didn't know a better way to make it handle both. Removed the expansion after embedding the VM in K8s.

pkg/provision/primitives/kubernetes.go
OmarElawady: Moved all the validation to Validate function.

pkg/storage/disk.go
```diff
@@ -101,6 +118,28 @@ func (d *vdiskModule) safePath(base, id string) (string, error) {
...
 }
 
+func (d *vdiskModule) expandfs(disk string) error {
+	dname, err := ioutil.TempDir("", "btrfs-resize")
+	if err != nil {
+		return errors.Wrap(err, "couldn't create a temp dir to mount the btrfs fs to resize it")
+	}
+	defer os.RemoveAll(dname)
+
+	cmd := exec.Command("mount", disk, dname)
```
muhamadazmy, line 128: please use the `syscall.Mount` command.

muhamadazmy (approved):

PR #1276, opened 2021-05-19
Add generic vms

OmarElawady:
The tap device created for the VM is deduced from the workload id to allow multiple VMs on the same node on the same network.


pkg/gridtypes/zos/kubernetes.go
```diff
@@ -104,46 +103,17 @@ var k8sSize = map[uint8]gridtypes.Capacity{
...
-
-	if wl.Type != PublicIPType {
-		return errors.Wrapf(err, "workload of name '%s' is not a public ip", k.PublicIP)
-	}
-
-	return nil
-}
-
 // Challenge implementation
 func (k Kubernetes) Challenge(b io.Writer) error {
```
muhamadazmy, line 102: This can be done as 
```go
func (k Kubernetes) Challenge(b io.Writer) error {
   k.VirtualMachine.Challenge(b)
   // add the "extra" fields to the challenge like the ClusterSecret and master ips
}
```

cmds/modules/provisiond/swagger/zos-api.yml
```diff
@@ -468,6 +470,28 @@ components:
           type: string
         disable_default_ingress:
           type: boolean
+    VirtualMachine:
```
muhamadazmy, line 494: in swagger, you can also do some sort of "inheritance" so a K8s struct can include the attrs from virtualmachine. look it up

pkg/gridtypes/zos/vm.go
```diff
@@ -0,0 +1,98 @@
...
+	Name     string   `json:"name"`
+	Network  string   `json:"network"`
+	IP       net.IP   `json:"ip"`
+	SSHKeys  []string `json:"ssh_keys"`
+	PublicIP string   `json:"public_ip"`
+	Size     uint8    `json:"size"`
+}
+
+// Valid implementation
+func (v VirtualMachine) Valid(getter gridtypes.WorkloadGetter) error {
```
muhamadazmy, line 92: looks good, but I don't see that u actually use this in the Kubernetes struct. the Valid from that one can then use the VirtualMachine.Valid() then validates the extra k8s specific attributes

pkg/gridtypes/zos/vm.go
```diff
@@ -0,0 +1,98 @@
...
+
+	rsu, ok := vmSize[v.Size]
+	if !ok {
+		return gridtypes.Capacity{}, fmt.Errorf("VM size %d is not supported", v.Size)
+	}
+	return rsu, nil
+}
+
+// Challenge creates signature challenge
+func (v VirtualMachine) Challenge(b io.Writer) error {
```
muhamadazmy, line 149: Also Challenge should be reused int he K8s struct

pkg/gridtypes/zos/vm.go
```diff
@@ -0,0 +1,98 @@
...
+
+	if wl.Type != PublicIPType {
+		return errors.Wrapf(err, "workload of name '%s' is not a public ip", v.PublicIP)
+	}
+
+	return nil
+}
+
+// Capacity implementation
+func (v VirtualMachine) Capacity() (gridtypes.Capacity, error) {
```
muhamadazmy, line 132: Since Capacity is exactly the same for K8s and Vm, the implementation from K8s should be deleted. the way Composition (inheritance) works in Go will make this method available for k8s

pkg/stubs/network_stub.go
```diff
@@ -328,7 +329,7 @@ func (s *NetworkerStub) RemovePubTap(ctx context.Context, arg0 string) (ret0 err
 	return
 }
 
-func (s *NetworkerStub) RemoveTap(ctx context.Context, arg0 zos.NetID) (ret0 error) {
+func (s *NetworkerStub) RemoveTap(ctx context.Context, arg0 string) (ret0 error) {
```
muhamadazmy, line 332: By the way, we have a tool to generate stubs to match the interface from pkg/. Do u change this manually ?

muhamadazmy (changes requested):

pkg/stubs/network_stub.go
OmarElawady: yes, got a lot of panics because of type mismatches. where can i find it?

pkg/gridtypes/zos/vm.go
OmarElawady: It's only implemented in vm?

pkg/stubs/network_stub.go
muhamadazmy: Install it like this 
```
go install github.com/threefoldtech/zbus/zbusc
```
make sure now `zbusc` in your path. 
then do 
```
cd zos/pkg
go generate ./...
```

The generate will use the zbus tool to generate the stubs.

muhamadazmy:
@sasha-astiadi please, this PR does not belong to 2.7 or 8 so don't add it to that project!
This belongs to tfgrid 3, the generic VM for 2.7 has already been merged. Please next time check with me if possible

muhamadazmy:
Closing this PR because some work still need to be done here. will be used as a base for ZMACHINE implementation

PR #1285, opened 2021-06-01
Protect zdb host process with volume id

DylanVerstraete:


pkg/provision/cleanup.go
```diff
@@ -389,6 +389,6 @@ func socketDir(containerID string) string {
 
 func newZdbConnection(id string) (zdb.Client, error) {
 	socket := fmt.Sprintf("unix://%s/zdb.sock", socketDir(id))
-	cl := zdb.New(socket)
+	cl := zdb.New(string(id), socket)
```
xmonader: id is already string here

pkg/zdb/zdb.go
```diff
@@ -98,7 +100,16 @@ func newRedisPool(address string) (*redis.Pool, error) {
 
 	return &redis.Pool{
 		Dial: func() (redis.Conn, error) {
-			return redis.Dial(u.Scheme, host, opts...)
+			con, err := redis.Dial(u.Scheme, host, opts...)
+			if err != nil {
+				return nil, err
+			}
+			_, err = con.Do("AUTH", password)
```
muhamadazmy, line 107: This is correct, but now there is an issue regarding "backward compatibility" since on a life node some zdbs will be running already with no protection. so if u fail on AUTH, you need to try to see if authentication is require or not. if not required then u return with no error, but if require u return error 

basically
```
try auth() {
  if err // failed to authorize {
    // check if auth is required (try to do a ping for example)
   if no error {
      //success
   else {
    // possibly other issue or bad passowrd, return error
   }
  }
}
```

muhamadazmy (changes requested):
You need to do some tests as follows:
- backward compatibility with a node that already running zdb
- zdbs after being restarted are running with passwords

muhamadazmy (approved):

PR #1292, opened 2021-06-04
Feat/provisiond message bus

DylanVerstraete:


pkg/mbus/mbus.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/rs/zerolog/log"
+)
+
+const (
+	systemLocalBus = "msgbus.system.local"
+	replyBus       = "msgbus.system.reply"
+)
+
+type Message struct {
+	Version    int    `json:"version"`
```
muhamadazmy: the json names **MUST** match the ones from the actual msgbus protocol as defined in the repo

pkg/mbus/mbus.go
```diff
@@ -0,0 +1,168 @@
...
+	Id         string `json:"id"`
+	Command    string `json:"command"`
+	Expiration int    `json:"expiration"`
+	Retry      int    `json:"retry"`
+	Data       string `json:"data"`
+	Twin_src   []int  `json:"twin_src"`
```
muhamadazmy: Already the CI is complaining that this name is not a conventional name, it should be TwinSrc instead (same for others)

pkg/mbus/mbus.go
```diff
@@ -0,0 +1,168 @@
...
+	Twin_dest  []int  `json:"twin_dest"`
+	Retqueue   string `json:"retqueue"`
+	Schema     string `json:"schema"`
+	Epoch      int64  `json:"epoch"`
+	Err        string `json:"err"`
+}
+
+type Messagebus struct {
```
muhamadazmy:
```suggestion
type MessageBus struct {
```

pkg/mbus/mbus.go
```diff
@@ -0,0 +1,168 @@
...
+	Context context.Context
+	pool    *redis.Pool
+}
+
+func New(port uint16, context context.Context) (*Messagebus, error) {
```
muhamadazmy: check Ci complaining about the arguments order, it's a convention to have `Context` as first argument

pkg/mbus/mbus.go
```diff
@@ -0,0 +1,168 @@
...
+	addr := fmt.Sprintf("tcp://[::]:%d", port)
```
muhamadazmy: Who said it's going to be tcp always? We probably gonna do it over unix socket, hence the New argument should actually accept a 'connection' address instead (uninx:/// or tcp://)

pkg/provision/mbus/mbus.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"context"
+
+	"github.com/rs/zerolog/log"
+	"github.com/threefoldtech/zos/pkg/mbus"
+	"github.com/threefoldtech/zos/pkg/provision"
+)
+
+const (
+	cmdDeploy = "msgbus.deploy"
```
muhamadazmy: Why the cmd has `msgbus` prefix ? i think this instead should be something like

`zos.deployment.deploy`
`zos.deployment.get`

`zos.network.config.blahblah` 
and so on.

pkg/provision/mbus/reservation.go
```diff
@@ -0,0 +1,69 @@
...
+	"time"
+
+	"github.com/threefoldtech/zos/pkg/gridtypes"
+	"github.com/threefoldtech/zos/pkg/mbus"
+	"github.com/threefoldtech/zos/pkg/provision"
+	"github.com/threefoldtech/zos/pkg/provision/mw"
+)
+
+func (a *WorkloadsMessagebus) CreateOrUpdate(ctx context.Context, message mbus.Message, create bool) (interface{}, mw.Response) {
+	bytes, err := base64.RawStdEncoding.DecodeString(message.Data)
```
muhamadazmy: This depends on the user of the library to actually know the encoding used in the message. hence it requires good knowledge of the underlying protocol. What happens if in a new version of the bus we decided to move to a binary message format ?

A better way to do this is something like
```
bytes, err := message.GetPayload() //or GetData
```
and that one takes care of decoding the message correctly

pkg/provision/mbus/mbus.go
```diff
@@ -0,0 +1,61 @@
...
+
+	api := &WorkloadsMessagebus{
+		engine: engine,
+		mbus:   messageBus,
+	}
+
+	return api, nil
+}
+
+func (w *WorkloadsMessagebus) Run() error {
```
muhamadazmy: I don't like many things about this method
- it (again) requires knowledge of the internal of the message bus, the "Message" structure, the streams, and so on.
- Also a call to StreamMessages is already a "blocking" call, it will work in an infinite loop to pull messages (which is fine) but what is not fine, that the 2nd call to StreamMessages will never happen because the first call will never return.
- Why do i have to define channels, and do select, and much much more things that can be optimized and abstracted by the library itself.
- For example, and HTTP library, doesn't require knowledge of how the headers or the http request parsing is happening, while u as a user can still "get" this information from the request object. 

In my opinion, the library should have the following Interface:
```go

bus := rmb.New(url) // url to redis endpoint

bus.Handle("zos.deployment.deploy", deployHandler)
bus.Handle("zos.deployment.get", getHandler)
bus.Handle("zos.deployment.delete", deleteHandler)

sub := bus.Subrouter("network")
sub.Handle("config.get") // this should be similar to `bus.Handle("network.config.get")` but is cleaner to define may be and allow later injection of middle ware for different parts

// finally
bus.Run()
```

a function then can be something like
```go
func getHandler(ctx context.Context, payload bytes[]) (interface{}, error) {
  // load your pay load to parse request params
  // do your things
 // if needed access extra messages fields from ctx, like
 message := rmb.GetMessage(ctx)
  // or others!
  return result, nil
}
```

muhamadazmy (changes requested):

pkg/rmb/rmb.go
```diff
@@ -0,0 +1,251 @@
...
+	message, ok := ctx.Value(messageKey{}).(Message)
+	if !ok {
+		return nil, errors.New("failed to load message from context")
+	}
+
+	return &message, nil
+}
+
+// SendReply send a reply to the message bus with some data
+func (m *MessageBus) SendReply(message Message, data interface{}) error {
```
muhamadazmy: This doesn't have to be exposed

pkg/rmb/rmb.go
```diff
@@ -0,0 +1,251 @@
...
+		log.Err(err).Msg("failed to push to reply messagebus")
+		return err
+	}
+
+	return nil
+}
+
+// PushMessage pushes a message to a topic
+// for testing purposes
+func (m *MessageBus) PushMessage(topic string, message Message) error {
```
muhamadazmy: Same here, this doesn't have to be public

pkg/rmb/rmb.go
```diff
@@ -0,0 +1,251 @@
...
+
+			err = m.SendReply(message, data)
+			if err != nil {
+				log.Err(err).Msg("err while sending reply")
+			}
+		}
+	}
+}
+
+func GetMessage(ctx context.Context) (*Message, error) {
```
muhamadazmy: This should panic instead of returning an error. because failure to get the message is a programming error, not a runtime error. A failure to get the message meas the MessageBus never injected the message on the context. hence panic is much better here. 

It will also make the handler code much easier

pkg/rmb/rmb.go
```diff
@@ -0,0 +1,251 @@
...
+				log.Err(err).Msg("err while parsing payload reply")
+			}
+
+			handler, ok := m.handlers[message.Command]
+			if !ok {
+				log.Warn().Msg("handler not found")
+			}
+
+			var tKey twinKeyID
+			ctx = context.WithValue(ctx, tKey, message.TwinSrc)
```
muhamadazmy: this should be 
`requestCtx := context.WithValue(ctx, twinKeyID{}, message.TwinSrc)`
the idea is to create a new context per request. now this is instead overriding the context passed to the function because this is in a loop. so u actually override the ctx each time

Hence the correct way to do it is to create a new context for each request which leaves the ctx passed to the function as it is. this main ctx will be used for cancellation later

pkg/rmb/rmb.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+// WithHandler adds a topic handler to the messagebus
+func (m *MessageBus) WithHandler(topic string, handler func(ctx context.Context, payload []byte) (interface{}, error)) {
+	m.handlers[topic] = handler
+}
+
+// Run runs listeners to the configured handlers
+// and will trigger the handlers in the case an event comes in
+func (m *MessageBus) Run() error {
```
muhamadazmy:
```suggestion
func (m *MessageBus) Run(ctx context.Context) error {
```

pkg/rmb/rmb.go
```diff
@@ -0,0 +1,251 @@
...
+	defer con.Close()
+
+	topics := make([]string, len(m.handlers))
+	for topic := range m.handlers {
+		topics = append(topics, topic)
+	}
+
+	jobs := make(chan Message, numWorkers)
+	for i := 1; i <= numWorkers; i++ {
+		go m.worker(context.Background(), jobs)
```
muhamadazmy:
```suggestion
		go m.worker(ctx, jobs)
```

pkg/rmb/rmb.go
```diff
@@ -0,0 +1,251 @@
...
+			continue
+		}
+
+		_, ok := m.handlers[string(data[0])]
+		if !ok {
+			log.Debug().Msg("handler not found")
+			continue
+		}
+
+		jobs <- message
```
muhamadazmy: while this should work but there is possible loss of messages that can happen if a message is popped from the queue but then this code here was stopped before the message get chance to be processed. I am not sure if this is something that we should handle ourselves or by the rmb itself! @maxux ?

muhamadazmy (changes requested):

pkg/rmb/rmb.go
DylanVerstraete: for testing purposes right now, I have a testproject which uses this method to push some messages to the queue. I will remove this method before we merge this PR

PR #1368, opened 2021-08-29
Hash public tap names before sending to network module

OmarElawady:
Fixes `could not get network namespace tap device name: tap name too long`,
- Hashing like for normal taps is done on the provisioning side.
- The prefixless original hashed tap names is storedin the VMIface to pass to DisconnectPubTap and RemoveTap.
- (Not tested) need help setting up v3 node

### DIFF
--- a/pkg/vm.go
+++ b/pkg/vm.go
@@ -35,6 +35,8 @@ type VMIface struct {
 	IP6DefaultGateway net.IP
 	// Private or public network
 	Public bool
+	// OriginalTapName without prefixes to help in decommissioning
+	OriginalTapName string
 }
 
 // VMNetworkInfo structure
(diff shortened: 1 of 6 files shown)


pkg/vm.go
```diff
@@ -35,6 +35,8 @@ type VMIface struct {
 	IP6DefaultGateway net.IP
 	// Private or public network
 	Public bool
+	// OriginalTapName without prefixes to help in decommissioning
+	OriginalTapName string
```
muhamadazmy: You can always drive the tap name from the IP workload name. this is redundant and should not be stored here. All what u have to do in the decomission is just recompute the tap name from the reservation id .

muhamadazmy: https://github.com/threefoldtech/zos/blob/master-3/pkg/primitives/vm.go#L312

muhamadazmy (changes requested):

PR #1370, opened 2021-09-01
Zos3/web gateway

OmarElawady:


pkg/gateway/gateway.go
```diff
@@ -0,0 +1,165 @@
...
+
+type Server struct {
+	Url string
+}
+
+func New(root string) pkg.Gateway {
+	configPath := filepath.Join(root, "proxy")
+	// where should service-restart/node-reboot recovery be handled?
+	if _, err := os.Stat(configPath); os.IsNotExist(err) {
+		err := os.MkdirAll(configPath, 0644)
```
muhamadazmy: MkdirAll is safe to call even if directory exists. the checks if it exists is a nice touch, but not needed.

pkg/gateway/gateway.go
```diff
@@ -0,0 +1,165 @@
...
+		if err != nil {
+			// ok to panic?
+			panic(errors.Wrap(err, "couldn't make gateway config dir"))
```
muhamadazmy: Nope, yo need to return the error. change the New(root string) to `New(root string) (pkg.Gateway, error)

pkg/gateway/gateway.go
```diff
@@ -0,0 +1,165 @@
...
+		if err != nil {
+			log.Error().Err(err).Msg("couldn't start traefik")
+		}
+	}
+	return g
+}
+func isTraefikStarted() bool {
+	z, err := zinit.New("")
+	if err != nil {
+		panic(errors.Wrap(err, "couldn't get zinit client"))
```
muhamadazmy: No panicing please, always return an error. it's always the responsibility of the most top layer to decide either to exit on error or not.

pkg/gateway/gateway.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	defer z.Close()
+	started := true
+	traefikStatus, err := z.Status(traefikService)
+	if err != nil {
+		started = false
+	}
+	log.Debug().Str("state", traefikStatus.State.String()).Msg("checking traefik state")
+	started = traefikStatus.State.String() == zinit.ServiceStateRunning
```
muhamadazmy:
```suggestion
	started = traefikStatus.State == zinit.ServiceStateRunning
```

pkg/primitives/gateway.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	// - Validation of ownership of the name (later)
+	// this must be done against substrate. Make sure that same user (twin) owns the
+	// name int he workload config
+
+	// - make necessary calls to gateway daemon.
+	// gateway := stubs.NewGatewayStub(p.zbus)
+	// gateway.SetNamedProxy(ctx context.Context, arg0 string, arg1 []string)
+	gateway := stubs.NewGatewayStub(p.zbus)
+	if err := gateway.SetNamedProxy(ctx, fqdn, backends); err != nil {
```
muhamadazmy: I think an extra paramter (unique global workload id is needed to be passed here) so it's (ctx, wlId, fqdn, backends). the point is your config "name" must be the unique id. this way on decomission you just need to delete this config.

pkg/primitives/gateway.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func (p *Primitives) wgDecommission(ctx context.Context, wl *gridtypes.WorkloadWithID) error {
+	deployment := provision.GetDeployment(ctx)
+	twinID := deployment.TwinID
+	var proxy zos.GatewayNameProxy
+	if err := json.Unmarshal(wl.Data, &proxy); err != nil {
+		return fmt.Errorf("failed to unmarshal gateway proxy from reservation: %w", err)
+	}
+	if err := validateNameContract(twinID, proxy.Name); err != nil {
```
muhamadazmy: If we use the wrokdload id as a "key" to the domain configuration, this validation will not be needed. you just call the DeleteNamedProxy(wlid)

this is why we always use workload id to name all resources related to a workload. it's unique and is validated to be owned by the user (only owner can do decommission)

pkg/gateway/gateway.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	if err := z.StartWait(time.Second*20, traefikService); err != nil {
+		return errors.Wrap(err, "waiting for trafik start timed out")
+	}
+	g.traefikStarted = true
+	return nil
+}
+
+func (g *gatewayModule) SetNamedProxy(fqdn string, backends []string) error {
+	if !namespace.Exists(types.PublicNamespace) {
```
muhamadazmy: I already added a method to do this check (but available only on my network changes branch) for now. create a method 
```
func (g *gatewayModule) isGatewaySupported()
```
that does this call. which then can be modified later
```

File: pkg/gateway/gateway.go [hunk 8]
```diff
@@ -0,0 +1,165 @@
...
+	log.Debug().Str("yaml_config", string(yamlString)).Msg("configuration file")
+	filename := filepath.Join(g.proxyConfigPath, fmt.Sprintf("%s.yaml", fqdn))
+	if err = os.WriteFile(filename, yamlString, 0644); err != nil {
+		return errors.Wrap(err, "couldn't open config file for writing")
+	}
+
+	return nil
+}
+func (g *gatewayModule) DeleteNamedProxy(fqdn string) error {
+	filename := filepath.Join(g.proxyConfigPath, fmt.Sprintf("%s.yaml", fqdn))
```
muhamadazmy: You need to pass the workload id to both Set and Delete method, and use this workload id as a granteed unique file name. also it is easy to reverse the config file to an actual deployment.

muhamadazmy (changes requested):

pkg/gateway/gateway.go
OmarElawady: the compiler complains with this form. There's `.Is` method in state.

PR #1374, opened 2021-09-05
Add TLS config

OmarElawady:


pkg/gateway/gateway.go
```diff
@@ -178,11 +188,19 @@ func (g *gatewayModule) SetNamedProxy(wlID string, prefix string, backends []str
...
+	httpsRoute := fmt.Sprintf("%s-https", wlID)
 	config := ProxyConfig{
 		Http: HTTPConfig{
 			Routers: map[string]Router{
-				wlID: {
+				httpsRoute: {
+					Rule:    rule,
+					Service: wlID,
+					Tls: TlsConfig{
+						CertResolver: CertResolver,
```
muhamadazmy: I assume u added this CerResolver only to force json to serialize this TlsConfig and not consider it as "empty". right ?

PR #1389, opened 2021-09-16
Add dns resolution and wildcard certificate generation

OmarElawady:
Note: traefik must be restarted manually because of the added environment variable

### DIFF
--- a/pkg/gateway/gateway.go
+++ b/pkg/gateway/gateway.go
@@ -30,7 +30,7 @@ const (
 	// letsencrypt email need to customizable by he farmer.
 	letsencryptEmail = "letsencrypt@threefold.tech"
 	// certResolver must match the one defined in static config
-	httpCertResolver = "httpresolver"
+	httpCertResolver = "resolver"
 	dnsCertResolver  = "dnsresolver"
 )
 
@@ -67,8 +67,8 @@ type TlsConfig struct {
 }
 
 type Domain struct {
-	Sans   []string `yaml:"sans,omitempty"`
-	Domain string   `yaml:"domain,omitempty"`
+	Sans []string `yaml:"sans,omitempty"`
+	Main string   `yaml:"main,omitempty"`
 }
 
 type Service struct {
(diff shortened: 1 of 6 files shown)


pkg/gateway/gateway.go
```diff
@@ -253,8 +272,10 @@ func (g *gatewayModule) verifyDomainDestination(ctx context.Context, cfg pkg.Pub
 }
 
 func (g *gatewayModule) startTraefik(z *zinit.Client) error {
+
 	cmd := fmt.Sprintf(
-		"ip netns exec public %s --configfile %s",
+		"EXEC_PATH=%s ip netns exec public %s --configfile %s",
```
muhamadazmy: zinit, has support for `env` config. please use this instead.

https://github.com/threefoldtech/zinit/tree/master/docs#service-configuration

if the go library doesn't have this int he InitService structure, please add it.

pkg/gateway/gateway.go
```diff
@@ -314,9 +344,17 @@ func (g *gatewayModule) SetFQDNProxy(wlID string, fqdn string, backends []string
...
+		Domains: []Domain{
+			{
+				Main: fqdn,
+			},
+		},
+	}
+	return g.setupRouting(wlID, fqdn, backends, gatewayTLSConfig, TLSPassthrough)
 }
-func (g *gatewayModule) setupRouting(wlID string, fqdn string, backends []string, TLSPassthrough bool) error {
+func (g *gatewayModule) setupRouting(wlID string, fqdn string, backends []string, TLSConfig TlsConfig, TLSPassthrough bool) error {
```
muhamadazmy: argument names always camel case with lower case initials

```suggestion
func (g *gatewayModule) setupRouting(wlID string, fqdn string, backends []string, tlsConfig TlsConfig, tlsPassthrough bool) error {
```

muhamadazmy (changes requested):
Good work man, you can see below few comments. 
But there is also one big concern. Who starts dnsmasq first time? 
Imagine this scenario:
- Node already have gateways deployed
- Node reboots for any reason
- On resume of operation we should expect dnsmasq to be already running

I suggest dnsmasq is auto started with traefik (and maintained) with zinit the same way we hook traefik into zinit on gateway start.

This way the bash script will only need to do `kill <pid`>` and it will auto starts by zinit for u.

make sure that dnsmasq (in the zinit exect) is started in the "foreground" it shouldn't daemonize otherwise zinit will not be able to monitor it.

muhamadazmy (approved):

PR #1393, opened 2021-09-22
Zos3/qsfs non network part

OmarElawady:
1. QSFSPrepare => QSFSDestroy => QSFSPrepare, failed because of `can't set hw addr: already in use`. Fixed by deleting the ygg link explicitly before removing the namespace.
2. Add `RootfsPropagation` as the default one sets the mounts to `slave` recursively.

### DIFF
--- a/pkg/qsfsd/qsfs.go
+++ b/pkg/qsfsd/qsfs.go
@@ -64,9 +64,11 @@ func (q *QSFS) Mount(wlID string, cfg zos.QuatumSafeFS) (string, error) {
 		Limit:    cfg.Cache,
 		Storage:  "zdb://hub.grid.tf:9900",
 	})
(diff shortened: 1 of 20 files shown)


pkg/qsfsd/qsfs.go
```diff
@@ -0,0 +1,168 @@
...
+		Mounts: []pkg.MountInfo{
+			{
+				Source: mountPath,
+				Target: "/mnt",
+			},
+		},
+		Elevated: true,
+		// the default is rslave which recursively sets all mountpoints to slave
+		// we don't care about the rootfs propagation, it just has to be non-recursive
+		RootFsPropagation: "slave",
```
muhamadazmy: Can u define some constants (and a type) for the possible values

muhamadazmy (approved):
there is only one trivial comment, then you can merge

PR #1405, opened 2021-09-30
Add vm support for qsfs mounting

OmarElawady:
- added waiting for qsfs readiness at the end of qsfs provisioning

### DIFF
--- a/pkg/gridtypes/zos/types.go
+++ b/pkg/gridtypes/zos/types.go
@@ -26,6 +26,7 @@ const (
 )
 
 func init() {
+	gridtypes.RegisterType(QuantumSafeFSType, QuatumSafeFS{})
 	gridtypes.RegisterType(ZMountType, ZMount{})
 	gridtypes.RegisterType(NetworkType, Network{})
 	gridtypes.RegisterType(ZDBType, ZDB{})
@@ -33,7 +34,6 @@ func init() {
 	gridtypes.RegisterType(PublicIPType, PublicIP{})
 	gridtypes.RegisterType(GatewayNameProxyType, GatewayNameProxy{})
 	gridtypes.RegisterType(GatewayFQDNProxyType, GatewayFQDNProxy{})
-	gridtypes.RegisterType(QuantumSafeFSType, QuatumSafeFS{})
 }
 
 // DeviceType is the actual type of hardware that the storage device runs on,
--- a/pkg/gridtypes/zos/types.go
+++ b/pkg/gridtypes/zos/types.go
@@ -26,7 +26,6 @@ const (
 )
 
 func init() {
-	gridtypes.RegisterType(QuantumSafeFSType, QuatumSafeFS{})
 	gridtypes.RegisterType(ZMountType, ZMount{})
 	gridtypes.RegisterType(NetworkType, Network{})
 	gridtypes.RegisterType(ZDBType, ZDB{})
@@ -34,6 +33,7 @@ func init() {
 	gridtypes.RegisterType(PublicIPType, PublicIP{})
 	gridtypes.RegisterType(GatewayNameProxyType, GatewayNameProxy{})
 	gridtypes.RegisterType(GatewayFQDNProxyType, GatewayFQDNProxy{})
+	gridtypes.RegisterType(QuantumSafeFSType, QuatumSafeFS{})
 }
 
 // DeviceType is the actual type of hardware that the storage device runs on,
--- a/pkg/primitives/vm.go
+++ b/pkg/primitives/vm.go
@@ -286,8 +286,7 @@ func (p *Primitives) virtualMachineProvisionImpl(ctx context.Context, wl *gridty
 			Type: pkg.BootDisk,
 			Path: info.Path,
 		}
-		config.Mounts = config.Mounts[1:]
-		if err := p.vmMounts(ctx, deployment, config.Mounts, false, &machine); err != nil {
+		if err := p.vmMounts(ctx, deployment, config.Mounts[1:], false, &machine); err != nil {
 			return result, err
 		}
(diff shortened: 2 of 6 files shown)


pkg/primitives/vm.go
```diff
@@ -20,7 +21,7 @@ import (
 
 const (
 	// this probably need a way to update. for now just hard code it
-	cloudContainerFlist = "https://hub.grid.tf/azmy.3bot/cloud-container.flist"
+	cloudContainerFlist = "https://hub.grid.tf/omar0.3bot/omarelawady-cloud-container-latest.flist"
```
muhamadazmy: we probably need to move this to official tf repos.

pkg/primitives/vm.go
```diff
@@ -52,7 +53,7 @@ func (p *Primitives) mountsToDisks(ctx context.Context, deployment gridtypes.Dep
 			return nil, errors.Wrapf(err, "failed to get disk '%s' workload", disk.Name)
 		}
 		if wl.Type != zos.ZMountType {
-			return nil, fmt.Errorf("expecting a reservation of type '%s' for disk '%s'", zos.ZMountType, disk.Name)
+			continue
```
muhamadazmy: I would prefer having one method (vmMounts) for example, that goes over zos.MachineMount, then based ont he type can take a different action (including validation)

Now the problem is with this here is that if you added a wrong name (another type) the problem will go un-noticed.

so instead
```
for mnt in vm.Mounts {
   obj := dl.Get(mnt.name)
   switch obj.Type {
   disk: // handle disk
   qsfs: // handle qsfs
   default: <that is an error>
}
```

the method can have a pointer to the Vmd VM object to fill in proper config

pkg/qsfsd/qsfs.go
```diff
@@ -100,8 +103,50 @@ func (q *QSFS) Mount(wlID string, cfg zos.QuatumSafeFS) (mountPath string, err e
...
+		Filesystems []struct {
+			Fstype string `json:"fstype"`
+		} `json:"filesystems"`
+	}
+
+	if err := json.Unmarshal(output, &result); err != nil {
+		return false, errors.Wrap(err, "failed to parse findmnt output")
+	}
+	l := len(result.Filesystems)
+	if result.Filesystems[l-1].Fstype != "fuse.zdbfs" {
```
muhamadazmy: Shouldn't this list always be of len == 1 ?

pkg/vm.go
```diff
@@ -136,6 +147,8 @@ type VM struct {
 	// Disks are a list of disks that are going to
 	// be auto allocated on the provided storage path
 	Disks []VMDisk
+	// Qsfs are a list of qsfs that are going to
+	Qsfs []Qsfs
```
muhamadazmy: VMD shouldn't be aware of `qsfs` at all. for it, it's just a shared directory from the host so it should be handled and named like this, a better name should be `Shared` or even `Host`

pkg/primitives/vm.go
OmarElawady: agree, this flist is built from this pr https://github.com/threefoldtech/cloud-container/pull/1, if it's merged we can use whatever flist is generated from there.

pkg/qsfsd/qsfs.go
OmarElawady: it's mounted twice, one bind mount for the child mounts to reflect on the host fs, and the second is the zdbfs mount. https://github.com/threefoldtech/zos/blob/07762a455f2a9a41b9cd8730463698312a9f22eb/pkg/qsfsd/qsfs.go#L146

pkg/primitives/vm.go
```diff
@@ -259,18 +286,21 @@ func (p *Primitives) virtualMachineProvisionImpl(ctx context.Context, wl *gridty
 			Type: pkg.BootDisk,
 			Path: info.Path,
 		}
-		// we don't format disks attached to VMs, it's up to the vm to decide that
-		disks, err = p.mountsToDisks(ctx, deployment, config.Mounts[1:], false)
-		if err != nil {
+		config.Mounts = config.Mounts[1:]
```
muhamadazmy: I don't get why this is done?

pkg/gridtypes/zos/types.go
```diff
@@ -26,14 +26,14 @@ const (
 )
 
 func init() {
+	gridtypes.RegisterType(QuantumSafeFSType, QuatumSafeFS{})
```
muhamadazmy: Order doesn't matter here as explained

pkg/qsfsd/qsfs.go
muhamadazmy: And are u sure the qsfs mount will always be the last one to show in the list? can u loop over the list of the filesystems and find the right one instead?

pkg/primitives/vm.go
OmarElawady: Don't know actually, it was just to not change the already-existing code semantics. But I think the first mount is meant to be used as the rootfs(?) mount so it's processed for boot.

pkg/qsfsd/qsfs.go
```diff
@@ -100,8 +103,51 @@ func (q *QSFS) Mount(wlID string, cfg zos.QuatumSafeFS) (mountPath string, err e
...
 		cont,
 	)
+	if lerr := q.waitUntilMounted(mountPath, 2*time.Minute); lerr != nil {
+		err = lerr
+		return
+	}
 	return
 }
+
+func (f *QSFS) waitUntilMounted(path string, timeout time.Duration) error {
```
muhamadazmy, line 113: Why this is done here not in qsfsd itself ?

muhamadazmy (approved):

PR #1406, opened 2021-10-04
Add name validation for gateways

OmarElawady:


cmds/modules/provisiond/reporter.go
```diff
@@ -194,12 +232,21 @@ func (r *Reporter) synchronize(ctx context.Context, reported []Consumption) erro
...
 
-	for _, contract := range contracts {
+	for _, contract := range nodeContracts {
+		deployment, err := r.engine.Storage().Get(uint32(contract.TwinID), uint64(contract.ContractID))
+		if err != nil && !errors.Is(err, provision.ErrDeploymentNotExists) {
+			log.Error().
+				Err(err).
+				Msgf("failed to get deployment %d for name contract validation", contract.ContractID)
+		} else if err == nil && r.containsExpiredNameWorkloads(deployment) {
+			// delete it
```
muhamadazmy: you say delete it but u actually continue.

cmds/modules/provisiond/reporter.go
```diff
@@ -194,12 +232,21 @@ func (r *Reporter) synchronize(ctx context.Context, reported []Consumption) erro
...
 	}
 
-	for _, contract := range contracts {
+	for _, contract := range nodeContracts {
+		deployment, err := r.engine.Storage().Get(uint32(contract.TwinID), uint64(contract.ContractID))
+		if err != nil && !errors.Is(err, provision.ErrDeploymentNotExists) {
+			log.Error().
+				Err(err).
+				Msgf("failed to get deployment %d for name contract validation", contract.ContractID)
+		} else if err == nil && r.containsExpiredNameWorkloads(deployment) {
```
muhamadazmy: This code (if it works) will decomission the entire deployment even if one name contract is invalid. May be it need to be re-written so you can decomission a single workload (the invalid name one). the engire already has a method to do so.

pkg/gateway/gateway.go
```diff
@@ -207,10 +227,101 @@ func New(ctx context.Context, cl zbus.Client, root string) (pkg.Gateway, error)
...
+				Err(err).
+				Msgf("failed to parse wlID %s parts", id)
+			continue
+		}
+		if !strings.HasSuffix(domain, baseDomain) {
+			// a fqdn workload, skip validating it
+			continue
+		}
+		name := strings.TrimSuffix(domain, fmt.Sprintf(".%s", baseDomain))
+		if err := g.validateNameContract(name, twinID); err != nil {
```
muhamadazmy: You try to decomission a name reservation on any error. IMHO validateNameContract should return specific set of errors (define some Err<Something>) where u can check. the decomission should happen always if the name contract does not exist or have been canceled. but not when you have a connection issue for example

pkg/gateway/gateway.go
```diff
@@ -299,7 +410,31 @@ func (g *gatewayModule) configPath(name string) string {
...
+func (g *gatewayModule) validateNameContract(name string, twinID uint32) error {
+	contractID, err := g.sub.GetContractIDByNameRegistration(string(name))
+	if errors.Is(err, substrate.ErrNotFound) {
+		return errors.New("a name contract must be reserved first")
+	}
+	if err != nil {
+		return err
+	}
+	contract, err := g.sub.GetContract(contractID)
+	if errors.Is(err, substrate.ErrNotFound) {
```
muhamadazmy, line 436: While you return 3 different errors but each one should be handled differently. for example, a NotFound and `!IsCreated` should decomission the domain, but any other error should be "logged" but don't act up on because it can be just a network glitch

PR #1417, opened 2021-10-07
Add prometheus endpoint and nft rules

OmarElawady:


pkg/gridtypes/zos/qsfs.go
```diff
@@ -190,16 +190,19 @@ func (c *QuantumSafeFSConfig) Challenge(w io.Writer) error {
 	return nil
 }
 
-type QuatumSafeFS struct {
+type QuantumSafeFS struct {
```
muhamadazmy, line 193: lol

pkg/network/qsfs.go
```diff
@@ -0,0 +1,109 @@
...
+    ct state invalid drop
+  }
+
+  chain input {
+    type filter hook input priority 0; policy drop;
+    jump base_checks
+    # port for prometheus
+    tcp dport 9100 iifname ygg0 accept
+	# accept only locally generated packets to zdbs
+    tcp dport 9900 iifname lo accept
```
muhamadazmy: Shouldn't we allow ALL lo traffic by default?
```
add rule filter input meta iif lo ct state new accept;
```

pkg/network/qsfs.go
```diff
@@ -0,0 +1,109 @@
...
+    jump base_checks
+  }
+}
+
+`
+
+func applyQSFSFirewall(netns string) error {
+	buf := bytes.Buffer{}
+
+	if err := qsfsNFTTmpl.Execute(&buf, nil); err != nil {
```
muhamadazmy: I don't see why this is needed? the template doesn't have any "inputs" so why not directly nft.Apply of the static string.

pkg/qsfsd.go
```diff
@@ -12,6 +12,12 @@ type QSFSMetrics struct {
 	Consumption map[string]NetMetric
 }
 
+type QSFSInfo struct {
+	Path            string
+	MetricsEndpoint string
```
muhamadazmy, line 17: Why u have entrypoint and port, not just a url as `http://[<ygg-ip>]:port/<path-to-metric-if-exists>`

pkg/qsfsd/qsfs.go
```diff
@@ -103,24 +130,62 @@ func (q *QSFS) Mount(wlID string, cfg zos.QuatumSafeFS) (mountPath string, err e
...
 		cont,
 	)
-	if lerr := q.waitUntilMounted(mountPath, 2*time.Minute); lerr != nil {
+	if lerr := q.waitUntilMounted(ctx, mountPath); lerr != nil {
 		err = lerr
 		return
 	}
+	info.Path = mountPath
+	info.MetricsPort = 9100
+	info.MetricsEndpoint, err = q.waitYggIPs(ctx, networkd, netns)
```
muhamadazmy: As suggested, it's better to construct a URL

pkg/qsfsd/qsfs.go
```diff
@@ -103,24 +130,62 @@ func (q *QSFS) Mount(wlID string, cfg zos.QuatumSafeFS) (mountPath string, err e
...
+
 	return
 }
 
-func (f *QSFS) waitUntilMounted(path string, timeout time.Duration) error {
-	for start := time.Now(); time.Since(start) < timeout; time.Sleep(1 * time.Second) {
-		mounted, err := f.isMounted(path)
-		if err != nil {
-			return errors.Wrap(err, "failed to check if zdbfs is mounted")
+func (q *QSFS) waitYggIPs(ctx context.Context, networkd *stubs.NetworkerStub, netns string) (string, error) {
```
muhamadazmy: Instead of "waiting", a call to networker.QSFSPrepare should instead return the proper Yggdrasil IP. so QSFS can use it directly instead.

pkg/network/qsfs.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	if n.ygg == nil {
+		return netNSName, "", nil
+	}
+	err = n.attachYgg(id, netNs)
+	if err != nil {
+		return "", "", err
+	}
+	ip, err := n.waitYggIPs(netNSName)
```
muhamadazmy: You don't need to wait at all, the Ip is detereminstic and is calcualted int he attachYgg function

muhamadazmy (approved):

PR #1419, opened 2021-10-11
Add rootfs size option for zmachines

OmarElawady:
uses the deprecated "size" option


muhamadazmy (approved):
Looks good, please fix CI before merging

PR #1423, opened 2021-10-11
Update Message struct and update go-rmb version

dmahmouali:


pkg/rmb/rmb.go
```diff
@@ -38,6 +38,7 @@ type Message struct {
 	Retqueue   string   `json:"ret"`
 	Schema     string   `json:"shm"`
 	Epoch      int64    `json:"now"`
+	Proxy      bool     `json:"pxy"`
```
muhamadazmy: Why this is needed int he client ? I see it's not used anywhere !

dmahmouali: > Why this is needed int he client ? I see it's not used anywhere !

I believe the `Unmarshal` is setting it false as default value if it does not exist in the struct. we tried several times and find it false

muhamadazmy (approved):

PR #1431, opened 2021-10-13
Rework workload update order

OmarElawady:
Updates are done (semantically) in this order:
1. deletes first in reverse dependency order (zmachine -> disks)
2. updates and creations are merged and deployed in order (disk create => disk update => zmachine update => zmachine create)

### DIFF
--- a/pkg/provision/engine.go
+++ b/pkg/provision/engine.go
@@ -93,11 +93,11 @@ type NativeEngine struct {
 
 	//options
 	// janitor Janitor
-	twins    Twins
-	admins   Twins
-	order    []gridtypes.WorkloadType
-	revOrder map[gridtypes.WorkloadType]int
-	rerunAll bool
+	twins     Twins
+	admins    Twins
+	order     []gridtypes.WorkloadType
+	typeIndex map[gridtypes.WorkloadType]int
+	rerunAll  bool
 	//substrate specific attributes
 	nodeID uint32
 	sub    *substrate.Substrate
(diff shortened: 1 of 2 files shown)


pkg/provision/engine.go
```diff
@@ -95,6 +96,7 @@ type NativeEngine struct {
 	twins    Twins
 	admins   Twins
 	order    []gridtypes.WorkloadType
+	revOrder map[gridtypes.WorkloadType]int
```
muhamadazmy: revOrder is a very confusing name, because from what I see it holds the 1 based index of the "ordered" types. So i would suggest name 'typeIndex' or something in that direction.

pkg/provision/engine.go
```diff
@@ -503,132 +518,131 @@ func (e *NativeEngine) uninstallWorkload(ctx context.Context, wl *gridtypes.Work
...
-				log.Error().Str("error", result.Error).Msg("failed to deploy workload")
-			}
+	return
+}
 
-			result.Created = gridtypes.Timestamp(time.Now().Unix())
-			wl.Result = *result
-			changed = true
+func (e *NativeEngine) updateDeployment(ctx context.Context, ops []gridtypes.UpgradeOp) (changed bool) {
+	opMap := func(op gridtypes.UpgradeOp) int {
```
muhamadazmy, line 628: This is nice implementation to prioritize Remove operation (in reverse order) and apply the update/create in the right creation order. But I would suggest 2 small things:
1) Move it to a separate method (sortOperations) or something like that (takes only ops as argument) 
2) Add enough code comments because I will look at this in 2 weeks and I will have no idea what is it for

muhamadazmy (approved):

PR #1432, opened 2021-10-14
Add qsfs update method

OmarElawady:
Stuff to figure out:
1. is there a way to get the old workload result as is to avoid recomputing it?
2. is there a way to get the old workload config to check if a non-destructive update is possible?

Update:
A1. Requires embedding the engine in the provisioner object and fetching the old deployment, or modifying the engine code to pass the old workload data.
A2. This was for the cache. Turns out it can be updated by storaged on the fly.

### DIFF
--- a/pkg/container/container.go
+++ b/pkg/container/container.go
@@ -25,6 +25,7 @@ import (
 	"github.com/containerd/containerd/oci"
 	"github.com/containerd/containerd/runtime/restart"
 	"github.com/google/shlex"
+	"github.com/google/uuid"
 	"github.com/patrickmn/go-cache"
 	"github.com/threefoldtech/zbus"
 	"github.com/threefoldtech/zos/pkg"
@@ -296,6 +297,53 @@ func (c *Module) Run(ns string, data pkg.Container) (id pkg.ContainerID, err err
 	return pkg.ContainerID(container.ID()), nil
 }
 
+// Exec executes a command inside the container
+func (c *Module) Exec(ns string, containerID string, timeout time.Duration, args ...string) error {
+	client, err := containerd.New(c.containerd)
+	if err != nil {
+		return err
+	}
+	defer client.Close()
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	ctx = namespaces.WithNamespace(ctx, ns)
+
+	container, err := client.LoadContainer(ctx, containerID)
+	if err != nil {
+		return errors.Wrap(err, "couldn't load container")
+	}
+	t, err := container.Task(ctx, nil)
+	if err != nil {
+		return errors.Wrap(err, "failed to create task")
+	}
+	var p specs.Process
+	p.Cwd = "/"
+	p.Args = args
+	taskID, err := uuid.NewUUID()
+	if err != nil {
+		return errors.Wrap(err, "failed to generate a uuid")
+	}
+	pr, err := t.Exec(ctx, taskID.String(), &p, cio.NullIO)
+	if err != nil {
+		return errors.Wrap(err, "failed to exec new porcess")
+	}
+	if err := pr.Start(ctx); err != nil {
+		return errors.Wrap(err, "failed to start process")
+	}
+	ch, err := pr.Wait(ctx)
+	if err != nil {
+		return errors.Wrap(err, "failed to wait for process")
+	}
+	<-ch
+	ex, err := pr.Delete(ctx)
+	if err != nil {
+		return errors.Wrap(err, "error deleting the created task")
+	}
+	if ex.ExitCode() != 0 {
+		return fmt.Errorf("non-zero exit code: %d", ex.ExitCode())
+	}
+	return nil
+}
 func (c *Module) ensureTask(ctx context.Context, container containerd.Container) error {
 	uri, err := url.Parse("binary://" + binaryLogsShim)
 	if err != nil {
--- a/pkg/container/container.go
+++ b/pkg/container/container.go
@@ -299,20 +299,21 @@ func (c *Module) Run(ns string, data pkg.Container) (id pkg.ContainerID, err err
 
 // Exec executes a command inside the container
 func (c *Module) Exec(ns string, containerID string, timeout time.Duration, args ...string) error {
+	var timedout bool
 	client, err := containerd.New(c.containerd)
 	if err != nil {
 		return err
 	}
 	defer client.Close()
-	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	ctx := namespaces.WithNamespace(context.Background(), ns)
+	createctx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
-	ctx = namespaces.WithNamespace(ctx, ns)
 
-	container, err := client.LoadContainer(ctx, containerID)
+	container, err := client.LoadContainer(createctx, containerID)
 	if err != nil {
 		return errors.Wrap(err, "couldn't load container")
 	}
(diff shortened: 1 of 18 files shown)


muhamadazmy:
> 1. is there a way to get the old workload result as is to avoid recomputing it?

Not at the moment, but this should be easy to implement. the engine already injects the "Deployment" on the context, we can also inject the "Source" deployment (in case of update) that holds the data from the old deployment so it's retrievable in the workload provision methods.

> 2. is there a way to get the old workload config to check if a non-destructive update is possible?
If engine also inject the source deployment in the context, then should comes with it free of charge

pkg/qsfsd/qsfs.go
```diff
@@ -181,6 +181,52 @@ func (f *QSFS) isMounted(path string) (bool, error) {
...
 
+func (q *QSFS) UpdateMount(wlID string, cfg zos.QuantumSafeFS) (pkg.QSFSInfo, error) {
+	var info pkg.QSFSInfo
+	zstorConfig := setQSFSDefaults(&cfg)
+	networkd := stubs.NewNetworkerStub(q.cl)
+	flistd := stubs.NewFlisterStub(q.cl)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
+	defer cancel()
+	netns := networkd.QSFSNamespace(ctx, wlID)
```
muhamadazmy: I would love to instead return the qsfs network info object which has the Ips u need, this way qsfs shouldn't know about interfaces names and/or namespaces. the networkd then can just recompute the ygg ip instead of actually getting all the Ips from the ygg0 interface.

pkg/qsfsd/qsfs.go
```diff
@@ -181,6 +181,52 @@ func (f *QSFS) isMounted(path string) (bool, error) {
...
+	ips, _, err := networkd.Addrs(ctx, "ygg0", netns)
+	if err != nil {
+		return info, errors.Wrap(err, "failed to get ygg ip")
+	}
+	if len(ips) != 1 {
+		return info, errors.Wrap(err, "multiple ips found at ygg interface")
+	}
+	yggIP := ips[0]
+	flistPath, err := flistd.Mount(ctx, wlID, qsfsFlist, pkg.MountOptions{
```
muhamadazmy: I strongly recommend that this instead be a new method (UpdateMount) that takes ONLY the wlID and new size) this method can then only process already mounted flists, to increase(or theoretically decrese) the volume.

My point is, this now gives the impression that an flist, or mount mode is also changible!. The side effect of this method is then not clear no consistent. right ?

An update method on the other hand, is explicit that first, I am doing an update of an already mounted flist, the mount update only supports size, and it works only for rw mounts, hence a ro mount should return an error.

pkg/flist/flist.go
```diff
@@ -353,7 +356,10 @@ func (f *flistModule) Mount(name, url string, opt pkg.MountOptions) (string, err
 	}
 
 	if err := f.valid(mountpoint); err == ErrAlreadyMounted {
-		return mountpoint, nil
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
muhamadazmy: What if this is a `ro` mount that doesn't have a backing volume! also please check the comment regarding the call to flist.Mount

pkg/qsfsd/qsfs.go
```diff
@@ -181,6 +181,52 @@ func (f *QSFS) isMounted(path string) (bool, error) {
...
+	cmd := exec.Command("runc", "--root", fmt.Sprintf("/run/containerd/runc/%s/", qsfsContainerNS), "exec", wlID, "/sbin/zinit", "kill", "zstor")
+	cmd.Stderr = cmd.Stdout
+	err = cmd.Run()
+	if err != nil {
+		op, _ := cmd.Output()
+		log.Error().Err(err).Str("stdout", string(op)).Msg("failed to restart zstor process inside qsfs")
+		return info, errors.Wrap(err, "failed to restart zstor process")
+	}
+	// until threefoldtech/0-stor_v2#71 is fixed
+	out, err := exec.Command("rm", "-f", filepath.Join(flistPath, "var/run/zstor.sock")).Output()
```
muhamadazmy: why do exec and not `os.Remove()` ?!

pkg/qsfsd/qsfs.go
```diff
@@ -181,6 +181,52 @@ func (f *QSFS) isMounted(path string) (bool, error) {
...
+		Limit:    cfg.Cache,
+	})
+	if err != nil {
+		return info, errors.Wrap(err, "failed to get qsfs flist mountpoint")
+	}
+	if err := q.writeQSFSConfig(flistPath, zstorConfig); err != nil {
+		return info, errors.Wrap(err, "couldn't write qsfs config")
+	}
+	mountPath := q.mountPath(wlID)
+	cmd := exec.Command("runc", "--root", fmt.Sprintf("/run/containerd/runc/%s/", qsfsContainerNS), "exec", wlID, "/sbin/zinit", "kill", "zstor")
```
muhamadazmy: this is exposing too much internals of the operation of containerd. 
for abstraction, a call to contd (over the stub) is made to "exec" something inside a container, in that case I imagine the call would be like `err := contd.Exec(containerID, cmd)`

In contd then you already have a client to `containerd` (the one that actually runs the container) with the client u can get the container tasks and then do a task.Exec` to run somehting inside that container (i never tried this before so you probably need to test is quickly)

You always have to keep "separation of concerns" in mind, otherwise we end up with a big pile of spaghetti code if daemons don't respect their poundries.

pkg/container/container.go
```diff
@@ -296,6 +297,53 @@ func (c *Module) Run(ns string, data pkg.Container) (id pkg.ContainerID, err err
...
+		return errors.Wrap(err, "failed to exec new porcess")
+	}
+	if err := pr.Start(ctx); err != nil {
+		return errors.Wrap(err, "failed to start process")
+	}
+	ch, err := pr.Wait(ctx)
+	if err != nil {
+		return errors.Wrap(err, "failed to wait for process")
+	}
+	<-ch
```
muhamadazmy: To avoid getting stuck forever, u also need to select between this and the ctx.Done() channel to make sure it's not stuck forever.

```suggestion
	select {
	  case <-ch:
	  case <-ctx.Done():
	    // force cancelation of the task
```

pkg/container/container.go
```diff
@@ -296,6 +297,53 @@ func (c *Module) Run(ns string, data pkg.Container) (id pkg.ContainerID, err err
...
+	defer client.Close()
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	ctx = namespaces.WithNamespace(ctx, ns)
+
+	container, err := client.LoadContainer(ctx, containerID)
+	if err != nil {
+		return errors.Wrap(err, "couldn't load container")
+	}
+	t, err := container.Task(ctx, nil)
```
muhamadazmy: This creates a new task and does not stop the main task of the container, right ? just confirimg

pkg/flist/flist.go
```diff
@@ -353,7 +356,10 @@ func (f *flistModule) Mount(name, url string, opt pkg.MountOptions) (string, err
...
+		defer cancel()
+		err := f.storage.VolumeUpdate(ctx, name, opt.Limit)
```
muhamadazmy: This section of the code should be removed, since update must be done via the UpdateMount method now

pkg/container/container.go
OmarElawady: yes

muhamadazmy (approved):

PR #1473, opened 2021-11-09
Adds container logs

ashraffouda:

### DIFF
--- a/pkg/container/container.go
+++ b/pkg/container/container.go
@@ -2,6 +2,7 @@ package container
 
 import (
 	"context"
+	"os/exec"
 
 	"github.com/BurntSushi/toml"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


pkg/container/container.go
```diff
@@ -297,6 +298,17 @@ func (c *Module) Run(ns string, data pkg.Container) (id pkg.ContainerID, err err
 	return pkg.ContainerID(container.ID()), nil
 }
 
+// Get logs of the container
+func (c *Module) Logs(ns string, containerID string) (logs string, err error) {
+	filePath := fmt.Sprintf("/var/cache/modules/contd/logs/%s/%s.log", ns, containerID)
```
muhamadazmy: No hard coded paths please. Nor Sprintf when concatinating paths. instead this should be 
```suggestion
	filePath := filepath.Join(c.root, ns, containerID)
```

pkg/qsfsd/qsfs.go
```diff
@@ -130,6 +130,8 @@ func (q *QSFS) Mount(wlID string, cfg zos.QuantumSafeFS) (info pkg.QSFSInfo, err
 	)
 	if lerr := q.waitUntilMounted(ctx, mountPath); lerr != nil {
 		err = lerr
+		logs, _ := contd.Logs(ctx, qsfsContainerNS, wlID)
+		err = fmt.Errorf("%s\nContainer Logs\n%s", lerr.Error(), logs)
```
muhamadazmy:
```suggestion
		err = errors.Wrapf(lerr, "Container Logs:\n%s", logs)
```

muhamadazmy (changes requested):

pkg/qsfsd/qsfs.go
ashraffouda: updated

pkg/container/container.go
ashraffouda: updated

muhamadazmy (approved):

PR #1475, opened 2021-11-09
support sr keys

OmarElawady:


pkg/gridtypes/deployment.go
```diff
@@ -257,12 +313,15 @@ func (d *Deployment) Valid() error {
 }
 
 // Sign adds a signature to deployment given twin id
-func (d *Deployment) Sign(twin uint32, sk ed25519.PrivateKey) error {
+func (d *Deployment) Sign(twin uint32, sk Signer, keyType string) error {
```
muhamadazmy: IMHO,  Signer interface can also return the KeyType, so u don't have to pass it as an extra paramter.

muhamadazmy (approved):

PR #1481, opened 2021-11-18
pass env vars through /.zosrc

OmarElawady:
Once https://github.com/threefoldtech/cloud-container/pull/6 is built, the image url should be updated (to `azmy.3bot/cloud-container-v0.1` or something to avoid breaking testnet)

### DIFF
--- a/pkg/vm/ch.go
+++ b/pkg/vm/ch.go
@@ -212,7 +212,7 @@ func (m *Machine) appendEnv(root string) error {
 		return errors.Wrap(err, "invalid entrypoint")
 	}
 	if len(parts) != 0 {
-		fmt.Fprintf(file, "export init=%s\n", quote(parts[0]))
+		fmt.Fprintf(file, "init=%s\n", quote(parts[0]))
 	}
 	if len(parts) > 1 {
 		var buf bytes.Buffer
--- a/pkg/vm/ch.go
+++ b/pkg/vm/ch.go
@@ -184,10 +184,6 @@ func (m *Machine) Run(ctx context.Context, socket, logs string) error {
 }
 
 func (m *Machine) appendEnv(root string) error {
-	if len(m.Environment) == 0 {
-		return nil
-	}
-
 	stat, err := os.Stat(root)
 	if err != nil {
 		return errors.Wrap(err, "failed to stat vm rootfs")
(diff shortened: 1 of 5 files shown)


OmarElawady:
The escaping here almost works. https://elixir.bootlin.com/linux/v5.6.19/source/lib/cmdline.c#L198 says that double-quotes can't be escaped. The method implementation checks if the first char is `"`, and sets `quote` to true. Then it iterates over the characters flipping quote each time a `"` is encountered. If a space is found while quote is false, the argument is considered to end at this point. So `init=/wrap.sh -- "hi" "hello` is treated as two args: hi and "hello. As indicated in the comment, the double-quotes can't be escaped. base64 is an option and another is to use an environment variable for the args instead of passing in cmdline.

maxux:
It works, just exporting stuff like this ?

OmarElawady:
> It works, just exporting stuff like this ?

The env vars are written in rootfs in `/.zosrc` in a format like:
```bash
export var1='something'
export var2='other "thing" with '"'"'quotes'"'"''
init='/wrap.sh'
set -- 'arg1' 'arg2'
```
The file is then sourced in [cloud-container](https://github.com/threefoldtech/cloud-container/pull/6/files#diff-bb54068aea85faa7e487530083366be9962390af822e4c71ef1aca7033c83e66R76), and the exported vars are passed down to the entrypoint process.

pkg/vm/ch.go
```diff
@@ -59,7 +62,7 @@ func (m *Machine) Run(ctx context.Context, socket, logs string) error {
 		args["--fs"] = filesystems
 	}
 
-	if len(m.Environment) > 0 {
+	if len(m.Environment) > 0 || m.Entrypoint != "" {
```
muhamadazmy: It's not clear here why u suddenly checking on the m.Entrypoint as well. (I know you do this because appendEnv also append the entrypoint args). But this is not clear from the current method name.

Instead `appendEnv` should become `buildZosRC` or something. to make it clear that it does more than just Env.

muhamadazmy: The check for `virtiofs` can also be moved to this method. so in Run you only call `buildZosRC` (or whatever) that checks if it actually has env or entrypoint. If no, return success. Otherwise find the root virtiofs (fail if can't find) then build the file.

muhamadazmy (changes requested):

muhamadazmy (approved):

Issue #1457, opened 2021-10-31
Zmachine can claim any public ip without reserving it


OmarElawady:
After reserving a machine with a public ip `185.206.122.46`. The user can add another public ip to the eth1 interface (`ip addr add ...`) and it will be announced to the network and curling it from outside arrives to the zmachine.

muhamadazmy:
I remember we had firewall rules only to prevent this. I will look into it.

muhamadazmy:
This issue is confirmed

We have nft rules to prevent traffic that is not originated from this IP+MAC combination assigned to the VM on creation, so changing the mac, and or the Ip will render the node un-reachable.

The problem we confirmed is that those rules are now "not working" the counters says the traffic going over the device does not hit the rule at all.

We suspect that this might relate to using macvtap instead of taps . we switched to macvtaps a while back because we had issues with cloud-hypervisor. We might need to test this again with latest release of cloud-hypervisor

muhamadazmy:
We definitely need to move to latest cloud-hypervisor anyway. Then test using taps again see if the issues related to networks was fixed

PR #1488, opened 2021-11-26
Vm ipv6

muhamadazmy:
This introduce a newer version of PublicIP type. In this type you select the kind of your Public IP

This can be `IPv4 Only, IPv6 Only, Or Both`
- IPv4 needs reservation via the contract. So IP workload will fail if contract doesn't have free public ips
- IPv6 are free, they require that the farm support it. Workload will fail is ipv6 is set to true, but no support in the farm.
- Both options can be active at the same time, in that case, the VM will get both kind of IPs
- At least one option must be selected

The reason that IPV6 is implemented as a `PublicIP` workload and a flag on the VM is because of the following:
- The VM gets ONE and ONLY ONE public interface, this interface can have both versions of the IP.
- The public TAP of the VM need to be configured and known the mac address in advance to be able to set it's firewall rules, and other configuration. 
- Think about the PublicIP reservations as a Public Interface. This public interface can have ipv4, ipv6 or both. So IP is a property of the interface, not the VM


pkg/provision/mbus/network.go
```diff
@@ -132,7 +132,7 @@ func (n *Network) listPublicIps() (interface{}, mw.Response) {
 			if err != nil {
 				return nil, mw.Error(errors.Wrap(err, "failed to load deployment"))
 			}
-			workloads := deployment.ByType(zos.PublicIPType)
+			workloads := deployment.ByType(zos.PublicIPv4Type)
```
OmarElawady: Should PublicIPType type be added as well?

pkg/primitives/public_ip.go
```diff
@@ -45,9 +47,10 @@ func (p *Primitives) getAssignedPublicIP(ctx context.Context, wl *gridtypes.Work
 	reserved := contract.PublicIPs
 
 	// make sure we have enough reserved IPs
-	ipWorkloads := deployment.ByType(zos.PublicIPType)
+	// we process both ipv4 type and ip type to be backward compatible
+	ipWorkloads := deployment.ByType(zos.PublicIPv4Type, zos.PublicIPType)
 	if len(ipWorkloads) > len(reserved) {
```
OmarElawady: ipWorkloads can contain ipv6-only workloads which doesn't have to have a corresponding reserved ip.

pkg/network/networker.go
```diff
@@ -415,17 +415,13 @@ nft 'add rule bridge filter prerouting iifname "{{.Iface}}" jump {{.Name}}-pre'
 nft 'add rule bridge filter postrouting oifname "{{.Iface}}" jump {{.Name}}-post'
 
 nft 'add rule bridge filter {{.Name}}-pre ip saddr . ether saddr != { {{.IPv4}} . {{.Mac}} } counter drop'
-{{if .IPv6}}
 nft 'add rule bridge filter {{.Name}}-pre ip6 saddr . ether saddr != { {{.IPv6}} . {{.Mac}} } counter drop'
-{{end}}
 
 nft 'add rule bridge filter {{.Name}}-pre arp operation reply arp saddr ip != {{.IPv4}} counter drop'
 nft 'add rule bridge filter {{.Name}}-pre arp operation request arp saddr ip != {{.IPv4}} counter drop'
```
OmarElawady, line 421: Should similar rules for ipv6 be added to prevent taking over an unassigned ip? Also, the scenario in this [issue](https://github.com/threefoldtech/zos/issues/1457) works with ygg ips.

pkg/gridtypes/zos/ipv4.go
```diff
@@ -1,36 +1,81 @@
...
 }
 
 // Capacity implementation
 func (p PublicIP) Capacity() (gridtypes.Capacity, error) {
-	return gridtypes.Capacity{IPV4U: 1}, nil
+	var c uint64
+	if p.V4 {
+		c = 1
+	}
+	return gridtypes.Capacity{IPV4U: c}, nil
```
OmarElawady, line 69: The capacity calculation here indicates that the user is billed only on active PublicIP workloads. So if a user made a contract with an ip and didn't use it in any deployment, would he be billed?

muhamadazmy: Yes, he will. the billing of the reserved IPv4 is handled by the contract itself. I think we have this here just for the completeness

pkg/primitives/public_ip.go
muhamadazmy: True, good catch

pkg/provision/mbus/network.go
muhamadazmy: Yes, will add it as well.

pkg/network/networker.go
muhamadazmy: No, as far as I remember this filter for arp was needed for ipv4.

PR #1494, opened 2021-12-01
add optional parameter for redis pool size in rmb client

### LINKED ISSUE #1493: Parameterize the size of redis pool in rmb client
- Currently it's tied to 5 connection pools which is not enough to process requests from grid proxy server in parallel

waleedhammam:
- closes: https://github.com/threefoldtech/zos/issues/1493

### DIFF
--- a/pkg/rmb/rmb.go
+++ b/pkg/rmb/rmb.go
@@ -338,7 +338,14 @@ func (m *Message) GetPayload() ([]byte, error) {
 	return base64.StdEncoding.DecodeString(m.Data)
 }
 
-func newRedisPool(address string) (*redis.Pool, error) {
+func newRedisPool(address string, size ...uint32) (*redis.Pool, error) {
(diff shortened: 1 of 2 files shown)


pkg/rmb/rmb.go
```diff
@@ -374,7 +381,7 @@ func newRedisPool(address string) (*redis.Pool, error) {
 
 			return nil
 		},
-		MaxActive:   5,
+		MaxActive:   int(poolSize),
```
muhamadazmy, line 384: if MaxActive is int, then there is no need to make argument a `uint32` can also be `...int` instead 🤷🏼

muhamadazmy (approved):

pkg/rmb/rmb.go
waleedhammam: Would be a sanity check if user entered a negative number

PR #1519, opened 2021-12-22
encrypt the zdb password with the identity private key

OmarElawady:
Instead of checking errors that can return the device id

### DIFF
--- a/pkg/primitives/zdb.go
+++ b/pkg/primitives/zdb.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/primitives/zdb.go
```diff
@@ -36,6 +37,22 @@ type ZDB = zos.ZDB
...
+
+func newSafeError(err error) error {
+	if err == nil {
+		return nil
+	}
+	return &safeError{err}
+}
+func (se *safeError) Error() string {
+	return regexp.
+		MustCompile(`([0-9a-f]{8})\b-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-\b[0-9a-f]{12}`).
```
muhamadazmy: You can compile the regex one time and leave it as a private global variable in this package. Then just use it as 
```go
re.ReplaceAllString(...)
```

muhamadazmy (approved):

PR #1528, opened 2021-12-30
dont wait for qsfs shutdown

OmarElawady:


pkg/qsfsd/qsfs.go
```diff
@@ -57,10 +52,12 @@ func New(ctx context.Context, cl zbus.Client, root string) (pkg.QSFSD, error) {
 		return nil, errors.Wrap(err, "couldn't make qsfs mounts dir")
 	}
 
-	return &QSFS{
+	qsfs := &QSFS{
 		cl:         cl,
 		mountsPath: mountPath,
-	}, nil
+	}
+	go qsfs.periodicCleanup(context.Background())
```
muhamadazmy: You should use the passed context

pkg/container/container.go
```diff
@@ -535,6 +535,40 @@ func (c *Module) List(ns string) ([]pkg.ContainerID, error) {
...
+		return err
+	}
+	// mark this container as perminant down. so the watcher
+	// does not try to restart it again
+	c.failures.Set(string(id), permanent, cache.DefaultExpiration)
+
+	task, err := container.Task(ctx, nil)
+	if err == nil {
+		// err == nil, there is a task running inside the container
+		_, err := task.Wait(ctx)
```
muhamadazmy: Why this call is needed  ?

pkg/qsfsd/qsfs.go
```diff
@@ -76,18 +73,20 @@ func setQSFSDefaults(cfg *zos.QuantumSafeFS) zstorConfig {
 func (q *QSFS) Mount(wlID string, cfg zos.QuantumSafeFS) (info pkg.QSFSInfo, err error) {
 	defer func() {
 		if err != nil {
-			if err := q.Unmount(wlID); err != nil {
-				log.Error().Err(err).Msg("error cleaning up after qsfs setup failure")
-			}
+			q.Unmount(wlID)
```
muhamadazmy, line 76: Why removing the log line?

pkg/qsfsd/qsfs.go
```diff
@@ -76,18 +73,20 @@ func setQSFSDefaults(cfg *zos.QuantumSafeFS) zstorConfig {
...
 		}
 	}()
 	zstorConfig := setQSFSDefaults(&cfg)
 	networkd := stubs.NewNetworkerStub(q.cl)
 	flistd := stubs.NewFlisterStub(q.cl)
 	contd := stubs.NewContainerModuleStub(q.cl)
 
-	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
+	ctx, cancel := context.WithTimeout(context.Background(), 40*time.Second)
```
muhamadazmy, line 84: Why 40 seconds is better then 60 seconds?

muhamadazmy (commented):
Good job as usual, only few questions

pkg/qsfsd/qsfs.go
OmarElawady: The passed context is for the cmdline, the qsfsd service is actually started later with a background context, and I don't know how to access it since it's managed by zbus.

pkg/qsfsd/qsfs.go
OmarElawady: The unmount doesn't return errors anymore it's modified to log errors. It's treated as a best-effort cleanup, because we can't do something about it when it fails. And I preferred to do the most cleanup that could be done (not return err early) even when one cleanup step fails.

pkg/qsfsd/qsfs.go
OmarElawady: Some info:
- It takes ~7 seconds to deploy on fresh zdbs.
- If the zdbs have data, the container will try to download it before exposing the mountpoint. This download phase time can't be known here.
- The container only downloads indexes and the a data file of size 16MB have an index of size 11KB. So if the qsfs was mounted before with 1 TB data written to it, it'll download ~700 MB.

The time to finish this depends on the network speed. And during this time provisiond will be blocked. There's no proper error reporting from the container to zos, when it fails, zinit tries to restart it indefinitely. So the full waiting time will be done for EVERY failed configuration.

This is ofc is no reason for why 40 seconds is better than a minute, but it looks like the fix as you mentioned should be in provisiond.

pkg/container/container.go
OmarElawady: Copy pasted it from the other routine :D but it's not need indeed

PR #1529, opened 2021-12-30
Traefik update

OmarElawady:


pkg/gateway/gateway.go
```diff
@@ -350,11 +352,30 @@ func (g *gatewayModule) isTraefikStarted(z *zinit.Client) (bool, error) {
 	return traefikStatus.State.Is(zinit.ServiceStateRunning), nil
 }
 
+func (g *gatewayModule) traefikBinary(ctx context.Context) (string, string, error) { // path, name, err
+	var exec struct{ Exec string }
+	f, err := os.Open("/etc/zinit/traefik.yaml")
```
muhamadazmy: can u move this logic to Zinit lib instead `pkg/zinit/service.go` it already has AddService, and RemoveService then u can do GetService to return a service object.

pkg/gateway/gateway.go
```diff
@@ -366,7 +387,23 @@ func (g *gatewayModule) ensureGateway(ctx context.Context, forceResstart bool) (
 	if err != nil {
 		return pkg.PublicConfig{}, errors.Wrap(err, "failed to check traefik status")
 	}
-	if running && forceResstart {
+	path, name, err := g.traefikBinary(ctx)
```
muhamadazmy: What if this is the first boot and the service was never configured. this will Always fail because u don't check if the file does not exist

pkg/gateway/gateway.go
```diff
@@ -350,11 +352,30 @@ func (g *gatewayModule) isTraefikStarted(z *zinit.Client) (bool, error) {
 	return traefikStatus.State.Is(zinit.ServiceStateRunning), nil
 }
 
+func (g *gatewayModule) traefikBinary(ctx context.Context) (string, string, error) { // path, name, err
+	var exec struct{ Exec string }
+	f, err := os.Open("/etc/zinit/traefik.yaml")
+	if err != nil {
```
muhamadazmy: you need to handle the IsNotExist error separately because this will get returned if the service is not configured. 

I also really think this has to be done through zinit as well. because the file can exist on the filesystem BUT zinit did never track it up (no zinit monitor was done)

muhamadazmy (changes requested):
There is some issue in the code:
- You don't consider the case when the system is booting for the first time and the traefik service was not configured yet. (on zinit traefik.yaml, and zinit monitor was never done)
In that case ensureGateway need to do this
 - if traefik is not running at all:
    - run traefik
 - if traefik is running:
    - is it running the latest version (yes, no)
  
the code currently assume that traefik.yaml always exists (and that the existance of the file mean the service is monitored, which both assumptions are wrong)

pkg/zinit/commands.go
```diff
@@ -132,6 +135,39 @@ func (c *Client) Status(service string) (result ServiceStatus, err error) {
...
 }
 
+// Exists checks whether a service is monitored or not
+func (c *Client) Exists(service string) (bool, error) {
+	var status ServiceStatus
+	err := c.cmd(fmt.Sprintf("status %s", service), &status)
+	if errors.Is(err, ErrUnknownService) {
+		return false, nil
+	}
+	if err == nil {
```
muhamadazmy:
```suggestion
	else if err == nil {
```

muhamadazmy (changes requested):
one tiny comment

PR #1533, opened 2022-01-03
create the tombstone outside of rootfs

OmarElawady:


pkg/storage/filesystem/device.go
```diff
@@ -243,11 +251,65 @@ func (l *lsblkDeviceManager) lsblk(ctx context.Context, output interface{}, devi
...
+	}
+	var mounts findmntOutput
+	if len(bytes) != 0 {
+		if err := json.Unmarshal(bytes, &mounts); err != nil {
+			return err
+		}
+	}
+	mountpoints := make(map[string]string)
+	for _, m := range mounts.Filesystems {
+		if strings.HasSuffix(m.Options, "subvol=/") {
```
muhamadazmy: i won't grantee that to be at the end of the options always. I would do proper parsing of the mount options or using regex

muhamadazmy:
Also tests are failing

PR #1548, opened 2022-01-09
use wget instead of ping to check internet connectivity

### LINKED ISSUE #1543: Ping check causes internet connected nodes to stop booting when ICMP is blocked
Some ISPs block ICMP or otherwise interfere with users ability to ping external IPs. When Zos boots up, it attempts to ping Google and stops booting if the ping fails. This means that some nodes which have internet connectivity will not boot.

Rather than ping, attempting to get the contents of a site would be a more reliable test of internet connectivity.

OmarElawady:
resolves https://github.com/threefoldtech/zos/issues/1543

### DIFF
--- a/cmds/internet/main.go
+++ b/cmds/internet/main.go
@@ -54,7 +54,7 @@ func main() {
 
 func check() error {
 	f := func() error {
-		cmd := exec.Command("ping", "-c", "1", "google.com")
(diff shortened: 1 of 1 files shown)


muhamadazmy (approved):
This is good and everything except that `internet` service is part of the base image. which means a new image need to be build and this fix will only be available for newly booted nodes.

PR #1554, opened 2022-01-10
expose haspublicipv6 on rmb

ashraffouda:

### DIFF
--- a/pkg/provision/mbus/network.go
+++ b/pkg/provision/mbus/network.go
@@ -39,6 +39,9 @@ func (n *Network) listPortsHandler(ctx context.Context, payload []byte) (interfa
 	return data, nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/provision/mbus/network.go
```diff
@@ -82,6 +86,10 @@ func (n *Network) listPorts(ctx context.Context) (interface{}, mw.Response) {
 	return ports, nil
 }
 
+func (n *Network) hasPublicIPv6(ctx context.Context) interface{} {
+	_, err := stubs.NewNetworkerStub(n.cl).GetPublicIPv6Subnet(ctx)
+	return err == nil
```
muhamadazmy: The error can be nil but the prefix list can also be empty. an error indicates the failure to check, the list actually contains the prefixes which still can be empty which means No Subnets
so this should be 
return err == nil && len(result) > 0

muhamadazmy (approved):

PR #1558, opened 2022-01-12
Support Workloads Updates

muhamadazmy:
So this pull request does the following:
- It changes how we store workload data (and results) on disks, we moved away from files to boltdb
- The storage API has changed so instead of storing the deployment as one blob, it instead keep history of changes to workloads, this allows:
  - Finding out the history of a certain workload
  - Failure on update operation, but yet we know the workload is still working (with older setup) so we can know the "live" state of a workload plus information about changes that did not take place
  - Monitoring now can use this to accurately calculate a workload usage during a specific slice of time.
- Now more workloads can support update, already ZDB and ZMOUNT has been implemented in this PR.
  - Each now can validate if they can do an update and return a proper error
  - Error returned from an update can actually set the workload state to "error" which means workload now is broken and not accessible anymore. OR they can return an `Unchanged` error which means the update has failed but workload state did not change.
  - A caller to update can then check the change log, he/she can tell that the update has failed, but can still see the workload state is okay.
- API changes:
  - A `deployment.get` operation will always return the `current` state of the deployment. So only states available on the workloads are `init` (which means received but not yet deployed), `ok`, and `error`
  -  A `deployment.changes` will return the entire deployment transaction logs in canonical order. Each entry in the log is a `workload` object this includes user input, and result of operation. The state then on the result can be `init`, `ok`, or `error` plus `unchanged` state which is an error that didn't affect previous workload state. And finally the `deleted` state which means the workload has been deleted.
 
pseudo example
```
- entry {name: disk, state: ok, timestamp: time} // disk with name name, has been created
- entry {name: disk, state: unchanged, error: msg, timestamp: time} // disk with name name, has failed to apply operation, (reason is in error) but this error did not affect disk state.  so last OK state applies
- entry {name: disk, state: deleted, timestamp: time} // disk has been removed, any call to deployment.get now will not show this disk. also means a new workload with the same name can be created (even with different type)
```

- [x] Rewrite workloads storage to use boltdb 
- [x] Handle update operation 
- [ ] Monitoring us transactions to calculate usage per twin


cmds/modules/provisiond/main.go
```diff
@@ -147,10 +153,28 @@ func action(cli *cli.Context) error {
...
+	fsStoragePath := filepath.Join(rootDir, fsStorageDB)
+	if _, err := os.Stat(fsStoragePath); err == nil {
+		// if it does we need to migrate this storage to new bolt storage
+		fs, err := fsStorage.NewFSStore(fsStoragePath)
+		if err != nil {
+			return err
+		}
+
+		if err := storageMigration(store, fs); err != nil {
+			log.Error().Err(err).Msg("storage migration failed")
```
OmarElawady: Error is logged and then the workloads dir is removed. Tracing the migration logic, it only returns errors for unlikely scenarios like the rootfs being absent. But other errors like failure to migrate a single deployment is not checked. If there's a bug in the migration logic the old data will be lost forever. Maybe the old workload removal logic can be moved to the migration method to make sure the deployments removed are only the successfully migrated ones?

pkg/primitives/zdb.go
```diff
@@ -537,34 +536,107 @@ func (p *Primitives) zdbDecommissionImpl(ctx context.Context, wl *gridtypes.Work
...
+		// While this makes sense fro ZDB. it will not make zos able to calculate the namespace
+		// consumption because it will only count the size used by the workload from the user
+		// data but not actual size on disk. hence shrinking is not allowed.
+		return result, provision.NewUnchangedError(fmt.Errorf("cannot shrink zdb namespace"))
+	}
 
-// 	if err := container.Delete(ctx, "zdb", containerID); err != nil {
-// 		return errors.Wrapf(err, "failed to delete container %s", containerID)
-// 	}
+	if new.Size == old.Size && new.Password == old.Password {
```
OmarElawady: `&& new.Public == old.Public` is missing

pkg/provision/storage/storage.go
```diff
@@ -1,344 +1,653 @@
...
+			}
+
+			typ, ok := names[workload.Name]
+			if !ok {
+				// not an active workload
+				continue
+			}
+
+			if workload.Type != typ {
+				return fmt.Errorf("database inconsistency wrong workload type")
```
OmarElawady, line 484: This can happen naturally if the user deployed a vm with a name removed it, deployed a zdb with the same name. He shouldn't ofc but we can just skip the mismatching type to avoid breaking the deployment if someone did so.

pkg/provision/storage/storage.go
```diff
@@ -1,344 +1,653 @@
...
+		return nil
+	})
+}
 
-	if err := json.NewEncoder(writer).Encode(d); err != nil {
-		return errors.Wrap(err, "failed to write workload data")
+// Migrate deployment creates an exact copy of dl in this storage.
+// usually used to copy deployment from older storage
+func (b *BoltStorage) Migrate(dl gridtypes.Deployment) error {
+	if err := b.Create(dl); err != nil {
```
OmarElawady: If someone deployed a network and then tried to deploy another deployment with the same network name, the second deployment's network workload would be in `StateError`. Going through the workloads in order would first add the first network workload and mark it in the `global` bucket. For the second one, the network workload addition would fail because it doesn't check for the workload state (that it's in `error` state) and one already exists in the `global` bucket.

cmds/modules/provisiond/main.go
```diff
@@ -347,3 +373,31 @@ func setupMessageBusses(router rmb.Router, cl zbus.Client, engine provision.Engi
...
 }
+
+func storageMigration(db *storage.BoltStorage, fs *fsStorage.Fs) error {
+	log.Info().Msg("starting storage migration")
+	twins, err := fs.Twins()
+	if err != nil {
+		return err
+	}
+	for _, twin := range twins {
+		dls, err := fs.ByTwin(twin)
```
OmarElawady, line 387: the deployments are returned in the directory listing order (lexicographical), right? If so it would cause problems if a vm depends on a network in a preceding deployment, it would be solved if they are sorted as integers instead.

cmds/modules/provisiond/main.go
OmarElawady: Another suggestion -> we can backup the workloads dir in this release to facilitate reverting in case the migration has a very breaking bug. And remove the backup dir in the next release. And the provisiond module doesn't start if the migration has any errors (even if a single deployment migration fails) to avoid receiving new deployments in the new storage (in case we have to revert).

pkg/provision/engine.go
```diff
@@ -489,71 +523,87 @@ func (e *NativeEngine) uninstallWorkload(ctx context.Context, wl *gridtypes.Work
...
 
 	result.Created = gridtypes.Timestamp(time.Now().Unix())
-	wl.Result = *result
-	return err
+
+	if err := e.storage.Transaction(twin, deployment, wl.Workload.WithResults(result)); err != nil {
+		return err
+	}
+
+	return e.storage.Remove(twin, deployment, name)
```
OmarElawady: The workload is removed even if its decommissioning fails, could cause resources leak? The only way to remove it after this happen is to decommission the whole deployment.

pkg/provision/interface.go
```diff
@@ -38,32 +44,95 @@ type Provisioner interface {
...
 	// failed to compute upgrade steps
 	ErrDeploymentUpgradeValidationError = fmt.Errorf("upgrade validation error")
 	// ErrInvalidVersion invalid version error
 	ErrInvalidVersion = fmt.Errorf("invalid version")
 )
 
+// ErrUnchanged can be returned by the Provisioner.Update it means
+// that the update has failed but the workload is intact
+type ErrUnchanged struct {
+	Cause error
```
OmarElawady: should cause be private because `ErrUnchanged` methods doesn't handle nil values (will panic if someone did `ErrUnchanged{nil}` instead of calling the proper constructor function)

pkg/provision/storage/storage.go
```diff
@@ -1,344 +1,653 @@
...
+			return err
+		}
+		if err := dl.Put([]byte(keyMetadata), []byte(deployment.Metadata)); err != nil {
+			return err
+		}
+		sig, err := json.Marshal(deployment.SignatureRequirement)
+		if err != nil {
+			return errors.Wrap(err, "failed to encode signature requirement")
+		}
+		if err := dl.Put([]byte(sig), sig); err != nil {
```
OmarElawady: `dl.Put([]byte(keySignatureRequirement), sig)`

cmds/modules/provisiond/main.go
OmarElawady: nvm, this has to do with storage only, no actual provisioning happen.

pkg/provision/storage/storage.go
OmarElawady: There's a check a couple of lines above if it's loaded, the iteration will be skipped.

cmds/modules/provisiond/main.go
muhamadazmy: I was thinking about this during the implementation. But I didn't want to over engineer it since main-net is still not actively used. May be another simpler solution is to delete deployments that has been succesfully migrated. 

The reason i didn't want to do this, is that if failed deployment were skipped, the migration won't be tried again unless the node has been rebooted or on a new update during this time the deploymend (the un-migrated one) will be un-available. calls to deployment.get will return not found.

I will see if i can keep the failed one to be tried again.

pkg/provision/engine.go
muhamadazmy: Hmm, may be if delete failed, we leave it as in Error state so it can be tried again.

pkg/provision/interface.go
muhamadazmy: True, will change that.

OmarElawady (approved):

PR #1562, opened 2022-01-16
Adds corex support

ashraffouda:

### DIFF
--- a/pkg/primitives/vm.go
+++ b/pkg/primitives/vm.go
@@ -5,7 +5,9 @@ import (
 	"crypto/md5"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net"
+	"os"
 	"path/filepath"
 	"strings"
 
@@ -282,6 +284,12 @@ func (p *Primitives) virtualMachineProvisionImpl(ctx context.Context, wl *gridty
 		if err := p.vmMounts(ctx, deployment, config.Mounts, true, &machine); err != nil {
 			return result, err
 		}
+		if config.Corex {
+			if err := p.copyFile("/usr/bin/corex", filepath.Join(mnt, "corex"), 0755); err != nil {
+				return result, errors.Wrap(err, "failed to inject corex binary")
+			}
+			entrypoint = "/corex --ipv6 -d 7 --interface eth0"
+		}
 	} else {
 		// if a VM the vm has to have at least one mount
 		if len(config.Mounts) == 0 {
@@ -340,6 +348,23 @@ func (p *Primitives) virtualMachineProvisionImpl(ctx context.Context, wl *gridty
 	}
 
 	return result, err
+}
+func (p *Primitives) copyFile(srcPath string, destPath string, permissions os.FileMode) error {
+	src, err := os.Open(srcPath)
+	if err != nil {
+		return errors.Wrapf(err, "Coludn't find %s on the node", srcPath)
+	}
+	defer src.Close()
+	dest, err := os.OpenFile(destPath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, permissions)
+	if err != nil {
+		return errors.Wrapf(err, "Coludn't create %s file", destPath)
+	}
+	defer dest.Close()
+	_, err = io.Copy(dest, src)
+	if err != nil {
+		return errors.Wrapf(err, "Couldn't copy to %s", destPath)
+	}
+	return nil
 }
 
 func (p *Primitives) vmDecomission(ctx context.Context, wl *gridtypes.WorkloadWithID) error {
--- a/pkg/gridtypes/zos/zmachine.go
+++ b/pkg/gridtypes/zos/zmachine.go
@@ -120,7 +120,8 @@ type ZMachine struct {
 	// is going to be used
 	Entrypoint string `json:"entrypoint"`
 	// Env variables available for a container
-	Env map[string]string `json:"env"`
+	Env   map[string]string `json:"env"`
+	Corex bool              `json:"corex"`
 }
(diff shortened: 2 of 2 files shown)


pkg/vm/ch.go
```diff
@@ -179,6 +186,15 @@ func (m *Machine) findVirtioFsMount() (*VirtioFS, error) {
 	return root, nil
 }
 
+func (m *Machine) copyCorex() error {
+	fs, err := m.findVirtioFsMount()
+	if err != nil {
+		return err
+	}
+	root := fs.Path
+	err = cp.Copy("/usr/bin/corex", filepath.Join(root, "/bin/corex"))
```
muhamadazmy: Why do u need a new module to do a file copy, u can just open the 2 files, and do io.Copy().

Also make sure that `<root>/bin` directory exists.

pkg/vm/ch.go
```diff
@@ -61,6 +62,12 @@ func (m *Machine) Run(ctx context.Context, socket, logs string) error {
 	if len(filesystems) > 0 {
 		args["--fs"] = filesystems
 	}
+	if m.Corex {
```
muhamadazmy: I think this code should not be part of `vmd` instead in the provision. Because this should only work in `container` mode. Hence provisiond (vm handler) can decide if it's possible or not to run corex. then it can override the entry point (also copy corex binary) before calling vmd which then has to do nothing.

pkg/vm/machine.go
```diff
@@ -139,6 +139,7 @@ type Machine struct {
 	Config      Config            `json:"machine-config"`
 	Environment map[string]string `json:"environment"`
 	Entrypoint  string            `json:"entrypoint"`
+	Corex       bool              `json:"corex"`
```
muhamadazmy: This is not needed anymore

pkg/vm/manager.go
```diff
@@ -408,6 +408,7 @@ func (m *Module) Run(vm pkg.VM) error {
 			HTEnabled: false,
 		},
 		Entrypoint:  vm.Entrypoint,
+		Corex:       vm.Corex,
```
muhamadazmy: This is not needed anymore

pkg/vm.go
```diff
@@ -144,6 +144,7 @@ type VM struct {
 	// it's up to the caller to check for the machine status
 	// and do clean up (module.Delete(vm)) when needed
 	NoKeepAlive bool
+	Corex       bool
```
muhamadazmy: I don't believe this is needed anymore

pkg/primitives/vm.go
```diff
@@ -280,6 +282,29 @@ func (p *Primitives) virtualMachineProvisionImpl(ctx context.Context, wl *gridty
 		if err := p.vmMounts(ctx, deployment, config.Mounts, true, &machine); err != nil {
 			return result, err
 		}
+		if config.Corex {
+			rootBinPath := filepath.Join(mnt, "bin")
+			_, err = os.Stat(rootBinPath)
+			if err != nil {
```
muhamadazmy: I don't get this check `mnt` exists for sure (granted by the success of the flist mount). the `<rootfs>/bin` on other hand is not granted to exist because it depends on the flist (what if the flist doesn't have bin) this should NOT stop the corex from working. If we want to avoid crating this bin folder under the container <rootfs>. We can simply just copy the corex binary to <rootfs>/ directly. Hence your entrypoint is going to be `/corex` not `/bin/corex`. Which I think is the right way.

pkg/primitives/vm.go
```diff
@@ -280,6 +282,29 @@ func (p *Primitives) virtualMachineProvisionImpl(ctx context.Context, wl *gridty
...
+				return result, errors.Wrap(err, "Couldn't find rootfs")
+			}
+			corexPath := "/usr/bin/corex"
```
muhamadazmy: Can the logic for the `copy` be moved to a new function? the logic should be eventually
```go
if config.Corex {
    if err := copy(<src>, filepath.Join(mnt, "corex"); err != nil {
       return errors.Wrap(err, "failed to inject corex binary")
   }
    entrypoint = "/bin/corex --ipv6 -d 7 --interface eth1"
}
```

go.mod
```diff
@@ -72,6 +72,7 @@ require (
 	github.com/vmihailenco/msgpack v4.0.4+incompatible // indirect
 	github.com/whs/nacl-sealed-box v0.0.0-20180930164530-92b9ba845d8d
 	github.com/yggdrasil-network/yggdrasil-go v0.4.0
+	github.com/yusufpapurcu/wmi v1.2.2 // indirect
```
muhamadazmy: We definetly don't need that. Please make sure both go.mod and go.sum has no changes. Your code does not require any changes to the dependencies, specially not a windows support one.

muhamadazmy (changes requested):

pkg/primitives/vm.go
```diff
@@ -282,6 +284,12 @@ func (p *Primitives) virtualMachineProvisionImpl(ctx context.Context, wl *gridty
 		if err := p.vmMounts(ctx, deployment, config.Mounts, true, &machine); err != nil {
 			return result, err
 		}
+		if config.Corex {
+			if err := p.copyFile("/usr/bin/corex", filepath.Join(mnt, "corex"), 0766); err != nil {
```
muhamadazmy: I think the right permissions are 0755 which is rwx r-x r-x 6 is rw-. it doesn't matter as long as the root has execute permissions but just for the consistency,

muhamadazmy (approved):

PR #1577, opened 2022-01-27
V2 migration

muhamadazmy:
Migration too to v3. it runs as a daemon until it's possible to match a farm payout address from v3 with the v2 farm it's currently on.


pkg/environment/environment.go
```diff
@@ -58,13 +60,15 @@ var (
 	envDev = Environment{
 		RunningMode: RunningDev,
 		BcdbURL:     "https://explorer.devnet.grid.tf/explorer",
+		GraphQlURL:  "https://graphql.dev,grid.tf/graphql",
```
LeeSmet:
```suggestion
		GraphQlURL:  "https://graphql.dev.grid.tf/graphql",
```

LeeSmet (changes requested):

cmds/modules/migrate/main.go
```diff
@@ -0,0 +1,218 @@
...
+
+	if err := crossFingers(); err != nil {
+		return errors.Wrap(err, "failed to reboot")
+	}
+	log.Info().Msg("I should never show up on screen")
+	return nil
+}
+
+func crossFingers() error {
+	f, err := os.OpenFile("/proc/sysrq-trigger", os.O_WRONLY, 0400)
```
maxux:
```go
syscall.Sync()
err = syscall.Reboot(syscall.LINUX_REBOOT_CMD_RESTART)
```

pkg/environment/environment.go
muhamadazmy: Oh good catch, luckily we have no devnet anyway

cmds/modules/migrate/main.go
muhamadazmy: Yup, I started this with a "script" mindset.

PR #1592, opened 2022-02-16
Storage fix for type detection

muhamadazmy:
This PR started as a fix for issue with seektime detection. Somtimes when it timesout we assume it's an hdd device. which then can cause issues if the device was actually previosuly detected as ssd and used for the cache volume.

The fix was to depend mainly on the ROTA flag reported by lsblk . But then i had to clean up lots of code issues regarding listing


OmarElawady (approved):

maxux:
`ROTA` should never be trusted, if there is something wrong with `seektime`, we need to fix it.
Do you know what's wrong ? Any faulty node I can test ?

PR #1619, opened 2022-03-11
New reporting

muhamadazmy:
- Track all VMs consumption every 5 min, create a report of the total consumption during a window every hour 
- Report include 
  - Contract ID
  - Report window
  - Report timestap
  - Network consumption only.
- It introduce only a new way to synchronize contracts with the grid, and to handle cancellation events. 

To be done
- Set contract used capacity on creation/update

### DIFF
--- a/cmds/modules/provisiond/main.go
+++ b/cmds/modules/provisiond/main.go
@@ -42,7 +42,8 @@ const (
 	statisticsModule = "statistics"
 	gib              = 1024 * 1024 * 1024
 
-	boltStorageDB = "workloads.bolt"
+	boltStorageDB    = "workloads.bolt"
+	metricsStorageDB = "metrics.bolt"
 
 	// deprecated, kept for migration
 	fsStorageDB = "workloads"
@@ -78,8 +79,11 @@ func action(cli *cli.Context) error {
 		rootDir      string = cli.String("root")
 	)
 
-	ctx := context.Background()
-	ctx, _ = utils.WithSignal(ctx)
+	ctx, _ := utils.WithSignal(context.Background())
+
+	utils.OnDone(ctx, func(_ error) {
+		log.Info().Msg("shutting down")
+	})
 
 	// keep checking if limited-cache flag is set
 	if app.CheckFlag(app.LimitedCache) {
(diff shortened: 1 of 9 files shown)


pkg/rrd/rrd.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+// Counters return increase in counter value since the given
+// start time.
+func (r *rrdBolt) Counters(since time.Time) (map[string]float64, error) {
+	ts := uint64(since.Unix())
+	ts = (ts / r.window) * r.window
+
+	// we start from the previous slot so we check from the last value.
+	ts -= r.window
```
OmarElawady, line 139: It's not very clear (to me) what increments `Counters` exactly considers. But subtracting the window after flooring to the last Slot will cause the increments inside the slot 70 to be counted twice when calculating `Counters(13) at 73 + Counters(73) at 133` when window is 5, since ts in the second call `ts` will end up equal to 65. Not subtracting the window will consider increments after since (not or at since) which will make consecutive calls consider distinct increments and not skip any slot.

pkg/rrd/rrd.go
```diff
@@ -0,0 +1,224 @@
...
+
+		return nil
+	})
+
+	return
+}
+
+// Counters return increase in counter value since the given
+// start time.
+func (r *rrdBolt) Counters(since time.Time) (map[string]float64, error) {
```
OmarElawady, line 134: if window = 5, and after some time the `getMetrics` method got called at 113 while `Counters` was called at 112. The `getMetrics` will assign the metrics at the slot 110, but it won't be seen by the `Counters` call and it will be skipped in the second `Counters` call (if the -r.window is removed as in the next comment). If it happened once, it'll probably happen for the subsequent calls since Counters and getMetrics are at regular intervals until the getMetrics and Counters calls are flipped.

pkg/rrd/rrd.go
```diff
@@ -0,0 +1,224 @@
...
+		for k, _ := cur.First(); k != nil; k, _ = cur.Next() {
+			if len(k) != 8 {
+				continue // uknown key size
+			}
+
+			if lu64(k) <= retain {
+				// delete the bucket
+				// are we sure this is safe to do while iterating ?
+				_ = tx.DeleteBucket(k)
+			}
```
OmarElawady, line 92: Since the keys are ordered by size, if k > retain the loop can be ended without checking the rest of keys in the db?

cmds/modules/provisiond/reporter.go
```diff
@@ -180,243 +138,179 @@ func (r *Reporter) pusher(ctx context.Context) {
...
+			}
+		}
 	}
+}
 
-	for _, contract := range contracts {
-		// is there a consumption report for a contract
-		delete(local, contract)
+func (r *Reporter) setLastReportTime(ts int64) error {
+	slot, err := r.rrd.Slot()
```
OmarElawady, line 188: is there a reason to not store last report time in a global key?

cmds/modules/provisiond/main.go
```diff
@@ -316,16 +320,31 @@ func action(cli *cli.Context) error {
...
 		return errors.Wrap(err, "failed to setup capacity reporter")
 	}
+
 	// also spawn the capacity reporter
 	go func() {
-		if err := reporter.Run(ctx); err != nil && err != context.Canceled {
-			log.Fatal().Err(err).Msg("capacity reported stopped unexpectedely")
+		for {
+			err := reporter.Run(ctx)
+			if err != context.Canceled {
```
OmarElawady, line 340: `!=` -> `==`

pkg/rrd/rrd.go
muhamadazmy: the keys are ordered sequentially actually, so when scanning the db. It will always be k (older timestamp) to latest timestamp

pkg/rrd/rrd.go
muhamadazmy: I don't get what you mean

pkg/rrd/rrd.go
muhamadazmy: Yeah, may be "Counters" name is not very clear but this computes the "increase" on a specific metric over time. The idea is i take a sample of the reading (network card) every "window" (let's say 5 min) then it's
```
0     5   10  15  20
|-----|-----|----|----|
    ^     ^     ^   ^
```
Then I can compute how much (traffic) was on the card from minute 0 to minute 20, by adding the differences between values as (v5-v0) + (v10-v5) + ... etc.

The problem is now I don't know what is the increase from value at -5, to value at 0 
So to get the cosumption at slot 0-5 i need to know the value at -5, right ? 
hence i start from the previous slot from the 'since'

cmds/modules/provisiond/main.go
muhamadazmy: Yes, thanks

PR #1636, opened 2022-03-30
Cloud init

muhamadazmy:
The idea here is to find a standard way to configure both VMs running with our own custom kernel (what we call container mode) , and VMs running with full user images.

To do this we dropping passing config via cmdline parameters and instead we started to provide config via seed disk image that is compatible with cloud-init. The idea is basically we right cloud-init config files on a small fat32 disk  (2MB in size) and attach it to the VM.

- We customized our own initram to be able to use the config from this seed disk instead of using kernel params.  
- Custom user images need to be a valid cloud images (that includes and starts cloud-init) and it will also use the seed disk and apply config.

This PR works only with https://github.com/threefoldtech/cloud-container/pull/14


OmarElawady (approved):

pkg/vm/manager.go
```diff
@@ -333,6 +329,27 @@ func (m *Module) Run(vm pkg.VM) error {
...
+		},
+	}
+
+	// TODO: user config should be added as another property on the
+	// VM workload data. For now we always use the SSH_KEY from the env
+	// if provided
+	if key, ok := vm.Environment["SSH_KEY"]; ok {
+		cfg.Users = append(cfg.Users, cloudinit.User{
+			Name: "root",
+			Keys: strings.Split(key, "\n"), // in case ssh_key container multiple keys. is this a usecase (?)
```
LeeSmet: I think `,` would make more sense as a separator?

LeeSmet (changes requested):
minor comment but looks good in general, we can start testing this already while figuring out how to pass the user data I think.

pkg/vm/manager.go
muhamadazmy: The reason i used `\n` instead is that what we did before with SSH_KEY is that we directly echoed that in authorized_keys. Hence the way to pass multiple keys was basically to put them as separate lines. I didn't want to break this in case someone is already using it like this.

LeeSmet: alright, not sure if we need it then but can we perhaps support both `\n` and `,` for now?

LeeSmet (approved):
:rocket:

PR #1699, opened 2022-05-16
Fix usage calculation to find disks candidate

### LINKED ISSUE #1693: ZOS only seems to be using the first SSD for deployments
When trying to deploy a vm with 2.5TB of SSD storage capacity (BSC full node), we seem to have stumbled upon an issue where it does not seem to be possible to use more then the first physical SSD in a server.

We mostly tested on mainnet node 293, 303 & 306. These have 7.68TB of SSD storage. When I deployed an ubuntu container (where you can see the host root size) it looked like these servers have 4x 2TB SSD's. What we tried:

- deploy one vm with 5 x 500 GB disks (so to use the block devices as one btrfs filesystem for 2.5TB of storage)
- tried the same with 5 x 400 & 600 GB disks
- one big root disk of 1.7TB and a second 1.7TB disk -> only root disk worked, a second disk not
- 2 separate vm's with each a root disk of 1.7TB -> first vm deployed, second not

We always get the same error back from terraform: 

`Error: error waiting deployment: workload 1 failed within deployment 1373 with error no space left on device`

Logs for node 293 during tests: https://mon.grid.tf/explore?orgId=1&left=%5B%221652281747501%22,%221652284900473%22,%22Loki%22,%7B%22refId%22:%22A%22,%22expr%22:%22%7Bnetwork%3D%5C%22production%5C%22,node%3D%5C%22293%5C%22%7D%22%7D%5D

Used flist: https://hub.grid.tf/tf-official-vms/ubuntu-20.04-lts.flist

muhamadazmy:
After switching vdisks allocations from fallocate to truncate. to suport
ssd over-provisioning. we needed to change and fix how we calculate
used space on a disk as follows

- subvolume with quota limit is assumed to be full usage of the
  subvolume
- subvolume with no-quota limit, we sum total size of all files in that
  subvolume

to find if disk has enough space for allocation, we calculate it as
 used + (requested) < ssd.size * 2

Fixes #1693

### DIFF
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -210,21 +210,22 @@ func (s *Module) initialize() error {
 				continue
 			}
 
+			// for development purposes only
+			if vm {
+				// force ssd device for vms
+				typ = zos.SSDDevice
+
+				if device.Path == "/dev/vdd" || device.Path == "/dev/vde" {
+					typ = zos.HDDDevice
+				}
+			}
+
 			log.Debug().Str("device", device.Path).Str("type", typ.String()).Msg("caching device type")
 			if err := s.cache.Set(device.Name(), typ); err != nil {
 				log.Error().Str("device", device.Path).Err(err).Msg("failed to cache device type")
 			}
 		} else {
 			log.Debug().Str("device", device.Path).Str("type", typ.String()).Msg("device type loaded from cache")
-		}
-
-		if vm {
-			// force ssd device for vms
-			typ = zos.SSDDevice
-
-			if device.Path == "/dev/vdd" || device.Path == "/dev/vde" {
-				typ = zos.HDDDevice
-			}
 		}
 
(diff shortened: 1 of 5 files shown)


pkg/storage/storage.go
```diff
@@ -719,7 +715,7 @@ func (s *Module) checkForCandidates(size gridtypes.Unit, mounted bool) ([]candid
 
 		candidates = append(candidates, candidate{
 			Pool:      pool,
-			Available: usage.Size - (reserved + uint64(size)),
+			Available: usage.Size - (usage.Size + uint64(size)),
```
OmarElawady: Shouldn't this be `SSDOverProvisionFactor * usage.Size - (usage.Used + uint64(size))`? Because of overprovisionning subtracting only from the size can cause overflow

OmarElawady (approved):

PR #1711, opened 2022-05-24
Support pausing workloads

muhamadazmy:
This PR is mainly a refactor to clean up how provisioners are implemented. This is most of the changes are dedicated to. It also allow provisioners to set specific status on workloads by returning a proper Response() as an error. 

This was done in a backward compatible way by still allowing to return a normal error on failure. So a provision action can still return err = nil, or err = some error, or a custom Response (using one of the available Response methods) to force some other status than ok, error.

This allowed us to easily implement Pause/Resume actions on workloads. By default any workload that has no implementation for those actions will still be marked as paused/resumed. 

Workload like VMs, and ZDBs will actually take a pause/resume action to suspend operation on the workload.


pkg/events/events.go
```diff
@@ -98,12 +100,38 @@ func (m *Manager) events(ctx context.Context) error {
...
 						Contract: uint64(e.ContractID),
 						TwinId:   uint32(e.Twin),
 					}
 				}
+
+				for _, e := range events.SmartContractModule_ContractGracePeriodStarted {
+					if e.NodeID != types.U32(m.node) {
+						continue
+					}
+					log.Info().Uint64("contract", uint64(e.ContractID)).Msg("got contract cancel update")
```
OmarElawady: The log in this event and the next is not updated

pkg/primitives/statistics.go
```diff
@@ -139,6 +143,11 @@ func (s *Statistics) hasEnoughCapacity(required *gridtypes.Capacity) error {
 	return nil
 }
 
+// Initialize implements provisioner interface
+func (s *Statistics) Initialize(ctx context.Context) error {
+	return nil
```
OmarElawady: should this call `inner.Initialize()`?

pkg/provision/engine.go
```diff
@@ -732,6 +827,18 @@ func (e *NativeEngine) installDeployment(ctx context.Context, getter gridtypes.W
 	}
 }
 
+func (e *NativeEngine) lockDeployment(ctx context.Context, getter gridtypes.WorkloadGetter, lock bool) {
+	for _, typ := range e.order {
```
OmarElawady, line 845: Does the order matter here depending on whether it's locking or unlocking? So if it's unlocking the order should be reversed as in the uninstallDeployment method?

OmarElawady:
Some places where StateOk may need to be replaced by IsOkay:
- https://github.com/threefoldtech/zos/blob/9f450c7448c4531d0678322bd09277af047c9af4/pkg/provision/storage/storage.go#L671 as the paused vms for example still has its resources allocated
- It's not very clear how the pause/resume would be implemented on the tfchain side, and thus whether it's possible to have ok workloads and paused workloads on the same deployment. But if this is the case, this can cause a problem for paused public ips getting ignored [here](https://github.com/threefoldtech/zos/blob/9f450c7448c4531d0678322bd09277af047c9af4/pkg/primitives/pubip/public_ip.go#L91) which can lead to conflict in cases like (make a deployment with 1 pub ip -> the pub ip gets paused -> a new pubip workload is added, it would get the one assigned to the paused workload -> the old pub ip revives).
- https://github.com/threefoldtech/zos/blob/9f450c7448c4531d0678322bd09277af047c9af4/pkg/provision/mbus/network.go#L146 not sure who depends on this. But paused Public IPs can't be reserved by other contracts, right? So it's considered still reserved by the hosting node even if it's paused?

pkg/primitives/statistics.go
muhamadazmy: True.

pkg/provision/engine.go
muhamadazmy: Yeah, you have a point my plan was actually to make it lock, and unlock in the same order as we do deprovision and provision. I think i was carried away with other changes

muhamadazmy:
@OmarElawady thanks so much for your review. Answer here some of your questions.

- Indeed the Result state comparison need to be reviesd to use IsOkay() when it's really needed. 
- And No, paused public IPs can't be used by other users. It will still be locked on the contract, and charged (will be added to the contract unbilled amount). 
- It's not possible to pause some workloads and not the others. Pause applies to the entire deployment. During a Pause. an update of the contract is not allowed (enforced by the grid by not allowing updating the contract hash)

OmarElawady (approved):

PR #1756, opened 2022-06-27
Br pub exit issue

muhamadazmy:
This PR includes
- Fix public exit nic detection, errors where not handled correctly which caused some nics to be selected when they shouldn't
- Show the current node setup on UI (single, dual) and what nic is selected
- Add authorized (only the farmer can make) calls to the node rmb interface:
  - List all physical devices on the node. 
  - Choose to switch which interface is used by br-pub as an exit interface. The choices can be either `zos` to fall back to single nic setup, or any other free physical nic.
  -


pkg/network/public/public.go
```diff
@@ -100,15 +102,117 @@ func setupPublicBridge(br *netlink.Bridge) error {
...
+			}
+		}
+	}
+
+	return nil, os.ErrNotExist
+}
+
+// SetPublicExitLink rewires the br-pub to a different exit (upstream) device.
+// this upstream device can either be a physical free device, or zos bridge.
+// the method does nothing if
```
LeeSmet: Incomplete docstring

cmds/modules/zui/net.go
```diff
@@ -2,6 +2,7 @@ package zui
 
 import (
 	"context"
+	"fmt"
 	_ "fmt"
```
LeeSmet: double import of "fmt"

LeeSmet (dismissed):
looks good, see comments

OmarElawady (approved):

Issue #1767, opened 2022-07-13
Clearing public config fails


scottyeager:
On mainnet, attempting to clear the public config from a node fails with the errors seen below.

![A14D25BA-8B9C-4370-B920-63D0AED7A19D](https://user-images.githubusercontent.com/17682591/178815605-01d1ddca-48c3-43f6-a0d4-7f76cff98b5d.png)

muhamadazmy:
Actually clearing public config is not supported. The node can't revert having public config cleanly (at the moment). 

@delandtj do you think this is a valid use case that need to be handled?

muhamadazmy:
The problem is the `public` setup on the node is not revertible on the fly without breaking user networks on the node (wireguard) so to support this we give the farmer an option to modify user setups. 

Anyway, the cleanist solution I can think of to is the node can "clean up" config state then reboot itself. This will force the node to have a "working" networking stack. But still affect user networks that was active before this change.

delandtj:
The public config is solely there for giving networking to the private networks. A Farmer can opt out of that and use his public IP's just for VMs. He can also only add an IPv6 public config (you know, IPv6 first, IPv4 as an afterthought). 
Either way, it is advisable to have a few (not necessarily for all nodes in the farm ) IPv4 public addrs on nodes for the Private network mesh and the farmer will earn tokens for that traffic (which can be substantial).
So it's a choice at setup right now, and the necessity to reboot for removing them is not that big of a problem.

delandtj:
Even more, once some WG connections are established, I would make sure we can't delete the public IP addrs any more, as it will break the networks.

muhamadazmy:
> 

Yeah, that's the main problem at the moment. Even if we rebooted the node to cleanely rebuild the network. The user still think his network still uses this specific public IP.

The thing is we can't prevent the user from "unsetting" the public config on the chain. and it's gonna be impossible to report back a setup error to the farmer. unless he checks the zos screen maybe.

muhamadazmy:
I was thinking if both `rebooting` and/or deleting the public namespace will break user networks anyway. And there is no way we can prevent the farmer from deleting the node public config on the chain then may be it's just better to delete the namespace. user networks will stop working in both cases. it's up to the user to reconfigure his deployments once this happens!

What do u think? @delandtj

scottyeager:
Thanks for the discussion guys. Not supporting removing the public config without a reboot makes sense to me, but I saw the errors suggesting that Zos was attempting it.

[...]

PR #1773, opened 2022-07-20
Unset public config

muhamadazmy:
This handles when  a farmer sets public config to empty values. A proper implementation on the chain is needed to actually unset the public config (set it to None), which is not possible at the moment. Hence we instead handle (also) when the values are unset (empty strings) for ipv4 and ipv6.

When Unsetting the public namespace. the system can not recover cleanly because there are many different parts that rely on the public namespace. (yggrasil, traefik gateway, and users networks). Some of them (like wireguard ports) can't be moved to host namespace hence a full reboot of the node is needed.

for that specific reason the unsetting of public config was not supported but seems it's a use case that we need to handle

Related to #1767

### DIFF
--- a/cmds/modules/networkd/main.go
+++ b/cmds/modules/networkd/main.go
@@ -87,7 +87,7 @@ func action(cli *cli.Context) error {
 	public.SetPersistence(root)
 
 	pub, err := public.LoadPublicConfig()
-	log.Debug().Err(err).Msgf("public interface configred: %+v", pub)
+	log.Debug().Err(err).Msgf("public interface configured: %+v", pub)
 	if err != nil && err != public.ErrNoPublicConfig {
 		return errors.Wrap(err, "failed to get node public_config")
 	}
--- a/cmds/modules/noded/public.go
+++ b/cmds/modules/noded/public.go
@@ -52,6 +52,12 @@ func public(ctx context.Context, nodeID uint32, env environment.Environment, cl 
 		return sub.GetNode(nodeID)
 	}
 
+	// this to check if the config actually is holding any values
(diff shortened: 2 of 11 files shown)


OmarElawady (approved):
Some thoughts after discussing with Azmy. This will cause the following to happen after the farmer unsets the public config:
- All VMs will be disconnected from the wireguard network. But it will still be accessible over ygg and public ip and otherwise functional.
- All k8s nodes deployed on this node will get disconnected and it has to be restarted to start working again.
- All wireguard access that used this node as the access point will be lost.

muhamadazmy:
Yeah, the side effects are catastrophic (for user private networks over wireguard) and there is noway to recover. this is why this feature was not implemented from the beginning. Although we support updating (changing the public IP) which has similar effects (user networks need to be reconfigured) but it's not as destructive as this one.

The reboot is needed to make sure everything will restart knowing that there is no public namespace. Hence the farmer need to be aware that this change is bad and is encourage against doing it. May be even apply a reward penalty if the farmer changed or deleted the node public config.

@LeeSmet what do u think?

LeeSmet:
Tokenomics currently don't include a penalty for any situation other than being offline, so while that is a good idea, it currently is not feasible. I'd say an automatic reboot is the best option indeed. In a perfect world, we wouldn't allow unsetting this, however this would lead to the farmer just breaking the network to the node anyway so there's nothing we can really do to stop them from doing this. Technically, the br-pub bridge could be refitted to be linked to zos, which would likely break existing connections but otherwise allow (public) traffic. The main issue here is that private networks will break if the remote ends don't have a publicly accessibe IP.

PR #1787, opened 2022-08-18
add tpm package

mariobassem:


muhamadazmy:
@MarioBassem you can already see that the github actions are failing here https://github.com/threefoldtech/zos/actions/runs/2883919726

muhamadazmy:
Okay, this might have been a github action issue, when i did re-run it started building. will see if it finishes correctly

.github/workflows/bins.yaml
```diff
@@ -101,3 +101,9 @@ jobs:
       package: node-exporter
     secrets:
       token: ${{ secrets.HUB_JWT }}
+  tpm:
+    uses: ./.github/workflows/bin-package.yaml
```
muhamadazmy: bin-package.yaml uses `ubunut-latest` for the build. This is mostly fine because all other packages are doing "static builds" so they don't dynamically link to libraries that are available on the base image. Unlike tpm which can't be build statically (or at least I couldn't figure out how) this means the tpm and tss need to be build against the same version of the base image libraries. 

The 0-initramfs builds on ubuntu-18.04 hence the tpm pacakge needs also to build against the same ubuntu version.

Hence i suggest that u make a copy of `bin-pacakge.yaml` (name it `bin-pacakge-18.04.yaml` or something) then change it to use the right ubuntu version, then use it in this line instead of this one.

bins/packages/tpm/tpm.sh
```diff
@@ -0,0 +1,98 @@
...
+extract_tpm() {
+    if [ ! -d "tpm2-tools-${TPM_VERSION}" ]; then
+        echo "[+] extracting: tpm2-tools-${TPM_VERSION}"
+        tar -xf ${DISTDIR}/tpm2-tools-${TPM_VERSION}.tar.gz -C ${WORKDIR}
+    fi
+}
+
+prepare_tpm() {
+    echo "[+] preparing tpm"
+    github_name "tpm-${TPM_VERSION}"
```
muhamadazmy, line 70: I added this line (check my other comment) because it's needed to let the build system know what is the full package name (name+version) to be created on the hub. After pushing this you can already see that this package was created https://hub.grid.tf/tf-autobuilder/tpm-5.2.flist.md

If you expand the flist content you will see all the files that are included in the package. It looks good from the first look, but it also include some files that are not needed. This include all ".h" files, all the ".a" files (static libs) and all the man pages and bash-completion scripts.

> I am wondering if there is a `make install` flag that avoid copying those files and will only install the necessary binaries and libraries please check that, otherwise those files need to be deleted from the output directory before publishing

muhamadazmy (changes requested):
Thanks @MarioBassem for you work, please check the PR comments.

maxux:
I could get a static build of tpm2-tools with this custom script (make to be run on alpine docker):
```bash
#!/bin/sh
apk add alpine-sdk cmake
apk add openssl-dev curl-dev json-c-dev linux-headers
apk add zlib-static brotli-static nghttp2-static curl-static openssl-libs-static

cd /mnt/tmp

rm -rf *.tar.gz
rm -rf json-c-*
rm -rf tpm2-*

export TPM_VERSION="5.2"
export TSS_VERSION="3.2.0"

wget https://github.com/json-c/json-c/archive/refs/tags/json-c-0.16-20220414.tar.gz
wget https://github.com/tpm2-software/tpm2-tools/releases/download/${TPM_VERSION}/tpm2-tools-${TPM_VERSION}.tar.gz
wget https://github.com/tpm2-software/tpm2-tss/releases/download/${TSS_VERSION}/tpm2-tss-${TSS_VERSION}.tar.gz


tar -xvf json-c-0.16-20220414.tar.gz

cd json-c-json-c-0.16-20220414/
cmake .
make -j 12
make install
cd ..

tar -xvf tpm2-tss-3.2.0.tar.gz

cd tpm2-tss-3.2.0/
./configure "CFLAGS=--static" --enable-shared=no
make -j 12
make install
cd ..

tar -xvf tpm2-tools-5.2.tar.gz

cd tpm2-tools-5.2/

export LIBS="/usr/lib/libbrotlienc.a /usr/lib/libbrotlidec.a /usr/lib/libbrotlicommon.a /usr/lib/libnghttp2.a /usr/lib/libssl.a /usr/lib/libcrypto.a /lib/libz.a"
./configure "CFLAGS=--static" --enable-shared=no --disable-hardening

make -j 12
make DESTDIR=/mnt/tmp/root install

file /mnt/tmp/root/usr/local/bin/tpm2
file /mnt/tmp/root/usr/local/bin/tss2
```

Which works, I have a static linked tpm2, but... it doesn't run correctly.
According to this issue: https://github.com/tpm2-software/tpm2-tools/issues/1770#issuecomment-547573517

It seems that internally, `tpm2-tools` use `dlopen()` to load stuff and this, require dynamic linking. So we are kinda stuck with dynamic library for now.

Or we use directly `/dev/tpm0` which is _way_ harder, haha

muhamadazmy (approved):

PR #1790, opened 2022-08-23
using tpm to store node identity seed

muhamadazmy:
Depends on #1787 

This will does the following:
- For nodes that has tpm, use tpm to store node seed
- Nodes without tpm, will continue use a file on disk
- Nodes that already has a key file, will move that key to tpm and delete the key if the node has tpm

This PR uses the tpm2 utils. The branch and build and install tpm is not merged yet. 

> There are changes related to the update to make sure identityd restarts after installing the latest tpm2 binaries

### DIFF
--- a/pkg/identity/builder.go
+++ b/pkg/identity/builder.go
@@ -37,6 +37,13 @@ func NewStore(root string) (store.Store, error) {
 		return tpm, nil
 	}
 
+	if ok, err := tpm.Exists(); err == nil && ok {
+		// so there is a key on disk, but tpm already has a stored key
+		// then we still just return no need for migration to avoid
+		// overriding the key in tpm
+		return tpm, nil
+	}
+
 	// if we failed to get the key from store
 	// may be better generate a new one?
 	// todo: need discussion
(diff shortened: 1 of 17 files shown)


pkg/identity/builder.go
```diff
@@ -0,0 +1,71 @@
...
+
+	if !exists {
+		return tpm, nil
+	}
+
+	if ok, err := tpm.Exists(); err == nil && ok {
+		// so there is a key on disk, but tpm already has a stored key
+		// then we still just return no need for migration to avoid
+		// overriding the key in tpm
+		return tpm, nil
```
abom, line 44: What about old data/signatures in this case?

muhamadazmy: This is actually to avoid overriding the key inside tpm if someone decided to put another key on disk. The idea is that key migration will happen one time (and one time only) and then delete the key file. If suddenly a key appeared then it means something is fishy and the new key file is just ignored.

abom: Ok, got it. I think we also assume there's no other key at our chosen address, right? (not managed by us).

muhamadazmy: That's actually hard to grantee. Maybe we should add a validation (and clearing step) if the key is invalid.

Issue #1788, opened 2022-08-19
qsfs decomission clean up


muhamadazmy:
I noticed that some nodes has qsfs error while doing a decomissions. but they all sound like an avoidable errros. It's not an error If you are deleting something that is already doesn't exist! so the code can be a little nicer during destruction to not fail if what it is trying to destroy is already not there

> We also need to check why this is happening as well, are we doing double decomission? 
 
```
[+] qsfsd: 2022-08-19T07:28:28Z error failed to destroy qsfs network error="didn't find qsfs namespace: file does not exist"
[+] qsfsd: 2022-08-19T07:28:28Z error couldn't get qsfs metrics error="didn't find qsfs namespace: file does not exist" id=51-5481-wed2710d2
[+] qsfsd: 2022-08-19T07:28:28Z error failed to delete qsfs container error="container \"51-5481-wed2710d2\" in namespace \"qsfs\": not found"
[+] qsfsd: 2022-08-19T07:28:28Z error failed to unmount mount path 1st time error="no such file or directory"
[+] qsfsd: 2022-08-19T07:28:28Z error failed to unmount mount path 1st time error="no such file or directory"
[+] storaged: 2022-08-19T07:28:28Z info Deleting volume 51-5481-wed2710d2
[+] storaged: 2022-08-19T07:28:28Z warn Could not find filesystem 51-5481-wed2710d2
[+] qsfsd: 2022-08-19T07:28:28Z error failed to destroy qsfs network error="didn't find qsfs namespace: file does not exist"
[+] qsfsd: 2022-08-19T07:28:28Z error couldn't get qsfs metrics error="didn't find qsfs namespace: file does not exist" id=51-5513-wed2710d2
[+] qsfsd: 2022-08-19T07:28:28Z error failed to delete qsfs container error="container \"51-5513-wed2710d2\" in namespace \"qsfs\": not found"
[+] qsfsd: 2022-08-19T07:28:28Z error failed to unmount mount path 1st time error="no such file or directory"
[+] qsfsd: 2022-08-19T07:28:28Z error failed to unmount mount path 1st time error="no such file or directory"
[+] storaged: 2022-08-19T07:28:28Z info Deleting volume 51-5513-wed2710d2
[+] storaged: 2022-08-19T07:28:28Z warn Could not find filesystem 51-5513-wed2710d2
[+] qsfsd: 2022-08-19T07:28:28Z error failed to destroy qsfs network error="didn't find qsfs namespace: file does not exist"
[...]

PR #1792, opened 2022-09-05
qsfs: deprovision: don't fail if it's already marked...

abom:
also don't fail if the container does not exist too.

Related:

* #1788

### DIFF
--- a/pkg/container/container.go
+++ b/pkg/container/container.go
@@ -551,7 +551,6 @@ func (c *Module) SignalDelete(ns string, id pkg.ContainerID) error {
 	container, err := client.LoadContainer(ctx, string(id))
 	if err != nil {
 		if errdefs.IsNotFound(err) {
-			log.Warn().Str("id", string(id)).Msg("attempt to delete a non-existing container")
 			return nil
 		}
 		return err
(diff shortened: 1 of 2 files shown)


pkg/qsfsd/qsfs.go
```diff
@@ -267,7 +267,8 @@ func (q *QSFS) SignalDelete(wlID string) error {
 	marked, err := q.isMarkedForDeletion(ctx, wlID)
 
 	if marked {
-		return errors.New("already marked for deletion")
+		log.Warn().Str("workload", wlID).Msg("already marked for deletion")
```
muhamadazmy: I don't believe the warn is valid or even needed here

pkg/container/container.go
```diff
@@ -550,8 +550,13 @@ func (c *Module) SignalDelete(ns string, id pkg.ContainerID) error {
 	// log.Debug().Str("id", string(id)).Msg("fetching container")
 	container, err := client.LoadContainer(ctx, string(id))
 	if err != nil {
+		if errdefs.IsNotFound(err) {
+			log.Warn().Str("id", string(id)).Msg("attempt to delete a non-existing container")
+			return nil
+		}
 		return err
 	}
+
```
muhamadazmy:
```suggestion
        if errdefs.IsNotFound(err) {
             return nil
        } else if err != nil {
             return err
        }
```

muhamadazmy: also I don't think a Warn is needed. it's a not warning case

muhamadazmy (changes requested):

muhamadazmy (approved):

Issue #1793, opened 2022-09-07
switch to using `dhcpcd` instead of `udhcpc`


muhamadazmy:
Potential fix for #1791 

The switch need to make sure current configured instanced of udhcpc is decently stopped, and replaced by the proper command for dhcpcd.

muhamadazmy:
@abom any updates? do you need help. I can guide you on what you need to do.

PR #1801, opened 2022-09-15
use dhcpcd instead of udhcpc

abom:
Related:

#1793

Testing replacing older running `networkd`:

```
~ # zinit status dhcp-npub4
---
name: dhcp-npub4
pid: 967
state: Running
target: Up
after: {}
~ # cat /etc/zinit/dhcp-npub4.yaml 
exec: ip netns exec ndmz /sbin/udhcpc -v -f -i npub4 -t 20 -T 1 -s /usr/share/udhcp/simple.script
~ # zinit status dhcp-npub4
---
name: dhcp-npub4
pid: 1282
state: Running
target: Up
after: {}
~ # cat /etc/zinit/dhcp-npub4.yaml 
exec: ip netns exec ndmz /usr/sbin/dhcpcd npub4 -B
```

### DIFF
--- a/cmds/internet/main.go
+++ b/cmds/internet/main.go
@@ -2,12 +2,11 @@ package main
 
 import (
 	"flag"
-	"fmt"
 	"os"
 	"os/exec"
 	"time"
 
-	"github.com/cenkalti/backoff/v3"
+	"github.com/cenkalti/backoff"
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog/log"
 	"github.com/vishvananda/netlink"
@@ -15,6 +14,7 @@ import (
 	"github.com/threefoldtech/zos/pkg/app"
 	"github.com/threefoldtech/zos/pkg/network/bootstrap"
 	"github.com/threefoldtech/zos/pkg/network/bridge"
+	"github.com/threefoldtech/zos/pkg/network/dhcp"
 	"github.com/threefoldtech/zos/pkg/network/ifaceutil"
 	"github.com/threefoldtech/zos/pkg/network/options"
 	"github.com/threefoldtech/zos/pkg/network/types"
@@ -22,8 +22,6 @@ import (
 
 	"github.com/threefoldtech/zos/pkg/version"
 )
-
-const dhcpService = "dhcp-zos"
 
 func main() {
 	app.Initialize()
@@ -74,8 +72,6 @@ func check() error {
 
 func configureZOS() error {
 	f := func() error {
-		z := zinit.Default()
-
 		log.Info().Msg("Start network bootstrap")
 
 		ifaceConfigs, err := bootstrap.AnalyzeLinks(
@@ -126,26 +122,19 @@ func configureZOS() error {
 			return errors.Wrapf(err, "could not bring %s up", zosChild)
 		}
 
-		log.Info().Msg("writing dhcpcd init service")
-
-		err = zinit.AddService(dhcpService, zinit.InitService{
-			Exec:    fmt.Sprintf("/usr/sbin/dhcpcd %s -B", types.DefaultBridge),
-			Oneshot: false,
-			After:   []string{},
-		})
-
-		if err != nil {
-			log.Error().Err(err).Msg("fail to create dhcp-zos zinit service")
+		dhcpService := dhcp.NewService(types.DefaultBridge, "", zinit.Default())
+		if err := dhcpService.DestroyOlderService(); err != nil {
+			log.Error().Err(err).Msgf("failed to destory older %s service", dhcpService.Name)
+			return err
+		}
+		// create the new service anyway
+		if err := dhcpService.Create(); err != nil {
+			log.Error().Err(err).Msgf("failed to create %s service", dhcpService.Name)
 			return err
 		}
 
-		if err := z.Monitor(dhcpService); err != nil {
-			log.Error().Err(err).Msg("fail to start monitoring dhcp-zos zinit service")
-			return err
-		}
-
-		if err := z.Start(dhcpService); err != nil {
-			log.Error().Err(err).Msg("fail to start dhcp-zos zinit service")
+		if err := dhcpService.Start(); err != nil {
+			log.Error().Err(err).Msgf("failed to start %s service", dhcpService.Name)
 			return err
 		}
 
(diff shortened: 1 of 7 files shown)


cmds/internet/main.go
```diff
@@ -124,10 +126,16 @@ func configureZOS() error {
...
 		}
 
-		log.Info().Msg("writing udhcp init service")
+		log.Info().Msg("writing dhcpcd init service")
+
+		// terminate older udhcpc service if found
+		if err := z.Terminate(dhcpService, "udhcpc"); err != nil {
+			log.Error().Err(err).Msg("fail to terminate older dhcp-zos (udhcpc) zinit service")
+			return err
+		}
```
muhamadazmy: Terminate is not enough. since zinit still know the service was that specific name, replacing the zinit yaml file will not override the service configuration in zinit memory.
in other words, when u terminate a service, then change the file then start it will still start with the original config (before replacing the file)
Instead, what you need to do is to `terminate`, `forget`, `change the config file`, then `monitor`.

That being said, the init service does not live forever, it starts one time on system boot and is included in the basic zos image. It means that internet service does NOT need to check or replace the config, because there is nothing before it starts, which means it can directly use the dhcpcd service. 

It's up to `networkd` which can be updated during the machine runtime to check if the dhcp service is in fact using the dhcpcd binary, otherwise it then should terminate, forget, repalce, and monitor using the new config.

muhamadazmy: I just noticed you added the new `Terminate` function to which does exactly that. which is nice, but i would rename it to something like `Destroy` instead of terminate.

I still think `internet` does not need to care since it's one time thing that starts on boot, hence there is no service before already.

pkg/zinit/commands.go
```diff
@@ -304,3 +304,27 @@ func (c *Client) Forget(service string) error {
 func (c *Client) Kill(service string, sig Signal) error {
 	return c.cmd(fmt.Sprintf("kill %s %s", service, string(sig)), nil)
 }
+
+// Terminate a service completely (stop, forget and remove) if it matches an exec pattern
+func (c *Client) Terminate(service string, execPattern string) error {
```
muhamadazmy: I have some concerns about this method. It's very use case specific that it shouldn't be part of the client. I mean there is no actual use case for it outside this PR. Hence i suggest that we need to either generalize the use case, or move it outside of the client.

Of course generalizing the use case is the best thing to do. And i suggest the following "interface"

```go
func (c *Client) Destroy(filter ...Filter) error {
   // get all services, and for each service that matches ALL the filters, 
   // apply the destroy. operation
}

type Filter interface{
    matches(service *Service) bool
}

type nameFilter struct{
  name string
  }
// implement FIlter for the nameFilter type
// add other filter like `exe` filter  

// then u can call the Destroy like this
client.Destoy(WithName("dhcp-zos"), WithExe("udhcpc"))

// also if filters list is empty the method should return an error that search pattern is too wide
```

pkg/zinit/commands.go
```diff
@@ -304,3 +304,27 @@ func (c *Client) Forget(service string) error {
...
+	initService, err := c.Get(service)
+	if err != nil {
+		// service is not there, mostly
+		return nil
+	}
+
+	if !strings.Contains(initService.Exec, execPattern) {
```
muhamadazmy: contains is not a good way to match on an exec instead do `filepath.Base(initService.Exec) == exec` to match exactly on the binary name. If you wanna implement a very powerful pattern matching then use regex. the problem with contains is that if your pattern is something like "bin" then you will kill all services.

so we can have 2 filters implemented
1) WithExec("bin-name") need to be documented this match on the executable name (not the full path)
2) WithExecRegex("<regex-pattern>") full regex match on the entire exec (not needed to be implemented now but can be easily added in the future.

Also since u pass the entire service object to the filter, it's easy now to build filters that also matches on other services patterns.

Even a better design for this api is to have 2 functions 
```
func (c *Client) Matches(filter ...Filter) ([]Service, error) 
// and 
func (c *Client) Destroy(wait, ...string) error
```

Which means filtering on service and destroying are 2 separate operations. So you can find services to do something else with and not always destroy

muhamadazmy (changes requested):

pkg/zinit/commands.go
abom: IMHO think that `Contains` is enough and it ony matches the exec of a single service and we won't allow a match with `bin` for example to be merged I think :laughing:.

But, I started to implement your suggestion, but got some concerns

* `Exec` field of `InitService` does not actually contain a "path" only (most of the time), it might even contain a shell script.
* `InitService` does not have a name field, but I think adding it would be easy (or using service name directly) and won't affect anything else.
* As contains, how would we control the input of `Regex` matches?

So, I suggest we can add these filtering check, but only apply it to exact "service" with a given name, not all services.

pkg/zinit/commands.go
```diff
@@ -304,3 +323,29 @@ func (c *Client) Forget(service string) error {
 func (c *Client) Kill(service string, sig Signal) error {
 	return c.cmd(fmt.Sprintf("kill %s %s", service, string(sig)), nil)
 }
+
+// Destroy a service completely (stop, forget and remove) if all filters are matched
+func (c *Client) Destroy(service string, filters ...Filter) error {
```
muhamadazmy: I suggested also that the searching `the matching` and `the destroying` are to be done sepeartely. So you have a Match method, and the Destroy method that takes list of service names.

muhamadazmy: Why you still provide the service name? this should be in the filters.

pkg/zinit/commands.go
```diff
@@ -304,3 +323,29 @@ func (c *Client) Forget(service string) error {
...
+	// FIXME: should we allow empty filter so it would only match with service name?
+	for _, filter := range filters {
+		if !filter.matches(&initService) {
+			return nil
+		}
+	}
+
+	if err := c.StopWait(30*time.Second, service); err != nil {
+		return err
+	}
```
muhamadazmy: Because you can actually destroy multiple services (check the comments above) it means you need to use the StopAll method instead

Again, we are trying to make this as generic as possible and not crafted for a specific use case

muhamadazmy (changes requested):

muhamadazmy (approved):

Issue #1842, opened 2022-11-20
Coupling  the `tls_passthrough` option with the type of the service protocol.


sameh-farouk:
In the current implementation of our gird gateways, I noticed that we are configuring a TCP service in Traefik routes in case of `tls_passthrough` is enabled (this allows you to proxy any TCP-based service, not just HTTP), otherwise, we configure an HTTP service.

This (inconsistency) raises a few questions:
- Why do we support TCP proxy only when tls_passthourgh is enabled?
https://github.com/threefoldtech/zos/blob/11516d751fe9cd69b46a3b548361452dda555982/pkg/gateway/gateway.go#L605-L609
- Why we are hiding the fact that by enabling `tcp_passthrough` we are proxy to any TCP service, by forcing the user to include `https` as a schema although **the schema has no effects** (aside from the (unnecessary) validation in zos, the schema is not used anywhere else, we use only the host part)?
https://github.com/threefoldtech/zos/blob/11516d751fe9cd69b46a3b548361452dda555982/pkg/gateway/gateway.go#L574-L576
https://github.com/threefoldtech/zos/blob/a5793cb13d6d9aa51f13b2872544212de795f727/pkg/gateway/gateway.go#L577-L579

IMHO, there is no reason for **coupling** the `tls_passthrough` option with the type of service protocol (TCP, HTTP). instead, we could have an extra flag, say `tcp_prxoy`, to enable Using TCP as the service protocol independently from the way encrypted traffic is handled (by passthrough, or offloading). regarding the validation, in case `tcp_proxy` is enabled the schema must be empty, otherwise, it should be one of HTTP, or HTTPS.

We need to support these scenarios:
  - I  have an HTTP service running in my backend
    - `tcp_prxoy` = false

  - I  have something like an FTP service (non-HTTP service) running in my backend
    - `tcp_prxoy` = true

  - For any of the services above I can do TLS passthrough and handle the encrypted traffic on my server, or opt-in to decrypt the data with a gateway. 
    - `tls_passthrough` could be `true` or `false` regardless of the service protocol set by the `tcp_proxy`  flag. 

In this way, the behavior of the gateway is more defined and easier to reason about.

muhamadazmy:
We agreed not to do this atm. the changes in the PR are enough for now

Issue #1794, opened 2022-09-07
Fqdn gateway with tls passthrough failed to start


mohamedamer453:
After creating a fqdn gateway with the `tls_passthrough` option set to `true` and the backend is a vm running a caddy server, the deployment was successful but i was not able to access it from the domain with the following error.

`This site can't provide a secure connection` 

and from caddy i got these errors.

```
2022/09/07 11:25:52.346  ERROR  tls.issuance.acme.acme_client  challenge failed  {"identifier": "http://fqdncaddy.gridtesting.xyz", "challenge_type": "http-01", "problem": {"type": "urn:ietf:params:acme:error:unauthorized", "title": "", "detail": "185.69.167.81: Invalid response from http://http://fqdncaddy.gridtesting.xyz/.well-known/acme-challenge/DXPVMC20AIcQsD43Gvg58SpyV4N8J7V0MReiT6Qw-L4: 404", "instance": "", "subproblems": []}}
2022/09/07 11:25:52.348  ERROR  tls.issuance.acme.acme_client  validating authorization  {"identifier": "http://fqdncaddy.gridtesting.xyz", "problem": {"type": "urn:ietf:params:acme:error:unauthorized", "title": "", "detail": "185.69.167.81: Invalid response from http://http://fqdncaddy.gridtesting.xyz/.well-known/acme-challenge/DXPVMC20AIcQsD43Gvg58SpyV4N8J7V0MReiT6Qw-L4: 404", "instance": "", "subproblems": []}, "order": "https://acme-v02.api.letsencrypt.org/acme/order/721377777/123320026447", "attempt": 1, "max_attempts": 3}
2022/09/07 11:25:53.669  INFO  tls.issuance.acme.acme_client  trying to solve challenge  {"identifier": "http://fqdncaddy.gridtesting.xyz", "challenge_type": "tls-alpn-01", "ca": "https://acme-v02.api.letsencrypt.org/directory"}
2022/09/07 11:26:04.515  ERROR  tls.issuance.acme.acme_client  challenge failed  {"identifier": "http://fqdncaddy.gridtesting.xyz", "challenge_type": "tls-alpn-01", "problem": {"type": "urn:ietf:params:acme:error:connection", "title": "", "detail": "185.69.167.81: Timeout during read (your server may be slow or overloaded)", "instance": "", "subproblems": []}}
[...]

PR #1843, opened 2022-11-21
Modify the format validation of fqdn gateway backend

sameh-farouk:
These changes Ensure that the backend should be one of:
- URL, which must have a scheme as `HTTP` (in case we configuring HTTP service, atm this happens when tls_passthrough is false)
- Address, which must have a port included (in case we configuring TCP service, atm this happens when tls_passthrough is true)
 
 What's changed:
- Remove the validation that requires the scheme to be `HTTPS` when `tls_passthrough` is true.
- Return an error if the scheme is not empty and `tls_passthrough` is true
- Returns an error if the port wasn't provided  and `tls_passthrough` is true

Related Issue:
https://github.com/threefoldtech/zos/issues/1794
https://github.com/threefoldtech/zos/issues/1842

### DIFF
--- a/pkg/gridtypes/zos/gw_fqdn.go
+++ b/pkg/gridtypes/zos/gw_fqdn.go
@@ -18,18 +18,21 @@ var (
 type Backend string
 
 // check if valid http://ip:port, http://ip or ip:port
-func (b Backend) Valid() error {
+func (b Backend) Valid(tls bool) error {
 	u, err := url.Parse(string(b))
 	if err != nil {
 		return errors.Wrap(err, "failed to parse backend")
 	}
-	if u.Scheme != "" {
-		if u.Scheme != "http" {
-			return fmt.Errorf("invalid scheme expected either http, or empty")
+	if tls {
+		if u.Scheme != "" {
+			return fmt.Errorf("scheme expected to be empty")
+		}
+		if u.Port() == "" {
+			return fmt.Errorf("missing port in backend address")
 		}
 	} else {
-		if u.Port() == "" {
-			return fmt.Errorf("backend missing either scheme or port")
+		if u.Scheme != "http" {
+			return fmt.Errorf("scheme expected to be http")
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/gridtypes/zos/gw_fqdn.go
```diff
@@ -17,14 +17,20 @@ var (
...
-	if u.Scheme != "http" && u.Scheme != "https" {
-		return fmt.Errorf("invalid scheme expected http, or https")
+	if u.Scheme != "" {
+		if u.Scheme != "http" {
+			return fmt.Errorf("invalid scheme expected either http, or empty")
+		}
+	} else {
+		if u.Port() == "" {
+			return fmt.Errorf("backend missing either scheme or port")
+		}
```
muhamadazmy: That's a weird check (nested) why not 
```go
if u.Schema == "" && u.Port() == "" {
   // port is missing
} else if u.Schema != "http" {
   // invalid schema
}
```

pkg/gridtypes/zos/gw_fqdn.go
```diff
@@ -17,14 +17,20 @@ var (
 
 type Backend string
 
-// check if valid http://x.x.x.x:port or [::]:port
+// check if valid http://ip:port, http://ip or ip:port
```
muhamadazmy, line 20: I would pass the `passthrough` flag to the Valid method (may be also make it private method) hence the workload validation will also validate that this backend matches the tls-passthrough requirements. hence give an early validation error when validating the workload before it attempts to deploly its.

Note that the workload validation is done ealry and also done in the client side. hence clients can catch the issue before it's sent to the node

muhamadazmy (changes requested):

pkg/gridtypes/zos/gw_fqdn.go
```diff
@@ -17,14 +17,23 @@ var (
...
-	if u.Scheme != "http" && u.Scheme != "https" {
-		return fmt.Errorf("invalid scheme expected http, or https")
+	if tls {
+		if u.Scheme != "" {
+			return fmt.Errorf("scheme expected to be empty")
+		}
+		if u.Port() == "" {
+			return fmt.Errorf("missing port in backend address")
+		}
+	} else {
```
muhamadazmy: why `else { if { ` when it can be `else if`

muhamadazmy (approved):

PR #1855, opened 2022-11-25
Events stream

muhamadazmy:
## Rewrite event distribution system
This improves on the internal delivery system of the events so instead of using zbus pub/sub streams we use redis streams which is much more powerful as follows:

- Events are scanned from the chain. If the events subsystem is down
  during the delivery of the event, the missed blocsk will be rescanned
- Received events are pushed to capped streams (length of 1024) event
- Consumers groups are created one per consumer. So each group has a max
  if one group. this allows us to track pending messages, and able to
processes un acknowledge messages
- To prvent the node from getting stuck an event ack is sent back if
  handling failed or succeeded. other protection mechanism is
implemented in the system to clean up.


cmds/modules/provisiond/events.go
```diff
@@ -7,21 +7,20 @@ import (
...
 	"github.com/threefoldtech/zos/pkg/provision"
-	"github.com/threefoldtech/zos/pkg/stubs"
 )
 
 type ContractEventHandler struct {
 	node   uint32
 	pool   substrate.Manager
 	engine provision.Engine
-	cl     zbus.Client
+	events *events.RedisConsumer
```
xmonader: shouldn't that be `redisConsumer` instead of events for readability?

xmonader (dismissed):

cmds/modules/provisiond/events.go
muhamadazmy: sure

PR #1879, opened 2023-01-03
Adds cloud-console integration

ashraffouda:

### DIFF
--- a/docs/manual/zmachine/cloud-console.md
+++ b/docs/manual/zmachine/cloud-console.md
@@ -1,0 +1,14 @@
+# Cloud console
+
+- `cloud-console` is a tool to view machine logging and interact with the machine you have deployed
+- It always runs on the machine's private network ip and port number equla to `20000 +last octect` of machine private IP
+- For example if the machine ip is `10.20.2.2/24` this means
+  - `cloud-console` is running on `10.20.2.1:20002`
+- For the cloud-console to run we need to start the cloud-hypervisor with option "--serial pty" instead of tty, this allows us to interact with the vm from another process `cloud-console` in our case
+- To be able to connect to the web console you should first start wireguard to connect to the private network
+
+```
+wg-quick up wireguard.conf
+```
+
+- Then go to your browser with the network router IP `10.20.2.1:20002`
--- a/pkg/vm/ch.go
+++ b/pkg/vm/ch.go
@@ -5,11 +5,11 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"math"
 	"net"
 	"os"
 	"os/exec"
 	"path/filepath"
-	"strconv"
 	"strings"
 	"syscall"
 
@@ -25,20 +25,30 @@ const (
 
 // StartCloudConsole Starts the cloud console for the vm on it's private network ip
 func (m *Machine) StartCloudConsole(ctx context.Context, namespace string, networkAddr net.IPNet, machineIP net.IPNet, ptyPath string) error {
-	ipOctets := strings.Split(machineIP.IP.String(), ".")
-	port, err := strconv.Atoi(ipOctets[len(ipOctets)-1])
-	if err != nil {
-		return err
+	ipv4 := machineIP.IP.To4()
+	if ipv4 == nil {
+		return fmt.Errorf("invalid vm ip address (%s) not ipv4", machineIP.IP.String())
 	}
-	port = 20000 + port
-	if port == 65535 {
+	port := 20000 + uint16(ipv4[3])
+	if port == math.MaxUint16 {
+		// this should be impossible since a byte max value is 512 hence 20_000 + 512 can never be over
+		// max of uint16
 		return fmt.Errorf("couldn't start cloud console port number exceeds %d", port)
 	}
-	args := []string{"setsid", "ip", "netns", "exec", namespace, cloudConsoleBin, ptyPath, networkAddr.IP.String(), strconv.Itoa(port)}
+	args := []string{
+		"setsid",
+		"ip",
+		"netns",
+		"exec", namespace,
+		cloudConsoleBin,
+		ptyPath,
+		networkAddr.IP.String(),
+		fmt.Sprint(port),
+	}
+
 	log.Debug().Msgf("running cloud-console : %+v", args)
 
 	cmd := exec.CommandContext(ctx, "busybox", args...)
-
 	if err := cmd.Start(); err != nil {
 		return errors.Wrap(err, "failed to start cloud-hypervisor")
 	}
--- a/pkg/vm/ch.go
+++ b/pkg/vm/ch.go
@@ -25,8 +25,8 @@ const (
 	cloudConsoleBin = "cloud-console"
 )
 
-// StartCloudConsole Starts the cloud console for the vm on it's private network ip
-func (m *Machine) StartCloudConsole(ctx context.Context, namespace string, networkAddr net.IPNet, machineIP net.IPNet, ptyPath string) error {
+// startCloudConsole Starts the cloud console for the vm on it's private network ip
+func (m *Machine) startCloudConsole(ctx context.Context, namespace string, networkAddr net.IPNet, machineIP net.IPNet, ptyPath string) error {
 	ipv4 := machineIP.IP.To4()
 	if ipv4 == nil {
 		return fmt.Errorf("invalid vm ip address (%s) not ipv4", machineIP.IP.String())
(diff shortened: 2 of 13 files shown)


pkg/vm/ch.go
```diff
@@ -17,9 +19,34 @@ import (
 )
 
 const (
-	chBin = "cloud-hypervisor"
+	chBin           = "cloud-hypervisor"
+	cloudConsoleBin = "cloud-console"
 )
 
+func (m *Machine) StartCloudConsole(ctx context.Context, namespace string, networkAddr net.IPNet, machineIP net.IPNet, ptyPath string) error {
```
xmonader: Exported function missing a comment

pkg/vm/ch.go
```diff
@@ -17,9 +19,34 @@ import (
 )
 
 const (
-	chBin = "cloud-hypervisor"
+	chBin           = "cloud-hypervisor"
+	cloudConsoleBin = "cloud-console"
```
xmonader, line 26: where does this binary come from? is there a PR for that bin?

pkg/vm/ch.go
```diff
@@ -17,9 +19,34 @@ import (
...
 )
 
+func (m *Machine) StartCloudConsole(ctx context.Context, namespace string, networkAddr net.IPNet, machineIP net.IPNet, ptyPath string) error {
+	machineIPLsit := strings.Split(machineIP.IP.String(), ".")
```
xmonader: machineIPLsit is a typo, not sure if you mean the ipOctets that are separated by dots if so please use ipOctets instead?

pkg/vm/ch.go
```diff
@@ -17,9 +19,34 @@ import (
...
+	port, err := strconv.Atoi(machineIPLsit[len(machineIPLsit)-1])
+	if err != nil {
+		return err
+	}
+	port = 20000 + port
```
xmonader: where does the 20000 number come from? and is there a guard against max port used 65535?

pkg/vm/ch.go
```diff
@@ -17,9 +19,34 @@ import (
...
+	args := []string{"setsid", "ip", "netns", "exec", namespace, cloudConsoleBin, ptyPath, networkAddr.IP.String(), strconv.Itoa(port)}
+	log.Debug().Msgf("running cloud-console : %+v", args)
+
+	cmd := exec.CommandContext(ctx, "busybox", args...)
+
+	if err := cmd.Start(); err != nil {
+		return errors.Wrap(err, "failed to start cloud-hypervisor")
+	}
+
+	if err := m.release(cmd.Process); err != nil {
```
xmonader, line 58: you can do return `m.release(cmd.Process)` directly

pkg/vm/ch.go
```diff
@@ -33,7 +60,7 @@ func (m *Machine) Run(ctx context.Context, socket, logs string) error {
 		"--memory": {fmt.Sprintf("%s,shared=on", m.Config.Mem.String())},
 
 		"--console":    {"off"},
-		"--serial":     {"tty"},
+		"--serial":     {"pty"},
```
xmonader: can you please add a comment here why the change to pty

pkg/vm/client.go
```diff
@@ -15,6 +15,12 @@ type Client struct {
 	client http.Client
 }
 
+type VMData struct {
+	CPUs    CPU
```
xmonader: We should use CPU or NumCPU instead of CPUs

pkg/vm/client.go
```diff
@@ -86,22 +92,22 @@ func (c *Client) Resume(ctx context.Context) error {
 }
 
 // Inspect return information about the vm
-func (c *Client) Inspect(ctx context.Context) (CPU, MemMib, error) {
+func (c *Client) Inspect(ctx context.Context) (VMData, error) {
```
xmonader, line 96: Not sure if Inspect is better than Info -specially we are calling vm.info :) - , but if Inspect is used everywhere we can be consistent

pkg/vm/client.go
```diff
@@ -86,22 +92,22 @@ func (c *Client) Resume(ctx context.Context) error {
...
 
 	response, err := c.client.Do(request)
 	if err != nil {
-		return 0, 0, errors.Wrap(err, "error calling machine info")
+		return VMData{}, errors.Wrap(err, "error calling machine info")
 	}
 
 	defer response.Body.Close()
 
 	if response.StatusCode != http.StatusOK {
```
xmonader, line 110: is it possible to have a body that we can also add in the error message along with the status code?

pkg/vm/manager.go
```diff
@@ -471,8 +472,34 @@ func (m *Module) Run(vm pkg.VM) error {
...
+			return errors.Wrapf(err, "failed to get network '%s'", ifc.NetID)
+		}
+
+		client := NewClient(m.socketPath(vm.Name))
+		vmData, err := client.Inspect(ctx)
+		if err != nil {
+			return errors.Wrapf(err, "failed to Inspect vm with id '%s'", vm.Name)
+		}
+
+		err = machine.StartCloudConsole(ctx, namespace, networkAddr, ifc.IPs[0], vmData.PTYPath)
```
xmonader: can you please add a comment on ifc.IPs[0] usage

pkg/vm/manager.go
```diff
@@ -471,8 +472,34 @@ func (m *Module) Run(vm pkg.VM) error {
...
+		if err != nil {
+			return errors.Wrapf(err, "failed to start cloud-console for vm id '%s'", vm.Name)
+		}
+		return nil
```
xmonader: is this meant to do an early return from the loop as soon as you started the cloud console?

xmonader:
I added some minor comments and questions, also please make sure to add docs file that explains the cloud console flow

xmonader (changes requested):
please check the comments

pkg/vm/manager.go
ashraffouda: yes

docs/manual/zmachine/cloud-console.md
```diff
@@ -0,0 +1,14 @@
+# Cloud console
+
+- `cloud-console` is a tool to view machine logging and interact with the machine you have deployed
+- It always runs on the machine's private network ip and port number equla to 20000+last octect of machine private IP
```
xmonader: equla -> equals

pkg/vm/ch.go
xmonader: resolved in bins/packages/cloudconsole/cloudconsole.sh

pkg/vm/ch.go
xmonader: the comment should start with the method name `StartCloudConsole`

docs/manual/zmachine/cloud-console.md
```diff
@@ -0,0 +1,14 @@
+# Cloud console
+
+- `cloud-console` is a tool to view machine logging and interact with the machine you have deployed
+- It always runs on the machine's private network ip and port number equla to `20000 +last octect` of machine private IP
+- For exampleif the machine ip is `10.20.2.2/24` this means
```
xmonader: for example, if the

xmonader (dismissed):

pkg/vm/ch.go
```diff
@@ -17,9 +19,43 @@ import (
 )
 
 const (
-	chBin = "cloud-hypervisor"
+	chBin           = "cloud-hypervisor"
+	cloudConsoleBin = "cloud-console"
 )
 
+// StartCloudConsole Starts the cloud console for the vm on it's private network ip
+func (m *Machine) StartCloudConsole(ctx context.Context, namespace string, networkAddr net.IPNet, machineIP net.IPNet, ptyPath string) error {
```
muhamadazmy: Why this function is defined on a Machine receiver not on the Manager. It does not access any of the Machine properties.

I think it's better for this to become a `private` manager method instead it fits more there.

pkg/vm/client.go
```diff
@@ -86,40 +93,48 @@ func (c *Client) Resume(ctx context.Context) error {
...
+		body, _ := ioutil.ReadAll(response.Body)
+		response.Body.Close()
+
+		return VMData{}, fmt.Errorf("got unexpected http code '%s' on machine info, Response: %s", response.Status, string(body))
 	}
 
 	var data struct {
 		Config struct {
-			CPUs struct {
+			CPU struct {
```
muhamadazmy, line 119: Why not use the `VMData` struct here instead of anonymous  struct then copy everything. Just make sure VMData has the right json annotiations then return data.Config directly

muhamadazmy (changes requested):

muhamadazmy:
BLOCKED ON: https://github.com/LeeSmet/cloud-console/issues/3

muhamadazmy (approved):

PR #1891, opened 2023-02-01
Power management redesign

muhamadazmy:


cmds/modules/powerd/main.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+const (
+	module = "power"
+)
+
+// Module is entry point for module
+var Module cli.Command = cli.Command{
+	Name:  "powerd",
+	Usage: "reports the node total resources",
```
xmonader: does it do that?

cmds/modules/powerd/main.go
```diff
@@ -0,0 +1,112 @@
...
+	ctx, _ := utils.WithSignal(cli.Context)
+	utils.OnDone(ctx, func(_ error) {
+		log.Info().Msg("shutting down")
+	})
+
+	env := environment.MustGet()
+
+	cl, err := zbus.NewRedisClient(msgBrokerCon)
+	if err != nil {
+		return errors.Wrap(err, "fail to connect to message broker server")
```
xmonader: fail -> failed

cmds/modules/powerd/main.go
muhamadazmy: of course not

cmds/identityd/main.go
```diff
@@ -88,7 +88,7 @@ func main() {
 
 	if farm {
 		env := environment.MustGet()
-		fmt.Println(env.FarmerID)
+		fmt.Println(env.FarmID)
```
xmonader, line 91: for that change do we have the relevant docs / bootstrap updated with that change?

cmds/identityd/main.go
```diff
@@ -355,7 +355,7 @@ func getIdentityMgr(root string, debug bool) (pkg.IdentityManager, error) {
 
 	log.Info().
 		Bool("orphan", env.Orphan).
-		Uint64("farmer_id", uint64(env.FarmerID)).
+		Uint64("farmer_id", uint64(env.FarmID)).
```
xmonader, line 358: shouldn't it be `farm_id` now?

cmds/identityd/main.go
muhamadazmy: No, it's only the internal code in zos that changed the name

cmds/identityd/main.go
muhamadazmy: yeah, the problem is the `cmdline` still uses `farmer_id` which is confusing because it should be farm id. So internally in the code i changed that to FarmID but not changing anything else.

pkg/identity/identityd.go
```diff
@@ -102,7 +102,7 @@ func (d *identityManager) Farm() (string, error) {
 	}
 	defer cl.Close()
 
-	farm, err := cl.GetFarm(uint32(d.env.FarmerID))
+	farm, err := cl.GetFarm(uint32(d.env.FarmID))
 	if errors.Is(err, substrate.ErrNotFound) {
 		return "", fmt.Errorf("wrong farm id")
 	} else if err != nil {
```
xmonader, line 108: let's do on a separate if condition no need for the else if here

pkg/utils/mark.go
```diff
@@ -0,0 +1,36 @@
+package utils
```
xmonader, line 1: can we please add some docs on this package (intention/usage)

pkg/power/ethtool.go
```diff
@@ -0,0 +1,69 @@
...
+			return "", err
+		}
+
+		line := strings.TrimSpace(scanner.Text())
+		parts := strings.Split(line, ":")
+		if parts[0] != string(flag) {
+			continue
+		}
+		if len(parts) != 2 {
+			return "", fmt.Errorf("invalid ethtool output format")
```
xmonader: can we include the line for more debugging purposes in the error?

pkg/power/ethtool.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+		return strings.TrimSpace(string(parts[1])), nil
+	}
+
+	return "", ErrFlagNotFound
```
xmonader, line 63: I think this error should be dynamic to include the flag that was requested in the error message?

pkg/power/power.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+
+	nic := filtered[0].Attrs().Name
+	log.Info().Str("nic", nic).Msg("enabling wol on interface")
+	support, err := ValueOfFlag(ctx, nic, SupportsWakeOn)
+
+	if errors.Is(err, ErrFlagNotFound) {
+		// no support for
+		return false, nil
+	} else if err != nil {
```
xmonader, line 96: let's move it outside of the else if

pkg/power/power.go
```diff
@@ -0,0 +1,312 @@
...
+		return fmt.Errorf("can't find mac address of node '%d'", node.ID)
+	}
+
+	return exec.Command("ether-wake", "-i", "zos", mac).Run()
+
+}
+
+func (p *PowerServer) shutdown() error {
+	if !p.enabled {
+		log.Info().Msg("ignoring shutdown because power managed is not enabled")
```
xmonader: power managed -> power management?

pkg/power/power.go
```diff
@@ -0,0 +1,312 @@
...
+		case <-time.After(10 * time.Second):
+		}
+	}
+}
+
+func (p *PowerServer) Start(ctx context.Context) error {
+	return p.events(ctx)
+}
+
+type Direct struct {
```
xmonader: What does this struct mean/do?

pkg/power/power.go
```diff
@@ -0,0 +1,312 @@
...
+	if !strings.Contains(support, string(MagicPacket)) {
+		// no magic packet support either
+		return false, nil
+	}
+
+	return true, SetWol(ctx, nic, MagicPacket)
+}
+
+func (p *PowerServer) syncSelf() error {
+	cl, err := p.sub.Substrate()
```
mariobassem, line 109: substrate connection was never closed?

pkg/utils/mark.go
muhamadazmy: yes

pkg/power/ethtool.go
muhamadazmy: I checked but this is returned when u are trying to get the value of `specific` flag. So when flag not found is returned it means this specific flag u asking for is not found. So when this error is returned, no extra information need to be attached to the error since the caller already know what it was asking

pkg/power/power.go
muhamadazmy: I really think it's more readable with the else if (specially if someone else going to work on this code) instead of multiple `if` 

The reason is i think it's more readable, and probably optimized, is that it works in both cases if you return or not return in the branch. Also normally if u don't return the else if optimized the code so not all of the branches are checked

so 
```go

if cond1 {
}

if cond1 {
}
```
both will be checked  in runtime

but using else if like
```go
if cond1 {
} else if cond2 {
}
```
the compiler knows cond1 and cond2 can't both happen at the same time, hence if cond1 is true, cond2 is not checked.

pkg/power/power.go
muhamadazmy: Old code, will delete

pkg/power/power.go
```diff
@@ -0,0 +1,312 @@
...
+	if event.FarmID != p.farm {
+		return nil
+	}
+
+	log.Debug().
+		Uint32("farm", uint32(p.farm)).
+		Uint32("node", p.node).
+		Msg("received power event for farm")
+
+	cl, err := p.sub.Substrate()
```
mariobassem, line 204: substrate connection never closed.

pkg/power/power.go
```diff
@@ -0,0 +1,312 @@
...
+		if err := p.event(&event); err != nil {
+			log.Error().Err(err).Msg("failed to process power event")
+		}
+	}
+
+	return nil
+}
+
+// start processing time events.
+func (p *PowerServer) events(ctx context.Context) error {
```
mariobassem, line 251: would opening and closing the substrate connection here help a bit??

pkg/power/uptime.go
```diff
@@ -0,0 +1,116 @@
...
+
+	for {
+		err := safeUptime(ctx)
+		if errors.Is(err, context.Canceled) {
+			log.Info().Msg("stop uptime reporting. context cancelled")
+			return
+		} else if err != nil {
+			log.Error().Err(err).Msg("sending uptime failed")
+		} else {
+			// context was cancelled
```
mariobassem, line 110: isn't this the case where uptime is reported successfully?
I'm asking if context is really canceled here or not.

pkg/power/power.go
mariobassem: i'm suggesting this because if there were many power management events, with each event being processed a substrate connection is opened, which may be costly???

pkg/power/power.go
muhamadazmy: Good catch, thanks

pkg/power/power.go
muhamadazmy: while the node receive all power events, not all events reaches here, only events intended to the node farm are actually propagated to this code

muhamadazmy: So instead of holding an open connection all the time it's actually better to only open when needed which should not be many

pkg/power/uptime.go
muhamadazmy: No, if u follow the code you will see that the `u.uptime()` actually starts an infinite loop that only returns succefully if the ctx is cancelled and the node is shutdown. Any other error means we need to start this infinite loop again

xmonader (approved):

mariobassem (approved):

PR #1903, opened 2023-02-15
also return system reserved capacity

muhamadazmy:


docs/manual/api.md
```diff
@@ -63,6 +63,8 @@ Capacity {
 }
 ```
 
+> Note that, `used` capacity equal the full workload reserved capacity PLUS the system reserved capacity
```
brandonpille, line 66: Nitpicking: capacity equals the full

brandonpille (approved):
Looks great. Thanks for the quick update.

Issue #1906, opened 2023-02-19
Cannot Deploy a gateway name while tls_passthrough is set to true


maayarosama:
Whenever I try to deploy a gateway name with a tls_passthrough set to true it fails giving me a ```failed to parse backend: parse "[300:170b:469a:2ed8:788f:6dd6:b647:8ed6]:443": first path segment in URL cannot contain colon``` error 

![gateway](https://user-images.githubusercontent.com/66883096/219958542-d8598be2-d4a7-49b5-8bed-6d7e82680f7a.png)

My GatewayNameModel setup is as follows:
- The planetary IP belongs to one of my deployed Vms. 
```bash
 const gw = new GatewayNameModel();
    gw.name = "test";
    gw.node_id = +(await grid3.capacity.filterNodes(gatewayQueryOptions))[0].nodeId;
    gw.tls_passthrough = true;
    gw.backends = ["[300:170b:469a:2ed8:788f:6dd6:b647:8ed6]:443"];
```

xmonader:
@rawdaGastan is this the one that needs to be `//IP:443`  same as https://github.com/threefoldtech/zos/issues/1886 ? if so @maayarosama please change the backends tline to be
```
    gw.backends = ["//[300:170b:469a:2ed8:788f:6dd6:b647:8ed6]:443"];
```

rawdaGastan:
yes same issue and it will work with @xmonader suggestion

muhamadazmy:
@rawdaGastan I was discussing the correct fix for this issue with @xmonader . Let me know if he explained to you otherwise I can re-explain

xmonader:
closed as duplicate of https://github.com/threefoldtech/zos/issues/1886

Issue #1886, opened 2023-01-22
validate a backend


rawdaGastan:
case: `ip:port` fails when validating a backend because it can't parse `ip:port` as a url 

![Screenshot from 2023-01-22 12-22-55](https://user-images.githubusercontent.com/47260239/213912339-6f88f60e-0ccb-4ad4-8b81-ae772b008f67.png)

example for `1.1.1.1:10`

![Screenshot from 2023-01-22 12-56-15](https://user-images.githubusercontent.com/47260239/213912375-c934915d-0291-4f1e-ba52-43a67c1ddb44.png)

muhamadazmy:
When tls passthrough is enabled that backend must always be in the `ip:port` format. The `ip` must be a valid ip v4 or v6. it tls passthrough is must be a full url in the format `http://ip:port` also the ip part must be a valid ipv4 or v6.

muhamadazmy:
Related to #1906

xmonader:
@maayarosama please verify on devnet your issue no need to add `//` now to the url

PR #1908, opened 2023-02-21
validate backend

rawdaGastan:
## Related issues

- https://github.com/threefoldtech/zos/issues/1886
- https://github.com/threefoldtech/zos/issues/1906

### DIFF
--- a/pkg/gateway/gateway.go
+++ b/pkg/gateway/gateway.go
@@ -564,19 +564,13 @@ func (g *gatewayModule) setupRouting(wlID string, fqdn string, backends []string
 	}
 	servers := make([]Server, len(backends))
 	for idx, backend := range backends {
+		if err := zos.Backend(backend).Valid(TLSPassthrough); err != nil {
+			return errors.Wrapf(err, "failed to validate backend '%s'", backend)
+		}
 		if TLSPassthrough {
-			host, err := zos.Backend(backend).Parse(true)
-			log.Debug().Str("hostname", host).Str("backend", backend).Msg("tls passthrough")
-			if err != nil {
-				return errors.Wrap(err, "couldn't parse backend host")
-			}
-			servers[idx] = Server{
-				Address: host,
-			}
+			servers[idx] = Server{Address: backend}
 		} else {
-			servers[idx] = Server{
-				Url: backend,
-			}
+			servers[idx] = Server{Url: backend}
 		}
 	}
 	route := fmt.Sprintf("%s-route", wlID)
--- a/pkg/gateway/gateway.go
+++ b/pkg/gateway/gateway.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"net"
-	"net/url"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -566,16 +565,13 @@ func (g *gatewayModule) setupRouting(wlID string, fqdn string, backends []string
 	servers := make([]Server, len(backends))
 	for idx, backend := range backends {
 		if TLSPassthrough {
-			if err := zos.Backend(backend).Valid(true); err != nil {
-				return errors.Wrap(err, "couldn't validate backend host")
-			}
-			u, err := url.Parse(backend)
-			log.Debug().Str("hostname", u.Host).Str("backend", backend).Msg("tls passthrough")
+			host, err := zos.Backend(backend).Parse(true)
+			log.Debug().Str("hostname", host).Str("backend", backend).Msg("tls passthrough")
 			if err != nil {
 				return errors.Wrap(err, "couldn't parse backend host")
 			}
 			servers[idx] = Server{
-				Address: u.Host,
+				Address: host,
 			}
 		} else {
 			servers[idx] = Server{
--- a/pkg/gridtypes/zos/gw_fqdn.go
+++ b/pkg/gridtypes/zos/gw_fqdn.go
@@ -20,45 +20,44 @@ var (
 type Backend string
 
 // Parse accepts http://ip:port, http://ip or ip:port
-// returns error if backend is invalid and returns hostname if it is valid
-// parses backend string if it is a valid string based on the tlsPassthrough parameter
+// checks if backend string is a valid string based on the tlsPassthrough parameter
 // ip:port is only valid in case of tlsPassthrough is true
 // http://ip:port or http://ip is valid in case of tlsPassthrough is false
-func (b Backend) Parse(tlsPassthrough bool) (string, error) {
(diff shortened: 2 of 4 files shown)


pkg/gridtypes/zos/gw_fqdn.go
```diff
@@ -19,24 +20,41 @@ type Backend string
 
 // check if valid http://ip:port, http://ip or ip:port
```
xmonader: let's expand the docstring to explain the behavior

pkg/gridtypes/zos/gw_fqdn.go
```diff
@@ -19,24 +20,41 @@ type Backend string
...
-	if err != nil {
-		return errors.Wrap(err, "failed to parse backend")
-	}
+	var hostName string
 	if tls {
-		if u.Scheme != "" {
-			return fmt.Errorf("scheme expected to be empty")
+		host, port, err := net.SplitHostPort(string(b))
+		if err != nil {
+			return fmt.Errorf("failed to parse backend with error: %w", err)
```
xmonader: inject the passed backend in the message

pkg/gridtypes/zos/gw_fqdn.go
```diff
@@ -19,24 +20,41 @@ type Backend string
...
+		}
+		if port == "" {
+			return fmt.Errorf("missing port in backend")
```
xmonader: inject the backend and include the expected format in the error message

pkg/gridtypes/zos/gw_fqdn.go
```diff
@@ -19,24 +20,41 @@ type Backend string
...
 		}
-		if u.Port() == "" {
-			return fmt.Errorf("missing port in backend address")
+
+		parsedPort, err := strconv.ParseInt(port, 10, 64)
+		if err != nil {
+			return fmt.Errorf("invalid port in backend: %s", port)
+		}
+
+		if parsedPort > 65535 || parsedPort < 0 {
```
xmonader: flip the OR'd items to be parsedPort < 0 || parsedPort > 65535

pkg/gridtypes/zos/gw_fqdn_test.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestValidBackend(t *testing.T) {
+	require := require.New(t)
+
+	t.Run("test ip:port cases", func(t *testing.T) {
```
xmonader, line 12: expand with IPv6 cases please

pkg/gridtypes/zos/gw_fqdn.go
```diff
@@ -19,24 +20,41 @@ type Backend string
...
-	}
+	var hostName string
 	if tls {
-		if u.Scheme != "" {
-			return fmt.Errorf("scheme expected to be empty")
+		host, port, err := net.SplitHostPort(string(b))
+		if err != nil {
+			return fmt.Errorf("failed to parse backend with error: %w", err)
+		}
+		if port == "" {
```
muhamadazmy: This is not necessary, if you check the code for the SplitHostPort function you see that it already return error if the format is wrong or if the port is empty.

But it does validate that the IP value or the Port value are valid. hence parsing of the splited ip, and port still need to be done of course

pkg/gridtypes/zos/gw_fqdn.go
```diff
@@ -19,24 +20,41 @@ type Backend string
...
+			return fmt.Errorf("missing port in backend")
 		}
-		if u.Port() == "" {
-			return fmt.Errorf("missing port in backend address")
+
+		parsedPort, err := strconv.ParseInt(port, 10, 64)
```
muhamadazmy: use `ParseUint` instead, since a negative value is illegal.

pkg/gridtypes/zos/gw_fqdn.go
muhamadazmy: Once possiblity of negative value is omitted, you can only validate the value is less than or equal `math.MaxUint16` please use constants, not hardcoded values

muhamadazmy (changes requested):

xmonader (dismissed):

muhamadazmy (changes requested):
Okay, this looks much better. Except this is only half the work. This only fix the validation issue. but in the code where the backend actually being used is still assuming it's always a URL. 

Please check the code here 
https://github.com/threefoldtech/zos/blob/main/pkg/gateway/gateway.go#L569

If you need to change few things so the validation and processing of the Backend value is not replicated everywhere.

I suggest replace the Valid function with instead some another function (say `Parse(tls bool)`, or something else that return `(string, error)` where the error is return if it's not valid, but the string is the value that need to be used as backend in the configuration

pkg/gridtypes/zos/gw_fqdn.go
```diff
@@ -17,28 +19,46 @@ var (
...
+			return "", fmt.Errorf("invalid port in backend: %s", port)
 		}
-	} else if u.Scheme != "http" {
-		return fmt.Errorf("scheme expected to be http")
+
+		if parsedPort > math.MaxUint16 {
+			return "", fmt.Errorf("port '%s' must be <= 65535", port)
+		}
+
+		hostName = host
```
muhamadazmy, line 43: this is wrong because you need to return `ip:port` in case of tls, or `http://ip:port` in case of no tls

pkg/gateway/gateway.go
```diff
@@ -566,16 +565,13 @@ func (g *gatewayModule) setupRouting(wlID string, fqdn string, backends []string
 	servers := make([]Server, len(backends))
 	for idx, backend := range backends {
 		if TLSPassthrough {
-			if err := zos.Backend(backend).Valid(true); err != nil {
-				return errors.Wrap(err, "couldn't validate backend host")
-			}
-			u, err := url.Parse(backend)
-			log.Debug().Str("hostname", u.Host).Str("backend", backend).Msg("tls passthrough")
+			host, err := zos.Backend(backend).Parse(true)
```
muhamadazmy: this is still wrong because instead it can be like 
```go 
for idx, backend := range backends {
   u, err := zos.Backend(backend).Parse(TLSPassthrout)
    // check error
   if TLSPassthrough {
   servers[ids] = Server{ Address: u} 
   } else {
   servers[idx] = Server{ Url: u} 
  } 
}

```

muhamadazmy (changes requested):

muhamadazmy (approved):

PR #1910, opened 2023-02-22
return pools capacity

muhamadazmy:
Add an external api over rmb to return the node pools, and their total size and used capacity for each pool. the idea is to make it easier to do capacity planning knowing the available space per pool

### DIFF
--- a/pkg/gridtypes/workload.go
+++ b/pkg/gridtypes/workload.go
@@ -194,6 +194,10 @@ func (w *Workload) Challenge(i io.Writer) error {
 		return err
 	}
 
+	if _, err := fmt.Fprintf(i, "%s", w.Name); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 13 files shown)


pkg/gridtypes/workload.go
```diff
@@ -194,6 +194,10 @@ func (w *Workload) Challenge(i io.Writer) error {
 		return err
 	}
 
+	if _, err := fmt.Fprintf(i, "%s", w.Name); err != nil {
```
xmonader, line 197: I checked on main branch, but it seems to be new. Isn't that a breaking change? won't that need an update in clients?

muhamadazmy: yes yes, i already talked to Hanafy and explained to him and he said it's okay and he will prepare the fix from his side

xmonader: okay CC @ashraffouda / @rawdaGastan please check if you need to update anything as well

PR #1913, opened 2023-02-24
Qsfs autobuild

muhamadazmy:
Autobuild qsfs flist from the docker image from `quatntum-storage` this way newer version of qsfs can be easily integrated and used in the system

### DIFF
--- a/bins/bins-extra.sh
+++ b/bins/bins-extra.sh
@@ -265,7 +265,7 @@ exclude_libs() {
 
 github_name() {
     # force github print
-    echo "name=${1}" >> $GITHUB_OUTPUT
+    echo "::set-output name=name::${1}"
     echo "[+] github exported name: ${1}"
 }
 
--- a/bins/bins-extra.sh
+++ b/bins/bins-extra.sh
@@ -144,6 +144,62 @@ download_file() {
     checksum ${output} $2 || false
 }
 
+download_docker_image() {
+    # set extra filename output or default output
+    if [ ! -z $3 ]; then
+        output=$3
+    else
+        output=$(basename "$1").tar
+    fi
+
+    # set default url
+    fileurl=$1
+
+    docker pull "${fileurl}"
+    docker save "${fileurl}" -o "${output}"
+}
+
+extract_docker_image() {
+    # you need to run `docker save <image> -o <file>.tar
+    # then run this script like `extrat-image.sh <file>.tar`
+    # it will create a directoyr `<file>.tar-root` which containers
+    # all files from the image extracted
+
+    input=$1
+    rootfs=$2
+    if [ ! -f "${input}" ]; then
+        echo "file does not exist"
+        exit 1
+    fi
+
+    tmpdir=".${input}"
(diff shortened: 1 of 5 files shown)


bins/bins-extra.sh
```diff
@@ -144,6 +144,62 @@ download_file() {
...
+    # set default url
+    fileurl=$1
+
+    docker pull "${fileurl}"
+    docker save "${fileurl}" -o "${output}"
+}
+
+extract_docker_image() {
+    # you need to run `docker save <image> -o <file>.tar
+    # then run this script like `extrat-image.sh <file>.tar`
```
xmonader: extrat -> extract

bins/bins-extra.sh
```diff
@@ -144,6 +144,62 @@ download_file() {
...
+    # it will create a directoyr `<file>.tar-root` which containers
```
xmonader: directoyr -> directory, which -> where

bins/packages/qsfs/qsfs.sh
```diff
@@ -0,0 +1,30 @@
+QSFS_VERSION="0.2.0-rc1"
```
xmonader: v0.3.0 was released https://github.com/threefoldtech/0-stor_v2/releases/tag/v0.3.0 for 0-stor does that reflect on qsfs version?

bins/packages/qsfs/qsfs.sh
```diff
@@ -0,0 +1,30 @@
...
+install_qsfs() {
+    echo "[+] install qsfs"
+
+    mkdir -p "${ROOTDIR}"
+
+    extract_docker_image "${DISTDIR}/qsfs-${QSFS_VERSION}.tar" "${ROOTDIR}"
+}
+
+build_qsfs() {
+    pushd "${DISTDIR}"
```
xmonader, line 23: variable should be DESTDIR instead of DISTDIR

bins/packages/qsfs/qsfs.sh
muhamadazmy: well, that's the version used by quatnum-storage https://github.com/threefoldtech/quantum-storage/blob/qsfs-docker/docker/Dockerfile#L4

bins/packages/qsfs/qsfs.sh
muhamadazmy: why? this variable is set by the base build script in the project and is used everywhere (in all packages) so changing that here means i need to change it everywhere 
also i think distdir is from (distribution dir) so it is correct

bins/bins-extra.sh
muhamadazmy: I will re-write the entire description because when i started i had something else in mind

xmonader (approved):

bins/packages/qsfs/qsfs.sh
muhamadazmy: Updated, i had to rebuild qsfs docker to use latest version of zstor

Issue #1877, opened 2023-01-03
add new commands on the statistics module to return total number of deployments and workloads 


xmonader:
add two new commands to the statistics e.g 

- zos.statistics.deployments_count
- zos.statistics.workloads_count

muhamadazmy:
I don't mind, but what is that needed for?

muhamadazmy:
instead of adding two commands, the statisitcs.get command was extended to return the extra information needed

PR #1916, opened 2023-03-05
add statistics counters

rawdaGastan:
## Issues
- https://github.com/threefoldtech/zos/issues/1877

### DIFF
--- a/pkg/primitives/statistics.go
+++ b/pkg/primitives/statistics.go
@@ -60,9 +60,21 @@ func NewStatistics(total gridtypes.Capacity, storage provision.Storage, reserved
 	}
 }
 
+// deploymentsCount Gets count of deployments
+func (s *Statistics) deploymentsCount() (int, error) {
+	_, deps, _, err := s.storage.Capacity()
+	return len(deps), err
+}
+
+// workloadsCount Gets count of workloads for each deployment
+func (s *Statistics) workloadsCount() (int, error) {
+	_, _, workloads, err := s.storage.Capacity()
+	return len(workloads), err
+}
+
 // Get all used capacity from storage + reserved
 func (s *Statistics) active() (gridtypes.Capacity, error) {
-	cap, _, err := s.storage.Capacity()
+	cap, _, _, err := s.storage.Capacity()
 	cap.Add(&s.reserved)
 	return cap, err
 }
@@ -175,6 +187,8 @@ func NewStatisticsMessageBus(router rmb.Router, stats *Statistics) error {
 func (s *statisticsMessageBus) setup(router rmb.Router) error {
 	sub := router.Subroute("statistics")
 	sub.WithHandler("get", s.getCounters)
+	sub.WithHandler("deployments_count", s.deploymentsCount)
+	sub.WithHandler("workloads_count", s.workloadsCount)
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


pkg/provision/storage/storage.go
```diff
@@ -719,7 +720,7 @@ func (b *BoltStorage) Capacity() (cap gridtypes.Capacity, active []gridtypes.Dep
 		}
 	}
 
-	return cap, active, nil
+	return cap, active, workloads, nil
```
muhamadazmy: That's duplicate data because the deployments include all the workloads anyway. So may be instead return the `number` of active workloads instead of list of active workloads.

pkg/provision/interface.go
```diff
@@ -123,7 +123,7 @@ type Storage interface {
 	// ByTwin return list of deployments for a twin
 	ByTwin(twin uint32) ([]uint64, error)
 	// return total capacity and active deployments
-	Capacity() (cap gridtypes.Capacity, active []gridtypes.Deployment, err error)
+	Capacity() (cap gridtypes.Capacity, active []gridtypes.Deployment, workloads []gridtypes.Workload, err error)
```
muhamadazmy: Check the other comment on the implementation

pkg/primitives/statistics.go
```diff
@@ -60,9 +60,21 @@ func NewStatistics(total gridtypes.Capacity, storage provision.Storage, reserved
...
+	_, deps, _, err := s.storage.Capacity()
+	return len(deps), err
+}
+
+// workloadsCount Gets count of workloads for each deployment
+func (s *Statistics) workloadsCount() (int, error) {
+	_, _, workloads, err := s.storage.Capacity()
+	return len(workloads), err
+}
+
```
muhamadazmy: the `storage.Capacity()` is kinda a heavy call. And we try to avoid it as much as possible. So why not create a single call to it instead that return the counts instead 
```
return len(deps), workloads, err 
```

pkg/primitives/statistics.go
```diff
@@ -175,6 +187,8 @@ func NewStatisticsMessageBus(router rmb.Router, stats *Statistics) error {
 func (s *statisticsMessageBus) setup(router rmb.Router) error {
 	sub := router.Subroute("statistics")
 	sub.WithHandler("get", s.getCounters)
+	sub.WithHandler("deployments_count", s.deploymentsCount)
+	sub.WithHandler("workloads_count", s.workloadsCount)
```
muhamadazmy: This means that the user of those counters need to make 2 calls to get the node statistics! it's a lot of round trips for related data. 

Instead please extend the getCounters function to return all node statistics. it already return `total`, `used`, and `system` can also have a sub object 
```json
"total": ...,
"used": ...,
"system": ...,
"user": {
    "deployments": <number>,
    "workloads": <number>,
}
```

this way the caller can get all node statistics in a single call

muhamadazmy (changes requested):

pkg/provision/interface.go
```diff
@@ -123,7 +123,7 @@ type Storage interface {
 	// ByTwin return list of deployments for a twin
 	ByTwin(twin uint32) ([]uint64, error)
 	// return total capacity and active deployments
-	Capacity() (cap gridtypes.Capacity, active []gridtypes.Deployment, err error)
+	Capacity() (cap gridtypes.Capacity, active []gridtypes.Deployment, workloads int, err error)
```
muhamadazmy: This returned argument set is becoming too long. May be create a `StorageCapacity` struct (in this file) and then change the interface to 
```go
Capacity() (StorageCapacity, error)
```

pkg/primitives/statistics.go
```diff
@@ -60,16 +60,16 @@ func NewStatistics(total gridtypes.Capacity, storage provision.Storage, reserved
 	}
 }
 
-// Get all used capacity from storage + reserved
-func (s *Statistics) active() (gridtypes.Capacity, error) {
-	cap, _, err := s.storage.Capacity()
+// Get all used capacity from storage + reserved / deployments count and workloads count
+func (s *Statistics) active() (gridtypes.Capacity, int, int, error) {
```
muhamadazmy: returning an `int` and `int` is confusing because it's not clear what they are without given names. Also the return argument list become too long imho 🤷🏼‍♂️ 

there is another comment about the return of Storage.Capacity() once it's done, this function can just get that, make sure to add the reserved to the capacity then return the same structure

muhamadazmy (changes requested):

muhamadazmy (approved):

PR #1924, opened 2023-03-14
Gateway private network

### LINKED ISSUE #1674: proxy gateway traffic into private networks (over wireguard)
Currently the playground seems to deploy in a way such that gateways and workloads are connected over yggdrasil. I assume because there is no (easy?) way to proxy from the gateway into the private network. Since yggdrasil has pretty bad performance, we should see if it is possible to use private networks instead.

muhamadazmy:
Fixes #1674

- Add optional `network`. If set the gateway can then proxy traffic to that network. Which means user can gateway traffic to a private workload that lives inside this network.
- Backward compatible with older configurations, user can still use yggdrasil backends although not recommended for reliability  
- Fix few issues in gateway start up. specially after node reboot.

### DIFF
--- a/pkg/gateway/gateway.go
+++ b/pkg/gateway/gateway.go
@@ -37,6 +37,10 @@ const (
 	httpCertResolver = "resolver"
 	dnsCertResolver  = "dnsresolver"
 	validationPeriod = 1 * time.Hour
+
+	configDir = "proxy"
+	metaDir   = "traefik"
+	zinitDir  = "zinit"
 )
 
 var (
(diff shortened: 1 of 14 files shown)


pkg/gateway/gateway.go
```diff
@@ -31,29 +33,36 @@ var (
...
+	volatile string
 	cl       zbus.Client
 	resolver *net.Resolver
 	sub      substrate.Manager
 	// maps domain to workload id
 	reservedDomains map[string]string
 	domainLock      sync.RWMutex
 
-	proxyConfigPath  string
+	//proxyConfigPath  string
```
LeeSmet: Leftover commented code

LeeSmet (dismissed):
LGTM

PR #1925, opened 2023-03-20
Resize cache as needed

### LINKED ISSUE #1781: Revisit reserved cache and memory size
Currently zos reserve 100g of the ssd storage, we need to revise this value because it's too much. 

Also revise the amount of reserved memory for the system

muhamadazmy:
Instead of reserving a 100G of ssd for cache. instead we make this dynamic, in increments of 5G. The cache also can shrink if needed.

Fixes #1781

### DIFF
--- a/pkg/storage/filesystem/filesystem.go
+++ b/pkg/storage/filesystem/filesystem.go
@@ -8,8 +8,32 @@ import (
 
 // Usage struct (in bytes)
 type Usage struct {
+	// Size is allocated space for this Volume
+	// if 0 it means it has no limit.
+	// if it has no-limit, the Used attribute
+	// will be the total size of actual files
+	// inside the volume. It also means the Used
+	// can keep growing to the max possible which
+	// is the size of the pool
 	Size uint64
+	// Used can be one of 2 things:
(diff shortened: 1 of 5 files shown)


pkg/storage/filesystem/filesystem.go
```diff
@@ -8,8 +8,32 @@ import (
...
+	// because it means we can't allocate over that.
+	//
+	// NOTE: Special case, if this is a `zdb` volume the Used is instead
+	// have the total size of reserved namespaces in that volume
 	Used uint64
+
+	// In case of `limited` volume (with quota) Excl will have a "guessed"
+	// value of the total used space by files. This value is not accurate
+	// and Used should be used instead for all capacity planning.
+	Excl uint64
```
xmonader, line 36: What does Excl stand for?

muhamadazmy: It's a `btrfs` terminology. stands for exclusive and tbh, i know it's a cryptic name for what it does but i didn't know what else to call it. 

image

I thought using the same terminology of what this value is and where it comes from might be less confusing in the future.

image

xmonader (approved):

PR #1942, opened 2023-04-03
Adds clean vms that is running with no active workload

ashraffouda:


pkg/vm/monitor.go
```diff
@@ -118,9 +119,17 @@ func (m *Module) monitor(ctx context.Context) error {
 }
 
 func (m *Module) monitorID(ctx context.Context, running map[string]Process, id string) error {
+	stub := stubs.NewProvisionStub(m.client)
 	log := log.With().Str("id", id).Logger()
 
-	if _, ok := running[id]; ok {
+	if ps, ok := running[id]; ok {
+		state, err := stub.GetWorkloadStatus(ctx, id)
```
muhamadazmy: can we validate that the `error` returned is related to either `deployment` or `workload` don't exist. the engine storage.Current returns a specific error type. 
I usually don't like checking the error by string matching. So I am wondering if we can change the GetWorkloadStatus to return an indication that the status does not exist for example that is different from an error 

for example GetWorkloadStatus() ( *Status, error) where it can return nil status if workload does not exist and error only in case of an actual error (which can be temporary errors)

muhamadazmy: Note, i am not sure if returning pointers works on `zbus` if it doesn't we can return `(status, bool, error)` instead

pkg/vm/monitor.go
```diff
@@ -118,9 +119,17 @@ func (m *Module) monitor(ctx context.Context) error {
...
+		// incase an error this means this workload doesn't exist
+		if err != nil || state == gridtypes.StateDeleted || state == gridtypes.StateError {
```
muhamadazmy: state object has method `IsAny()` so can be used here as `state.IsAny(StateDeleted, StateError)`

pkg/provision/engine.go
```diff
@@ -952,3 +952,17 @@ func (e *NativeEngine) DecommissionCached(id string, reason string) error {
...
+
+// GetWorkloadStatus get workload status
+func (e *NativeEngine) GetWorkloadStatus(id string) (gridtypes.ResultState, error) {
+	globalID := gridtypes.WorkloadID(id)
+	twin, dlID, name, err := globalID.Parts()
+	if err != nil {
+		return "", err
+	}
+	wl, err := e.storage.Current(twin, dlID, name)
+	if err != nil {
```
muhamadazmy: Check the comment below, but we can check if this error is `errors.Is()` against multiple possible errors returned from the the storage.Current() (check the code) 

I suggest something like `storage.IsNotFound(err)` that checks errors.Is() against DeploymentNotFound, WorkloadNotFound and TransactionNotFound errors internally

muhamadazmy (changes requested):

pkg/provision/engine.go
```diff
@@ -952,3 +952,21 @@ func (e *NativeEngine) DecommissionCached(id string, reason string) error {
...
+		if e.storage.IsNotFoundError(err) {
```
muhamadazmy: This should not be on the insterface. when i said `storage.IsNotFound()` i meant on the storage package so you can call this function as `storage.IsNotFound(err)` directly !

pkg/provision/interface.go
```diff
@@ -133,6 +133,8 @@ type Storage interface {
 	ByTwin(twin uint32) ([]uint64, error)
 	// return total capacity and active deployments
 	Capacity() (StorageCapacity, error)
+	// check if the given error is not found error(ErrWorkloadNotExist, ErrDeploymentNotExists, ErrTransactionNotExist)
+	IsNotFoundError(err error) bool
```
muhamadazmy: Please remove

pkg/provision/storage/storage.go
```diff
@@ -773,3 +773,10 @@ func (b *BoltStorage) CleanDeleted() error {
 
 	return nil
 }
+
+func (b *BoltStorage) IsNotFoundError(err error) bool {
```
muhamadazmy:
```suggestion
func IsNotFound(err error) bool {
```

pkg/provision/engine.go
muhamadazmy: also No need for nested checks it can be instead like
```go
if storage.IsNotFound(err) {
   // not found error
} else if err != nil {
   // another error
}

// no error
```

pkg/vm/monitor.go
```diff
@@ -118,9 +119,20 @@ func (m *Module) monitor(ctx context.Context) error {
...
 	log := log.With().Str("id", id).Logger()
 
-	if _, ok := running[id]; ok {
+	if ps, ok := running[id]; ok {
+		state, err := stub.GetWorkloadStatus(ctx, id)
+		if err != nil {
+			return errors.Wrapf(err, "failed to get workload status for vm:%s ", id)
+		}
+		// state = nil when workload doesn't exist
+		if state == "" || state.IsAny(gridtypes.StateDeleted, gridtypes.StateError) {
```
muhamadazmy: I don't like having an empty state as a `not found` indication since it's not explicit either return an extra bool, or return a completey new struct 
```go
type OptionStatus struct{
   Exists bool
   State ResultState
}
```

pkg/provision.go
```diff
@@ -12,6 +12,7 @@ import (
 // Provision interface
 type Provision interface {
 	DecommissionCached(id string, reason string) error
+	GetWorkloadStatus(id string) (gridtypes.ResultState, bool, error)
```
muhamadazmy, line 16: Can we add a doc string on top of this to clarify what are the returned values (what the bool is for someone who is reading the code for the first time)

xmonader:
@ashraffouda please make sure to continue on this one sunday please

muhamadazmy (approved):

PR #1944, opened 2023-04-03
fix gateway regex in name validation

### LINKED ISSUE #1943: gateway name & fqdn validation is incorrect?
names in name contracts are validated on the chain [here](https://github.com/threefoldtech/tfchain/blob/e3868e9e6e533e83fa4babedf3596b780b6a3031/substrate-node/pallets/pallet-smart-contract/src/name_contract.rs#L46)
which doesn't match the validation in zos done by the regex [here](https://github.com/threefoldtech/zos/blob/main/pkg/gridtypes/zos/gw_fqdn.go#L12)

should the regex be updated?

mariobassem:
Fixes #1943

### DIFF
--- a/pkg/gridtypes/zos/gw_fqdn.go
+++ b/pkg/gridtypes/zos/gw_fqdn.go
@@ -8,7 +8,7 @@ import (
 	"github.com/threefoldtech/zos/pkg/gridtypes"
 )
 
-var FQDNRegex = regexp.MustCompile(`^([a-zA-Z0-9-_]+\.)+[a-zA-Z0-9-_]{2,}$`)
+var fqdnRegex = regexp.MustCompile(`^([a-zA-Z0-9-_]+\.)+[a-zA-Z0-9-_]{2,}$`)
 
 // GatewayFQDNProxy definition. this will proxy name.<zos.domain> to backends
 type GatewayFQDNProxy struct {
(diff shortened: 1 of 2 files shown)


muhamadazmy (dismissed):

pkg/gridtypes/zos/gw_fqdn.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 var (
-	gwNameRegex = regexp.MustCompile(`^[a-zA-Z0-9-.]+$`)
+	gwNameRegex = regexp.MustCompile(`^[a-zA-Z0-9-_.]+$`)
```
muhamadazmy: There is a big issue here that is not related to your fix. It's nice that you add the `_` that was indeed a bug. but the bigger problem here is that the same regex is used for both NAME and FQDN validation which is not right! 

A name must not have a `.` (dot) because a name is going to be suffixed by the node domain. So a name can be `mario` or `mario-bassem` but NOT `mario.bassem`.

On the other hand a FQDN must be a valid domain name hence again this regex is invalid because this allow the dot to appear ANY where in the domain name like `.mario.bassem` or `mario..bassem` or `marrio.bassem..` which are all wrong.

IMHO nameRegex should be `^[a-zA-Z0-9-_]+$` same as above but without the dot
the fqdnRegex then should be  `^([a-zA-Z0-9-_]+\.)+[a-zA-Z0-9-_]{2,}$`
the fqdnRegex means the pattern for the name can be repeated multiple times separated by `.` but must end up with a min of 2 characters. so `mario.bassem.com` works but `mario.bassem.` or `mario.basse.m` won't work

muhamadazmy (changes requested):
Check comment

muhamadazmy (approved):

Issue #1952, opened 2023-04-20
ZDBResult type has no json tags


mariobassem:
https://github.com/threefoldtech/zos/blob/ab8fd8cf145ae8ec30f3d5f167e920f22f0f7398/pkg/gridtypes/zos/zdb.go#L81-L85

PR #1953, opened 2023-04-20
adds json tags for ZDBResult struct

mariobassem:
## Related issues:

- #1952


xmonader (changes requested):
Looks good, however this is a breaking change for all of the clients in typescript and and go, won't merge until there're ready PRs in both of the sdks

muhamadazmy (approved):

muhamadazmy:
Please make sure that @xmonader comment is handled before merging. I checked in Golang side and this does not break the Go code (decoding data with default tags) but need to be handled in the right way in the other clients

Issue #1682, opened 2022-05-07
Present hardware details on the ZOS-console


weynandkuijpers:
Can we use a small portion of the ZOS console to represent the installed resources?
![image](https://user-images.githubusercontent.com/13766992/167240939-c879a89f-3c8a-4a8f-809e-1be7b8d01caf.png)

I guess this is simple as we do represent percentages of usage of these resources, why not add the quantity asl well 

CPU [available/reserved] : 8/1 etc...

muhamadazmy:
Can you be more specific about the `installed` resources? The idea is that a farmer shouldn't care or worry about what kind of capacity is used by users. But he can still know how many and what contracts are active on his node by checking the block-chain. (get active contracts on a specific node id). But this won't expose the actual workloads (vms, disks, zdbs, etc...) that are actually being used by those contracts.

Hence we think showing utilization of the node is actually better for the farmer (he can tell how heavily his node is used)

Issue #1325, opened 2021-06-30
Show available resources on screen after boot


scottyeager:
Currently, Zos only displays resources in use on the display. Many new farmers are confused by this, and think that their drives or ram is not being recognized.

Propose feature: Show total cru, mru, hru, and sru on the screen after boot.

PR #1971, opened 2023-05-22
show available resources in zui

mariobassem:
## Related issues:

- #1325 
- #1682

### DIFF
--- a/cmds/modules/zui/main.go
+++ b/cmds/modules/zui/main.go
@@ -27,7 +27,7 @@ func (f *signalFlag) Signal() {
 	atomic.SwapInt32((*int32)(f), 1)
 }
 
-//Signaled checks if flag was raised, and lowers the flag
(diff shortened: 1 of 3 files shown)


muhamadazmy:
image

The UI is trimmed because it doesn't fit on the smallest window size. I think you will have to try to compress the tables more. (for example the  but the resources and the usage tables next to each other instead of on top of each other.

muhamadazmy:
The ttysize is 190x43 character

muhamadazmy:
I think the tty height is 24 lines only.

muhamadazmy:
I pushed a commit to actually use table columns instead of formatting everyting in one cell, the `CPU` and `Memory` titles can have their own cell, then the values is shown in a separate cell. then updates need to only change the value in the correct cell

image

this reads easier imho

muhamadazmy (approved):

PR #1973, opened 2023-06-02
Gpu support

### LINKED ISSUE #1972: Implement GPU support
Limitations:
 - We start by supporting only one (and only one) GPU for now, while keeping in mind that more GPU can be available in zos node in the future 

## Changes
- [x] Identify and list all attached GPUs
- [ ] On registration, a node need to update its set of features on the chain (depends on chain change https://github.com/threefoldtech/tfchain/issues/720)
- [x] ZMachine can have an extra `gpu` flag (list of GPUs) 
   - [x] This is only accepted if node is dedicated (otherwise an error)
   - [X] A GPU can be only used by one ZMachine
- [X] Expose the gpu device to the VM (based on research here https://github.com/threefoldtech/zos/issues/1293#issuecomment-1070693016)

muhamadazmy:
Fixes #1972 

- Identify and list available GPU
- Allow Zmachine deployment to define a list of GPU that can be used from the host
- Setup vfio, and friends 
- Pass devices to zmachine.

The code takes into account:
- Node must be rented
- Support for multiple GPUs on the Host (zos)
- Support passing multiple GPUs to the VMs
- Validate that a GPU is not used by another VM
- Make sure all devices that share IOMMU are passed together


client/node.go
```diff
@@ -205,6 +212,12 @@ func (n *NodeClient) Pools(ctx context.Context) (pools []pkg.PoolMetrics, err er
 	return
 }
 
+func (n *NodeClient) GPUs(ctx context.Context) (gpus []GPU, err error) {
```
xmonader, line 215: If the node can have multiple GPUs, the original plan was the user will be reserving the full node + the gpu (i assume it was just one) or can it be multiple ?

muhamadazmy: there can be many GPUs available on one node. To have access to the GPUs you will need to rent the nodes. But once it's rented you can choose which GPU(s) go to which VMs. so you can actually run multiple VMs on the same node each with a sub-set of the GPUs available on th node

pkg/capacity/pci.go
```diff
@@ -0,0 +1,273 @@
...
+			}
+		}
+	}
+
+	return vendors, nil
+}
+
+// GetVendor gets a vendor object given Vendor ID
+func GetVendor(vendor uint16) (Vendor, bool) {
+	v, ok := data[vendor]
```
xmonader, line 113: let's change to `return data[vendor]`

pkg/capacity/pci.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+type Vendor struct {
+	ID      uint16
+	Name    string
+	Devices map[uint16]Device
+}
+
+var (
+	//go:embed pci/pci.ids
```
xmonader, line 41: How do we keep this file always updated?

pkg/capacity/pci_test.go
```diff
@@ -0,0 +1,34 @@
...
+	require.Equal(t, "Advanced Micro Devices, Inc. [AMD/ATI]", vendor.Name)
+	require.Equal(t, "Navi 10 [Radeon RX 5600 OEM/5600 XT / 5700/5700 XT]", device.Name)
+}
+
+func TestListPCI(t *testing.T) {
+	devices, err := ListPCI()
+	require.NoError(t, err)
+
+	for _, device := range devices {
+		fmt.Println(device)
```
xmonader, line 23: Shouldn't we assert here for inclusion of a device in the devices slice?

pkg/capacity/pci_test.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+}
+
+func TestListGPU(t *testing.T) {
+	devices, err := ListPCI(GPU)
+	require.NoError(t, err)
+
+	for _, device := range devices {
+		fmt.Println(device)
```
xmonader, line 32: shouldn't we here assert for gpu devices inclusion in device list instead of debug prints?

pkg/gridtypes/zos/zmachine.go
```diff
@@ -100,6 +101,23 @@ func (m *MachineMount) Challenge(w io.Writer) error {
...
+// Used by a VM, a GPU id is in the format <slot>/<vendor>/<device>
+// This can be queried either from the node features on the chain
+// or listed via the node rmb API.
+// example of a valid gpu definition `0000:28:00.0/1002/731f“
+type GPU string
+
+func (g GPU) Parts() (slot, vendor, device string, err error) {
+	parts := strings.Split(string(g), "/")
+	if len(parts) != 3 {
+		err = fmt.Errorf("invalid GPU id format")
```
xmonader: let's please include the passed in gpu string we tried to split in the error

pkg/capacity/capacity.go
```diff
@@ -173,3 +173,7 @@ func (r *ResourceOracle) GetHypervisor() (string, error) {
 
 	return "", nil
 }
+
+func (r *ResourceOracle) GPUs() ([]PCI, error) {
```
xmonader, line 178: As an exported function I believe it should have a comment

pkg/capacity/pci.go
```diff
@@ -0,0 +1,273 @@
...
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+const (
+	pciDir = "/sys/bus/pci/devices"
+)
+
+type Device struct {
```
xmonader, line 20: Can we document the exported types please

pkg/primitives/vm/gpu.go
```diff
@@ -0,0 +1,150 @@
...
+			if err := os.WriteFile("/sys/bus/pci/drivers_probe", []byte(pci.Slot), 0200); err != nil {
+				return errors.Wrapf(err, "failed to bind device '%s' to vfio", pci.Slot)
+			}
+		}
+	}
+
+	return nil
+}
+
+// expandGPUs expands the set of provided GPUs with
```
xmonader: expand the GPUs with what?

pkg/provision/engine.go
```diff
@@ -262,6 +263,21 @@ func withContract(ctx context.Context, contract substrate.NodeContract) context.
...
+func IsRentedNode(ctx context.Context) bool {
+	v := ctx.Value(rentKey{})
+	if v == nil {
+		return false
+	}
+
+	return v.(bool)
+}
+
+// mark node as rent internall
```
xmonader: What does rent internal mean?

pkg/provision/engine.go
```diff
@@ -577,6 +593,13 @@ func (e *NativeEngine) contract(ctx context.Context, dl *gridtypes.Deployment, n
 	}
 	ctx = withContract(ctx, contract.ContractType.NodeContract)
 
+	rent, err := sub.GetNodeRentContract(e.nodeID)
+	if err != nil && !errors.Is(err, substrate.ErrNotFound) {
+		return nil, fmt.Errorf("failed to check node rent state")
+	}
+
```
xmonader, line 600: We can reach this line in case of error is ErrNotFound,  also I don't understand the logic in error == nil && rent != 0 and how it can be affected with ErrNotFound :(

pkg/capacity/pci.go
muhamadazmy: Unfortunately it's not possible it's Go magic because in go both
```go
v, ok := data[vendor]
// and
v := data[vendor]
```
are valid, so when u try to do `return data[vendor]` it uses the `v := data[vendor]` symatic then fails to compile because the function returns 2 variables.

pkg/capacity/pci.go
muhamadazmy: Good point, i will add a `go:generate` statement that auto-download the file. the `go generate` call still has to be done before building. I will add generate to the CI as well

pkg/capacity/pci_test.go
muhamadazmy: Well, the returned values are not predictable because they completely depends on the machine they run (the actual PCI in the systems, or the GPUs if exists. The prints where mainly for during debugging but there are no solid values i can assert against

pkg/capacity/pci_test.go
muhamadazmy: Same as above

pkg/primitives/vm/gpu.go
muhamadazmy: it's classified 😛

pkg/provision/engine.go
muhamadazmy: I have no idea what i wanted to say, i will update the docs

pkg/provision/engine.go
muhamadazmy: as you said, this only handles any error during the "contract get" call above that is ErrNotFound. so basically return if we have an error that is not ErrNotFound. 

The thing is, if there are NO rent contract for the node the call normally returns `0` but imho this is not correct it instead should return NotFound error. So i am trying to be a little bit more defensive in the code by handling the error correctly.

But anyway, if no error or ErrNotFound error was returned then (in both cases) 0 means No rent, and any other value is rented, hence later i make sure that the node is only marked as rented if and only if there is no error and the rent contract has non zero value.

xmonader (approved):
All looks good to me, thanks @muhamadazmy

PR #1978, opened 2023-06-14
use latest tfchain client and use uptimev2

muhamadazmy:
The uptime v2 also uses a local timestamp hint to make sure that reported uptie is in an acceptable range

### DIFF
--- a/pkg/power/uptime.go
+++ b/pkg/power/uptime.go
@@ -57,7 +57,7 @@ func (u *Uptime) SendNow() (types.Hash, error) {
 		return types.Hash{}, errors.Wrap(err, "failed to get uptime")
 	}
 
-	return sub.UpdateNodeUptime(u.id, uptime)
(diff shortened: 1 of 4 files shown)


pkg/power/uptime.go
```diff
@@ -57,7 +57,7 @@ func (u *Uptime) SendNow() (types.Hash, error) {
 		return types.Hash{}, errors.Wrap(err, "failed to get uptime")
 	}
 
-	return sub.UpdateNodeUptime(u.id, uptime)
+	return sub.UpdateNodeUptimeV2(u.id, uptime, uint64(time.Now().Unix()))
```
xmonader, line 60: can you please add in the documentation the change in behavior

muhamadazmy: you are right, done

xmonader (approved):

PR #1989, opened 2023-06-28
Vm flist from files

### LINKED ISSUE #1986: allow fullvms flists to be normal flists
right now the flist of fullvm is an flist of a single raw file, we need to support as if it was normal flist with kernel files

feel free to rephrase the title and the body

muhamadazmy:
Fixes #1986

Enable loading custom kernel from the flist. This means that you can run your own custom kernel without the need to use a full vm image. 

The PR includes docs step by step how to build an ubuntu image from scratch

### DIFF
--- a/docs/manual/zmachine/zmachine.md
+++ b/docs/manual/zmachine/zmachine.md
@@ -9,10 +9,168 @@ scenario, the `Zmachine` is started from an `Flist`.
 > Note, both VM and Container on ZOS are actually served as Virtual Machines. The
 only difference is that if you are running in VM mode, you only need to provide a **raw**
 disk image (image.raw) in your flist.
+
+## Container
+
+A container is meant to host `microservice`. The `microservice` architecture generally
+dictates that each service should be run in it's own container (therefore providing
+a level of isolation), and communicate with other containers it depends on over the
+network.
+
+Similar to docker. In Zos, a container is actually also run in a virtualized environment.
+Similar to containers, some setup is done on behalf of the user. After setup this is done,
+the users `entrypoint` is started.
+
+It should be noted that a container has no control over the kernel
+used to run it, if this is required, a `VM` should be used instead. Furthermore,
+a container should ideally only have 1 process running. A container can be a single
+binary, or a complete filesystem. In general, the first should be preferred, and
+if you need the latter, it might be an indication that you actually want a `VM`.
+
+For containers, the network setup will be created for you. Your init process can
+assume that it will be fully set up (according to the config you provided) by the
+time it is started. Mountpoints will also be setup for you. The environment variables
+passed will be available inside the container.
+
 ## VM
 
-In a VM mode, you run your own operating system (for now only linux is supported)
+In container mode, zos provide a minimal kernel that is used to run a light weight VM
+and then run your app from your flist. If you need control over the kernel you can actually
+still provide it inside the flist as follows:
+
+- /boot/vmlinuz
+- /boot/initrd.img [optional]
+
+**NOTE**: the vmlinuz MUST be an EFI kernel (not compressed) if building your own kernel, or you can use the [extract-vmlinux](https://github.com/torvalds/linux/blob/master/scripts/extract-vmlinux) script to extract the EFI kernel. To test if your kernel is a valid elf kernel run command
+`readelf -n <path/to/vmlinuz>`
+
+Any of those files can be a symlink to another file in the flist.
+
+If ZOS found the `/boot/vmlinuz` file, it will use this with the initrd.img if also exists. otherwise zos will use the built-in minimal kernel and run in `container` mode.
+
+### Building an ubuntu VM flist
+
+This is a guide to help you build a working VM flist.
+
+This guide is for ubuntu `jammy`
+
+prepare rootfs
+
+```bash
+mkdir ubuntu:jammy
+```
+
+bootstrap ubuntu
+
+```bash
+sudo debootstrap jammy ubuntu:jammy  http://archive.ubuntu.com/ubuntu
+```
+
+this will create and download the basic rootfs for ubuntu jammy in the directory `ubuntu:jammy`.
+After its done we can then chroot into this directory to continue installing the necessary packages needed and configure
+few things.
+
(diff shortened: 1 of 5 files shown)


pkg/primitives/vm/utils.go
```diff
@@ -25,6 +26,145 @@ var (
...
+
+	machine.KernelImage = filepath.Join(cloudImage, "hypervisor-fw")
+	var disk *gridtypes.WorkloadWithID
+	disk, err := deployment.Get(config.Mounts[0].Name)
+	if err != nil {
+		return err
+	}
+
+	if disk.Type != zos.ZMountType {
+		return fmt.Errorf("mount is not not a valid disk workload")
```
xmonader: double not

pkg/primitives/vm/utils.go
```diff
@@ -25,6 +26,145 @@ var (
...
+func (p *Manager) prepVirtualMachine(
+	ctx context.Context,
+	cloudImage string,
+	imageInfo FListInfo,
+	machine *pkg.VM,
+	config *zos.ZMachine,
+	deployment *gridtypes.Deployment,
+	wl *gridtypes.WorkloadWithID,
+) error {
+	var (
```
xmonader: Wondering why var block is needed here?

pkg/primitives/vm/utils.go
```diff
@@ -25,6 +26,145 @@ var (
...
+		storage = stubs.NewStorageModuleStub(p.zbus)
+	)
+	// if a VM the vm has to have at least one mount
+	if len(config.Mounts) == 0 {
+		return fmt.Errorf("at least one mount has to be attached for Vm mode")
+	}
+
+	machine.KernelImage = filepath.Join(cloudImage, "hypervisor-fw")
+	var disk *gridtypes.WorkloadWithID
```
xmonader: why 48 and 49 not going together?

pkg/primitives/vm/utils.go
```diff
@@ -25,6 +26,145 @@ var (
...
+	if err = storage.DiskWrite(ctx, disk.ID.String(), imageInfo.ImagePath); err != nil {
+		return errors.Wrap(err, "failed to write image to disk")
+	}
+
+	machine.Boot = pkg.Boot{
+		Type: pkg.BootDisk,
+		Path: info.Path,
+	}
+
+	if err := p.vmMounts(ctx, deployment, config.Mounts[1:], false, machine); err != nil {
```
xmonader: probably can return p.vmMounts call directly as it's the last statement

pkg/vm/machine.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	virtioRootFsTag = "/dev/root"
+	virtioRootFsTag = "vroot"
```
xmonader, line 16: Not sure I understand this one

pkg/primitives/vm/utils.go
```diff
@@ -171,18 +311,68 @@ func (p *Manager) newPubNetworkInterface(ctx context.Context, deployment gridtyp
...
+		switch mod.Type() {
+		case 0:
+			// regular file, do nothing
+		case os.ModeSymlink:
+			// this is a symlink. we
+			// need to make sure it's a safe link
+			// to a location inside the flist
+			link, err := os.Readlink(path)
+			if err != nil {
+				return flist, errors.Wrapf(err, "failed to link at '%s", rel)
```
xmonader: failed to read link

pkg/primitives/vm/utils.go
muhamadazmy: it had more stuff then removed then the block remained

pkg/vm/manager.go
```diff
@@ -436,6 +436,7 @@ func (m *Module) Run(vm pkg.VM) (pkg.MachineInfo, error) {
 		// arguments
 		cmdline["root"] = virtioRootFsTag
 		cmdline["rootfstype"] = "virtiofs"
+		cmdline["rootdelay"] = "30"
```
xmonader, line 439: what does this value affect?

pkg/primitives/vm/utils.go
muhamadazmy: left over from the older code

pkg/primitives/vm/utils.go
muhamadazmy: true

docs/manual/zmachine/zmachine.md
```diff
@@ -36,82 +199,74 @@ to take the full disk size allocated in the deployment.
...
     suboptimal performance right now.
 - It's not currently possible to get container logs.
 - TODO: more testing
 
 ## Creating VM image
+
 This is a simple tutorial on how to create your own VM image
 > Note: Please consider checking the official vm images repo on the hub before building your own
-image. this can save you a lot of time (and network traffice) here https://hub.grid.tf/tf-official-vms
+image. this can save you a lot of time (and network traffice) here <https://hub.grid.tf/tf-official-vms>
```
xmonader: traffice -> traffic

pkg/vm/machine.go
muhamadazmy: this is a tag that is given to the rootfs virtiofs and is needed to be used to mount the `/` inside the VM from a virtiofs that is shared from the host. the problem is ubuntu is really really not happy that the tag has `/dev` prefix because now it thinks it's an actual device then fails to mount the filesystem because of checks did not pass.

Hence i had to change the tag. the tag has no special meaning and can be anything (was recommended by CH to use /dev/root) because it's basically and id of which fs to mount.

Changing this to `vroot` or any other unique string should fix the problem

pkg/vm/manager.go
muhamadazmy: Ah i needed it during the testing it's basically how long the init script waits for the filesystem before it's available. this is completely ignored by anything that is not using ubuntu initramfs. But i thought to keep it since it give ubuntu more time to wait before it fails, and has no effect with anything else

xmonader (approved):

xmonader (approved):

Issue #1994, opened 2023-07-03
`gpu.list` returns an incorrect GPU device


AbdelrahmanElawady:
Using https://gridproxy.dev.grid.tf/nodes/93/gpu to query GPUs on node 93 shows a device name of `Navi 31 [Radeon RX 7900 XT/7900 XTX]` as the GPU on the node.
However, `RX 7900 XT` is a different than `RX 7900 XTX` https://en.wikipedia.org/wiki/Radeon_RX_7000_series. So, it should only return one of them (the one used).

AbdelrahmanElawady:
According to PCI IDs [repository](https://pci-ids.ucw.cz/mods/PC?action=help?help=pci) the ID is separated into vendor, device and subsystem. For some GPUs device ID is not enough to identify the GPU and subsystem ID is used to identify it.
For example:

SSID doesn't cover all variations of the card:
```
	744c  Navi 31 [Radeon RX 7900 XT/7900 XTX]
		1da2 e471  NITRO+ RX 7900 XTX Vapor-X
		1eae 7901  RX-79XMERCB9 [SPEEDSTER MERC 310 RX 7900 XTX]
```

SSID for both variations:
```
	1e78  TU102GL [Quadro RTX 6000/8000]
		10de 13d8  Quadro RTX 8000
		10de 13d9  Quadro RTX 6000
```

So, using SSID will further improve determining what GPU is used. However, SSID data is not well maintained so it won't cover all GPUs.

muhamadazmy:
Yeah, i was gonna comment exactly that on the issue but you already came to the same conclusion. The problem is the system only expose the Vendor and Device ID. while the db already has a list of (sub devices) i couldn't find the sub-device id exposed by the system hence it was not possible to identify

That being said, All devices that has the same device ID are used exactly the same way (with same drivers and everything) while they are different they should expose the same `interface` and be used the same way.

So there might be differences but it shouldn't be big enough to be identified as a different device.

muhamadazmy:
Just notices, there is a PR to do exactly that, i will take a look. thank you

AbdelrahmanElawady:
actually in some cases the difference is double the VRAM in `Quadro RTX 6000/8000` example. 6000 has 24 GB VRAM while 8000 has 48 GB VRAM and that could affect users needs a lot.

muhamadazmy:
I am wondering if we really actually need to "make" a difference because they use the same `chip`. 
[...]

PR #1996, opened 2023-07-04
Use SSID to identify PCI devices

AbdelrahmanElawady:
Use SSID to further specify which device is connected to PCI.

Related: #1994 

Notes:
- Some GPUs will still not be identified accurately due to `pci.ids` not having SSID for all devices.

### DIFF
--- a/pkg/capacity/pci.go
+++ b/pkg/capacity/pci.go
@@ -16,16 +16,24 @@ const (
 	pciDir = "/sys/bus/pci/devices"
 )
 
+// Subdevice is subdevice information to a PCI device
+type Subdevice struct {
+	// SubsystemVendorID is device subsystem vendor ID according to PCI database
+	SubsystemVendorID uint16
+	// SubsystemVendorID is device subsystem ID according to PCI database
+	SubsystemDeviceID uint16
+	// Name is subdevice name according to PCI database
+	Name string
+}
+
 // Device is a PCI device
 type Device struct {
 	// ID is device id according to PCI database
 	ID uint16
 	// Name is device name according to PCI database
 	Name string
-	// SubsystemVendorID is device subsystem vendor ID according to PCI database
-	SubsystemVendorID uint16
-	// SubsystemVendorID is device subsystem ID according to PCI database
-	SubsystemDeviceID uint16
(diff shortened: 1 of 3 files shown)


pkg/capacity/pci.go
```diff
@@ -115,12 +145,18 @@ func GetVendor(vendor uint16) (Vendor, bool) {
...
 	if !ok {
 		return
 	}
-	d, ok = v.Devices[device]
+	key := fmt.Sprintf("%04x%04x%04x", device, subsystemVendorID, subsystemDeviceID)
+	d, ok = v.Devices[key]
+	if ok {
+		return
+	}
+	key = fmt.Sprintf("%04x", device)
```
muhamadazmy: Why this has to be string?

pkg/capacity/pci.go
```diff
@@ -31,7 +35,7 @@ type Vendor struct {
 	// Name of the vendor
 	Name string
 	// All known devices by this vendor
-	Devices map[uint16]Device
+	Devices map[string]Device
```
muhamadazmy: This is not a good structure right now because it does not reflect the nature hierarchy of the devices . Instead this can go as
```
Devices map[uint16]Device
```
then device itself becomes 
```
type Device struct {
   ID uint16
   Name string
   Subdevices []SubDevice
}
```
Where subdevices now can has a list of subdevices which only holds subdevice information (including sub vendor and sub device id and name

muhamadazmy (changes requested):
Also please read comments on the issue

pkg/capacity/pci.go
AbdelrahmanElawady: makes sense, and the user calling `pci.GetDevice()` gets the Device with its Subdevices and I should add a `pci.GetSubdevice` that returns `(Subdevice, bool)`?

muhamadazmy: Yes, that is possible. the Device id returned by the API should still be only `slot/vendor/device` ids without the subdevice id, but the Information can include name, and subsystem vendor and device.

Because the ID is still unique (because of the slot) so no extra information is needed.

muhamadazmy (approved):

Issue #1985, opened 2023-06-26
prettify shutting down


muhamadazmy:
Currently when zos is shutting down because of farmer bot power mgmt, a panic is shown on zos screen that redis connection (socket) is gone. 
This is not harmful because zos stops redis immediately before it power off. 
Although not harmful but it's really ugly and makes the farmer worries that something is wrong . We need to prevent this message by making sure zui shutdown nicely (and clean up the terminal) and may be show a system shutting down message.

xmonader:
https://github.com/threefoldtech/zos/commit/0e7b51fa983679ae0680d60256f6a791b33fe518

scottyeager:
It seems that maybe the solution didn't work as intended. A farmer is reporting they still see the following on screen when a node with Zos 3.8.4 shuts down:

![image](https://github.com/threefoldtech/zos/assets/17682591/7f59b3c6-0eff-4fc3-b1eb-6cb250bfa9dc)

scottyeager:
Still seeing cases of the ugly shutdown screen. Reported today:

![image](https://github.com/threefoldtech/zos/assets/17682591/3bd53ad7-57ff-4326-b02b-78452bf06fc2)

Could you have another look at this, @MarioBassem?

PR #2003, opened 2023-07-11
update zui service to be after redis in zinit service tree

mariobassem:
## Related issues:

- #1985

### DIFF
--- a/etc/zinit/zui.yaml
+++ b/etc/zinit/zui.yaml
@@ -12,3 +12,4 @@ exec: |
   '
 after:
   - quiet
+  - redis


xmonader:
Where is the issue this is linked to?

mariobassem:
> Where is the issue this is linked to?

#1985 
it does not actually prettify shutting down, but it just shuts down zui before redis to prevent showing panic

muhamadazmy:
Thank you @MarioBassem . but in the office i quickly mentioned that depending on redis directly might be not needed. instead we can use an umbrella service (called node-ready) https://github.com/threefoldtech/zos/blob/main/etc/zinit/node-ready.yaml 
the service is just a preparation script and depends on all basic services (including redis)

Anyway, yesterday after our discussion i was working on a PR and already added this to ZUI and it was merged yesterday. so this change is no longer needed 

https://github.com/threefoldtech/zos/commit/0e7b51fa983679ae0680d60256f6a791b33fe518

Thanks again

Issue #1887, opened 2023-01-23
use a list of geoip services in zos


xmonader:
blocked on https://github.com/threefoldtech/tf_operations/issues/1374

related to https://github.com/threefoldtech/zos/issues/1884

muhamadazmy:
https://github.com/threefoldtech/tf_support/issues/444#issuecomment-1493802378

xmonader:
It's ongoing now to deploy more geoip instances https://github.com/threefoldtech/tf_operations/issues/1374#issuecomment-1604523903, will move back to accepted

xmonader:
actually moving to 3.12 now

muhamadazmy:
@rawdaGastan any update on this, the list of extra geoip services are available for a while now.

coesensbert:
https://docs.grid.tf/threefold/itenv_threefold_main/src/branch/master/Grid-operations/geo-ip-service.md#current-deployments

- https://geoip.grid.tf/
- https://02.geoip.grid.tf/
- https://03.geoip.grid.tf/

PR #2015, opened 2023-07-31
add more geoip service urls

mariobassem:
## Related issues:

- #1887

### DIFF
--- a/pkg/geoip/geoip.go
+++ b/pkg/geoip/geoip.go
@@ -25,7 +25,7 @@ func Fetch() (Location, error) {
 	for _, url := range geoipURLs {
 		l, err := getLocation(url)
 		if err != nil {
-			log.Err(err).Msg("failed to fetch location. retrying...")
+			log.Err(err).Msgf("failed to fetch location from geoip service %s. retrying...", url)
 			continue
 		}
 
@@ -35,20 +35,27 @@ func Fetch() (Location, error) {
 	return Location{}, errors.New("failed to fetch location information")
 }
 
-func getLocation(url string) (Location, error) {
-	resp, err := http.Get(url)
+func getLocation(geoIPService string) (Location, error) {
+	l := Location{
+		Longitute: 0.0,
+		Latitude:  0.0,
+		Continent: "Unknown",
+		Country:   "Unknown",
+		City:      "Unknown",
+	}
+
+	resp, err := http.Get(geoIPService)
 	if err != nil {
-		return Location{}, err
+		return l, err
 	}
 	defer resp.Body.Close()
 
 	if resp.StatusCode != 200 {
-		return Location{}, errors.New("error fetching location")
+		return l, errors.New("error fetching location")
 	}
 
-	l := Location{}
 	if err := json.NewDecoder(resp.Body).Decode(&l); err != nil {
-		return Location{}, err
+		return l, err
 	}
 
 	return l, nil
--- a/pkg/geoip/geoip.go
+++ b/pkg/geoip/geoip.go
@@ -22,34 +22,34 @@ type Location struct {
 func Fetch() (Location, error) {
 	geoipURLs := []string{"https://geoip.grid.tf/", "https://02.geoip.grid.tf/", "https://03.geoip.grid.tf/"}
 
-	l := Location{
-		Longitute: 0.0,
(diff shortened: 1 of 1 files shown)


pkg/geoip/geoip.go
```diff
@@ -26,19 +30,26 @@ func Fetch() (Location, error) {
 		City:      "Unknown",
 	}
 
-	resp, err := http.Get("https://geoip.grid.tf")
-	if err != nil {
-		return l, err
-	}
-	defer resp.Body.Close()
+	for i := 0; i < len(geoipURLs); i++ {
```
xmonader: why not a for range?

pkg/geoip/geoip.go
```diff
@@ -26,19 +30,26 @@ func Fetch() (Location, error) {
...
+		resp, err := http.Get(geoipURLs[i])
+		if err != nil {
+			log.Err(err).Msgf("failed to make http call to geoip service %s. retrying...", geoipURLs[i])
+			continue
+		}
+		defer resp.Body.Close()
```
xmonader: this won't be properly handled (you're calling the defer within a for-loop)

xmonader (commented):
I was expecting that we would fire up multiple requests in the background against the 3 geoip services and as soon one of the returns a healthy result we can use it, but  @muhamadazmy can decide how to approach it properly

pkg/geoip/geoip.go
```diff
@@ -22,34 +22,34 @@ type Location struct {
...
 
-		if err := json.NewDecoder(resp.Body).Decode(&l); err != nil {
-			log.Err(err).Msgf("failed to decode location data from %s", geoipURLs[i])
-			continue
-		}
+	return Location{}, errors.New("failed to fetch location information")
+}
 
-		return l, nil
+func getLocation(url string) (Location, error) {
```
xmonader: see how many Location{} ? can use a named parameter or even create it on the top of the function and reuse it

pkg/geoip/geoip.go
```diff
@@ -22,34 +22,34 @@ type Location struct {
...
-		City:      "Unknown",
-	}
-
-	for i := 0; i < len(geoipURLs); i++ {
-		resp, err := http.Get(geoipURLs[i])
+	for _, url := range geoipURLs {
+		l, err := getLocation(url)
 		if err != nil {
-			log.Err(err).Msgf("failed to make http call to geoip service %s. retrying...", geoipURLs[i])
+			log.Err(err).Msg("failed to fetch location. retrying...")
```
xmonader: encode which broken service

pkg/geoip/geoip.go
xmonader: can use a better name for url ( e.g geoIPService) because it's not apparent from the first sight what url that is

xmonader (dismissed):

pkg/geoip/geoip.go
```diff
@@ -18,6 +20,22 @@ type Location struct {
 
 // Fetch retrieves the location of the system calling this function
 func Fetch() (Location, error) {
+	geoipURLs := []string{"https://geoip.grid.tf/", "https://02.geoip.grid.tf/", "https://03.geoip.grid.tf/"}
```
muhamadazmy: can this be defined in the global scope (outside the function body) not on the stack. it's more of a constant than a variable. like

```go
var (
   geoipURLs = []string{...}
)
```

pkg/geoip/geoip.go
```diff
@@ -18,6 +20,22 @@ type Location struct {
...
+
+	for _, url := range geoipURLs {
+		l, err := getLocation(url)
+		if err != nil {
+			log.Err(err).Msgf("failed to fetch location from geoip service %s. retrying...", url)
```
muhamadazmy: the log here is kinda misleading, we are not `retrying` instead we are trying a completely different url. Instead the error should show which url is not working + the cause. Also please use the structured logs correctly. Use the available, Str, Int, etc.. not formatting as much as possiblel

```go
log.Err(err).Str("url", url).Msg("failed to fetch location from geoip service")
```

Is more correct form than the one in code

muhamadazmy (changes requested):

muhamadazmy (approved):
Please merge only after testing

Issue #2016, opened 2023-07-31
[provision][disk] Make sure to order disks in deployments to bigger ones are allocated first


muhamadazmy:
Related to https://github.com/threefoldtech/tfgrid-sdk-ts/issues/886

muhamadazmy:
For DEV

https://github.com/threefoldtech/zos/blob/main/pkg/provision/engine.go#L872

this is where we sort workloads by types, before deploying, a second sort on workloads of disk type, a sort must be done so bigger disks comes first. (descending order on disks size)

PR #2017, opened 2023-07-31
Sort zmount workloads in descending order

AbdelrahmanElawady:
Sort zmount workloads so larger disks get assigned first.

Related: #2016

### DIFF
--- a/pkg/provision/engine.go
+++ b/pkg/provision/engine.go
@@ -15,6 +15,7 @@ import (
 	substrate "github.com/threefoldtech/tfchain/clients/tfchain-client-go"
 	"github.com/threefoldtech/zos/pkg"
 	"github.com/threefoldtech/zos/pkg/gridtypes"
+	"github.com/threefoldtech/zos/pkg/gridtypes/zos"
 )
 
 // EngineOption interface
@@ -867,9 +868,37 @@ func (e *NativeEngine) uninstallDeployment(ctx context.Context, dl *gridtypes.De
 
 }
 
+func getZmountSize(wl *gridtypes.WorkloadWithID) (gridtypes.Unit, error) {
+	dataI, err := wl.WorkloadData()
+	if err != nil {
+		return 0, err
+	}
+	data, ok := dataI.(*zos.ZMount)
+	if !ok {
+		return 0, errors.New("not a zmount")
+	}
+	return data.Size, nil
+}
+
 func (e *NativeEngine) installDeployment(ctx context.Context, getter gridtypes.WorkloadGetter) {
 	for _, typ := range e.order {
 		workloads := getter.ByType(typ)
+
+		if typ == zos.ZMountType {
+			sort.Slice(workloads, func(i, j int) bool {
+				sizeI, err := getZmountSize(workloads[i])
+				if err != nil {
+					log.Error().Err(err).Stringer("id", workloads[i].ID).Msg("failed to get zmount size")
+				}
+
+				sizeJ, err := getZmountSize(workloads[j])
+				if err != nil {
+					log.Error().Err(err).Stringer("id", workloads[j].ID).Msg("failed to get zmount size")
+				}
+
+				return sizeI > sizeJ
+			})
+		}
 
 		for _, wl := range workloads {
 			if err := e.installWorkload(ctx, wl); err != nil {
--- a/pkg/provision/engine.go
+++ b/pkg/provision/engine.go
@@ -873,11 +873,29 @@ func getZmountSize(wl *gridtypes.WorkloadWithID) (gridtypes.Unit, error) {
 	if err != nil {
 		return 0, err
 	}
-	data, ok := dataI.(*zos.ZMount)
+	zmount, ok := dataI.(*zos.ZMount)
 	if !ok {
-		return 0, errors.New("not a zmount")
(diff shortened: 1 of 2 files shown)


pkg/provision/engine.go
```diff
@@ -867,10 +868,38 @@ func (e *NativeEngine) uninstallDeployment(ctx context.Context, dl *gridtypes.De
 
 }
 
+func getZmountSize(wl *gridtypes.WorkloadWithID) (gridtypes.Unit, error) {
+	dataI, err := wl.WorkloadData()
+	if err != nil {
+		return 0, err
+	}
+	data, ok := dataI.(*zos.ZMount)
```
xmonader: at this point it should be called zmount instead of data

pkg/provision/engine.go
```diff
@@ -867,10 +868,38 @@ func (e *NativeEngine) uninstallDeployment(ctx context.Context, dl *gridtypes.De
...
+	if !ok {
+		return 0, errors.New("not a zmount")
+	}
+	return data.Size, nil
```
xmonader: zmount.Size after the rename

pkg/provision/engine.go
```diff
@@ -867,10 +868,38 @@ func (e *NativeEngine) uninstallDeployment(ctx context.Context, dl *gridtypes.De
...
 }
 
+func getZmountSize(wl *gridtypes.WorkloadWithID) (gridtypes.Unit, error) {
+	dataI, err := wl.WorkloadData()
+	if err != nil {
+		return 0, err
+	}
+	data, ok := dataI.(*zos.ZMount)
+	if !ok {
+		return 0, errors.New("not a zmount")
```
xmonader: I would be curious to see the raw data included here in the error, also the message should get improved instead of `not a zmount`

pkg/provision/engine.go
```diff
@@ -867,10 +868,38 @@ func (e *NativeEngine) uninstallDeployment(ctx context.Context, dl *gridtypes.De
...
+				if err != nil {
+					log.Error().Err(err).Stringer("id", workloads[i].ID).Msg("failed to get zmount size")
+				}
+
+				sizeJ, err := getZmountSize(workloads[j])
+				if err != nil {
+					log.Error().Err(err).Stringer("id", workloads[j].ID).Msg("failed to get zmount size")
+				}
+
+				return sizeI > sizeJ
```
xmonader: How can we depend on sizeI and sizeJ while they're supposed to be the zero value returns of getZmountSize in case of errors? I believe each branch after the error check should return a bool (e.g false in the first case and true in the 2nd one)

xmonader: Also, maybe extract the code to sort ZMount workloads into its own function, so it could be easily tested

pkg/provision/engine.go
```diff
@@ -867,10 +868,44 @@ func (e *NativeEngine) uninstallDeployment(ctx context.Context, dl *gridtypes.De
 
 }
 
+func getZmountSize(wl *gridtypes.WorkloadWithID) (gridtypes.Unit, error) {
```
muhamadazmy:
```suggestion
func getZmountSize(wl *gridtypes.Workload) (gridtypes.Unit, error) {
```
you don't need to pass the wrapper type `WorloadWithID`

pkg/provision/engine.go
```diff
@@ -867,10 +868,44 @@ func (e *NativeEngine) uninstallDeployment(ctx context.Context, dl *gridtypes.De
...
+	dataI, err := wl.WorkloadData()
```
muhamadazmy: Don't use the `I` suffix, `data` is enough. It add no value to the variable name
```suggestion
	data, err := wl.WorkloadData()
```

pkg/provision/engine.go
```diff
@@ -867,10 +868,44 @@ func (e *NativeEngine) uninstallDeployment(ctx context.Context, dl *gridtypes.De
...
+		return 0, fmt.Errorf("failed to get workload data as a zmount '%v'", dataI)
+	}
+	return zmount.Size, nil
+}
+
+func sortZmountWorkloads(workloads []*gridtypes.WorkloadWithID) {
+	sort.Slice(workloads, func(i, j int) bool {
+		sizeI, err := getZmountSize(workloads[i])
+		if err != nil {
+			log.Error().Err(err).Stringer("id", workloads[i].ID).Msg("failed to get zmount size")
```
muhamadazmy: i get why you log but i think it's not needed here. I won't show an error in the node logs because of wrong user data. This can confuse operators to see node giving errors that might make them think something wrong with the node itself while it's actual just user data that is wrong and that it was handled. 

I would personally not log it.

pkg/provision/engine.go
```diff
@@ -867,10 +868,44 @@ func (e *NativeEngine) uninstallDeployment(ctx context.Context, dl *gridtypes.De
...
+			return false
+		}
+
+		sizeJ, err := getZmountSize(workloads[j])
+		if err != nil {
+			log.Error().Err(err).Stringer("id", workloads[j].ID).Msg("failed to get zmount size")
+			return true
```
muhamadazmy: Why do u return true here but false in the first case ? just wondering. I think you should return false always in case if both elements are `equal`.

muhamadazmy (changes requested):
Great job. Just few minor comments

muhamadazmy (approved):

Issue #1970, opened 2023-05-22
SSD with node id later detected as HDD


scottyeager:
Farmer reported that their node 3791 detected the cache SSD as HDD after the node rebooted. There's two concerns here:

1. Disk type should not change (I thought this was now stored somewhere after initial detection?)
2. Maybe it makes sense to check for the presence of the node id on a disk, even if Zos thinks it's a HDD, to avoid generating a second node id in case there is some problem?

muhamadazmy:
The disk type is stored on a volatile storage (tmpfs) after the node is booted because disk type should not change while the node is running. This is like this because after a runtime update, storage daemon restarts and then it will need to re-detect the disks but because disks might be under heavy load detection can be off, hence it uses the stored detection that happened during boot.

On a restart the disk types are wiped out and then has to  be re-done, but this happens before any workload is restarted so it should be safe to run. 

We also need to rescan the disks anyway on boot in case new disks are added or disks have been swapped. 

The failure to detect the disk as SSD can also mean the disk of a low quality that is now not peformant enough to be considered SSD. Note that we don't rely on the "announced" disk type, we instead run a speed test to make sure the disks speed is actually SSD grade.

Last thing is using an HDD as a cache disk will really hurt the node performance, hence zos ignore hdd for cache (and hence id storage)

scottyeager:
Ah, I do recall now that disk type was fixed for the time that the node is up.

I wondered a similar thing about a disk quality issue or performance degradation that dropped its seek time into the HDD category. Definitely the node should not use HDD or bad performing SSD as its cache disk. But if the node id or some workload data is already stored on a disk originally detected as SSD and that disk is later detected as HDD, then that data is essentially lost.

It seems the system would be a bit more resilient if something like the following could happen:

1. Node scans disks and detects that a node identity file or workload data is present on a disk labeled as HDD
2. We assume this disk has another disk detected as SSD. The identity is copied to this disk and it is used now for cache
3. In the case or workload data, it remains available to that workload. Performance will be bad but at least the data can be copied elsewhere
4. Don't store anymore data to this disk and don't count it towards minting rewards. Disks which malfunction in this way are probably close to complete failure and the farmer should replace it anyway

muhamadazmy:
I know this is an old issue but we might be able to process this now. I like your suggestion but I also suggest we do this as follows:
[...]

PR #2023, opened 2023-08-14
save device type instead of caching it

rawdaGastan:
### Issues
 - https://github.com/threefoldtech/zos/issues/1970

### DIFF
--- a/pkg/storage/filesystem/device.go
+++ b/pkg/storage/filesystem/device.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog/log"
-	"github.com/threefoldtech/zos/pkg"
 	"github.com/threefoldtech/zos/pkg/gridtypes/zos"
 )
 
@@ -76,17 +75,21 @@ func (d *DeviceInfo) DetectType() (zos.DeviceType, error) {
 }
 
 // SetType sets the device type to the disk
-func (d *DeviceInfo) SetType(typ pkg.DeviceType) error {
-	if err := os.WriteFile(filepath.Join("/mnt", d.Name(), ".seektime"), []byte(typ), 0644); err != nil {
-		return errors.Wrapf(err, "failed to store device type for '%s'", d.Name())
+func (d *DeviceInfo) SetType(typ zos.DeviceType) error {
+	diskFilePath := filepath.Join("/mnt", d.Name(), ".seektime")
+	if err := os.WriteFile(diskFilePath, []byte(typ), 0644); err != nil {
+		return errors.Wrapf(err, "failed to store device type for '%s' in '%s'", d.Name(), diskFilePath)
 	}
 
 	return nil
 }
 
 // Type gets the device type from the disk
+// return the device type, if it's found and an error
+// that's based on .seektime file existing in the /mnt/DeviceName/.seektime, that contains the device type being SSD or HDD maybe
 func (d *DeviceInfo) Type() (zos.DeviceType, bool, error) {
-	data, err := os.ReadFile(filepath.Join("/mnt", d.Name(), ".seektime"))
+	diskFilePath := filepath.Join("/mnt", d.Name(), ".seektime")
+	diskType, err := os.ReadFile(diskFilePath)
 	if os.IsNotExist(err) {
 		return "", false, nil
 	}
@@ -95,11 +98,11 @@ func (d *DeviceInfo) Type() (zos.DeviceType, bool, error) {
 		return "", false, err
 	}
 
-	if len(data) == 0 {
+	if len(diskType) == 0 {
 		return "", false, nil
 	}
 
-	return pkg.DeviceType(data), true, nil
+	return zos.DeviceType(diskType), true, nil
 }
 
 func (d *DeviceInfo) Mountpoint(ctx context.Context) (string, error) {
--- a/pkg/storage/filesystem/device.go
+++ b/pkg/storage/filesystem/device.go
@@ -87,7 +87,7 @@ func (d *DeviceInfo) SetType(typ pkg.DeviceType) error {
 // Type gets the device type from the disk
 func (d *DeviceInfo) Type() (zos.DeviceType, bool, error) {
 	data, err := os.ReadFile(filepath.Join("/mnt", d.Name(), ".seektime"))
-	if errors.Is(err, os.ErrNotExist) {
+	if os.IsNotExist(err) {
 		return "", false, nil
 	}
 
--- a/pkg/storage/filesystem/device.go
+++ b/pkg/storage/filesystem/device.go
@@ -4,12 +4,14 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog/log"
+	"github.com/threefoldtech/zos/pkg"
 	"github.com/threefoldtech/zos/pkg/gridtypes/zos"
 )
 
@@ -68,8 +70,29 @@ func (i *DeviceInfo) Used() bool {
 	return len(i.Label) != 0 || len(i.Filesystem) != 0
 }
 
-func (d *DeviceInfo) Type() (zos.DeviceType, error) {
+func (d *DeviceInfo) DetectType() (zos.DeviceType, error) {
 	return d.mgr.Seektime(context.Background(), d.Path)
+}
+
+func (d *DeviceInfo) SetType(typ pkg.DeviceType) error {
+	if err := os.WriteFile(filepath.Join("/mnt", d.Name(), ".seektime"), []byte(typ), 0644); err != nil {
+		return errors.Wrapf(err, "failed to store device type for '%s'", d.Name())
+	}
+
+	return nil
+}
+
+func (d *DeviceInfo) Type() (zos.DeviceType, bool) {
+	data, err := os.ReadFile(filepath.Join("/mnt", d.Name(), ".seektime"))
+	if err != nil {
+		return "", false
+	}
+
+	if len(data) == 0 {
+		return "", false
+	}
+
+	return pkg.DeviceType(data), true
 }
 
 func (d *DeviceInfo) Mountpoint(ctx context.Context) (string, error) {
--- a/pkg/storage/filesystem/device.go
+++ b/pkg/storage/filesystem/device.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"os"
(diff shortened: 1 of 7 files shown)


pkg/storage/filesystem/device.go
```diff
@@ -68,10 +70,34 @@ func (i *DeviceInfo) Used() bool {
...
+		return errors.Wrapf(err, "failed to store device type for '%s'", d.Name())
+	}
+
+	return nil
+}
+
+// Type gets the device type from the disk
+func (d *DeviceInfo) Type() (zos.DeviceType, bool) {
+	data, err := os.ReadFile(filepath.Join("/mnt", d.Name(), ".seektime"))
+	if err != nil {
```
muhamadazmy: this can be any error type, but only `NotFound` is false, hence this method actually should return `(type, bool, error)

muhamadazmy: Consider that any error is a `false` can hide other problems with the disk or the file system where in matter of fact the disk should be considered `broken` or in a wrong state

pkg/storage/storage.go
```diff
@@ -206,31 +198,38 @@ func (s *Module) initialize(ctx context.Context) error {
...
 
-			log.Debug().Str("device", device.Path).Str("type", typ.String()).Msg("caching device type")
-			if err := s.cache.Set(device.Name(), typ); err != nil {
-				log.Error().Str("device", device.Path).Err(err).Msg("failed to cache device type")
+				log.Debug().Str("device", device.Path).Str("type", typ.String()).Msg("saving device type")
+				if err := device.SetType(typ); err != nil {
+					log.Error().Str("device", device.Path).Err(err).Msg("failed to save device type")
+				}
+			} else {
+				log.Debug().Str("device", device.Path).Str("type", typ.String()).Msg("device type loaded from cache")
```
muhamadazmy: this type then need to be stored to disk. So both cases if you get type by detection or from cache has to be written to disk.

muhamadazmy (changes requested):
I have few comments. but also i personally don't deep nesting like that, so the code might get some re-organized and or moved to separate functions.

for example
```go

typ, err := s.diskType(device)

// where we have 
func diskType(device)  {
 //get type 
 if found {
   return
} 
 
 cached, found := cache.get()
 if !found {
   cached = detected() 
  }

// store on device
return
}
 
```

pkg/storage/filesystem/device.go
```diff
@@ -68,10 +70,38 @@ func (i *DeviceInfo) Used() bool {
...
+		return errors.Wrapf(err, "failed to store device type for '%s'", d.Name())
+	}
+
+	return nil
+}
+
+// Type gets the device type from the disk
+func (d *DeviceInfo) Type() (zos.DeviceType, bool, error) {
+	data, err := os.ReadFile(filepath.Join("/mnt", d.Name(), ".seektime"))
+	if errors.Is(err, os.ErrNotExist) {
```
muhamadazmy: The correct way to do this is
```suggestion
	if os.IsNotExist(err) {
```

pkg/storage/storage.go
```diff
@@ -157,6 +149,50 @@ func (s *Module) dump() {
...
+	if ok {
+		log.Debug().Str("device", device.Path).Str("type", typ.String()).Msg("device type loaded from disk")
+		return typ, nil
+	}
+
+	log.Debug().Str("device", device.Path).Msg("checking device type in cache")
+	typ, ok = s.cache.Get(device.Name())
+	if ok {
+		log.Debug().Str("device", device.Path).Str("type", typ.String()).Msg("device type loaded from cache")
+		return typ, nil
```
muhamadazmy: don't return here because this type need to be stored on the disk, if u check my psudo code, I check if NOT cache, i do detect then ALWAYS store the result on disk

pkg/storage/storage.go
```diff
@@ -157,6 +149,50 @@ func (s *Module) dump() {
...
+
+	// for development purposes only
+	if vm {
+		// force ssd device for vms
+		typ = zos.SSDDevice
+
+		if device.Path == "/dev/vdd" || device.Path == "/dev/vde" {
+			typ = zos.HDDDevice
+		}
+	}
```
muhamadazmy: I think this check can be moved to the top of this function no need to store the type because we force ssd ALWAYS except for vdd and vde which are always HDD

pkg/storage/storage.go
```diff
@@ -206,31 +242,9 @@ func (s *Module) initialize(ctx context.Context) error {
...
-
-			log.Debug().Str("device", device.Path).Str("type", typ.String()).Msg("caching device type")
-			if err := s.cache.Set(device.Name(), typ); err != nil {
-				log.Error().Str("device", device.Path).Err(err).Msg("failed to cache device type")
-			}
-		} else {
-			log.Debug().Str("device", device.Path).Str("type", typ.String()).Msg("device type loaded from cache")
+		typ, err := s.deviceType(device, vm)
+		if err != nil {
+			log.Error().Str("device", device.Path).Err(err).Msg("failed to get device type")
```
muhamadazmy, line 249: you need to continue after logging the error. (might also be a good idea to add them to broken pools)

muhamadazmy (changes requested):

muhamadazmy (dismissed):
Please don't merge unless code is fully tested

pkg/storage/filesystem/device.go
```diff
@@ -68,10 +70,38 @@ func (i *DeviceInfo) Used() bool {
 	return len(i.Label) != 0 || len(i.Filesystem) != 0
 }
 
-func (d *DeviceInfo) Type() (zos.DeviceType, error) {
+// DetectType returns the device type according to seektime
```
xmonader, line 71: First time for me to know about the seektime tool, maybe we can enhance it to include that it's an approximation too https://github.com/maxux/seektime not 100% correct

pkg/storage/filesystem/device.go
```diff
@@ -68,10 +70,38 @@ func (i *DeviceInfo) Used() bool {
...
+func (d *DeviceInfo) DetectType() (zos.DeviceType, error) {
 	return d.mgr.Seektime(context.Background(), d.Path)
 }
 
+// SetType sets the device type to the disk
+func (d *DeviceInfo) SetType(typ pkg.DeviceType) error {
```
xmonader: Why don't we use zos.DeviceType? I'm not sure why we created that alias declaration DeviceType = DeviceType?
Also the rest of the functions DetectType, Type are working with zos.DeviceType

pkg/storage/filesystem/device.go
```diff
@@ -68,10 +70,38 @@ func (i *DeviceInfo) Used() bool {
...
+	if err := os.WriteFile(filepath.Join("/mnt", d.Name(), ".seektime"), []byte(typ), 0644); err != nil {
+		return errors.Wrapf(err, "failed to store device type for '%s'", d.Name())
```
xmonader: please include the full path as well so `/mnt/d.Name(), ...`

pkg/storage/filesystem/device.go
```diff
@@ -68,10 +70,38 @@ func (i *DeviceInfo) Used() bool {
...
 
-func (d *DeviceInfo) Type() (zos.DeviceType, error) {
+// DetectType returns the device type according to seektime
+func (d *DeviceInfo) DetectType() (zos.DeviceType, error) {
 	return d.mgr.Seektime(context.Background(), d.Path)
 }
 
+// SetType sets the device type to the disk
+func (d *DeviceInfo) SetType(typ pkg.DeviceType) error {
+	if err := os.WriteFile(filepath.Join("/mnt", d.Name(), ".seektime"), []byte(typ), 0644); err != nil {
```
xmonader: let's extract the first argument of WriteFile into a varilable to improve the readability

pkg/storage/filesystem/device.go
```diff
@@ -68,10 +70,38 @@ func (i *DeviceInfo) Used() bool {
...
+		return errors.Wrapf(err, "failed to store device type for '%s'", d.Name())
+	}
+
+	return nil
+}
+
+// Type gets the device type from the disk
```
xmonader: let's expand the docstring, that's based on `.seektime` file existing in the /mnt/DeviceName.seektime, that contains the device type being SSD or HDD maybe

pkg/storage/filesystem/device.go
```diff
@@ -68,10 +70,38 @@ func (i *DeviceInfo) Used() bool {
...
+func (d *DeviceInfo) Type() (zos.DeviceType, bool, error) {
+	data, err := os.ReadFile(filepath.Join("/mnt", d.Name(), ".seektime"))
```
xmonader: - let's extract to its own variable
- let's rename data to disktype

pkg/storage/filesystem/device.go
xmonader: also i'm not sure of the behavior of the return type device, bool, err

pkg/storage/filesystem/device.go
xmonader: actually I believe we shouldn't use `.seektime` extension, as it denotes the results of seektime (rpm/seektime) instead of actual harddisk type string, however if it's considered as "the output of seektime" tool, I'm good with that

xmonader:
I noticed this in the seektime code https://github.com/maxux/seektime/issues/2 , not sure how it affects the accuracy or the calculations @maxux @muhamadazmy ?

maxux:
Regarding https://github.com/maxux/seektime/issues/2, this doesn't affect any accuracy in our case.

But this https://github.com/threefoldtech/seektime/pull/3 was ready for a long time now and were never merged, but this could provide a way better help to troubleshoot in case of issue.

Does merging this PR would affect something in zos ? The json reply would change, maybe you need to support it ?

This pull requests will print `min` and `max` seektime, this helped me previously to diagnostic a bug where HDD were sleeping and disks needed to be woke up on the first check; `max` was set to few of seconds and `min` were real value, this helped to understand what happened. This should be merged IMO. This can aswell add some precision on the check and is always helpful.

muhamadazmy:
@maxux current seektime version in zos return this

```
seektime -j /dev/vda
{"device": "/dev/vda", "type": "HDD", "elapsed": 564}
```
if the format is going to change then zos also need to have some modification to support his.

Right now it seems that seektime is only provided by the base image. so after merging of your branch, and if zos image is rebuilt, then zos will break. Instead what we need to do is provide a runtime package update for seektime so both the seektime version and the code to handle it to be installed in one go.

Can you please provide a `seektime` pre-build in the seektime release. A static build is the best or in worst case make sure it uses the same libc version of the base zos image.

Once this build is available, an flist can be built for seektime that can be included in zos

muhamadazmy (approved):

PR #2031, opened 2023-08-25
wip: sync power state

### LINKED ISSUE #2029: Nodes didn't respond to power target event
We have a report from a farmer that some of their nodes did not shut down as expected when instructed to do so by the farmerbot. The affected node ids are: 5800, 5943, 5951, 5952, 5954, 5955, 5956.

I reviewed the logs for node 5800 and correlated the events with the logs from the farmer's farmerbot.

Here, the farmerbot is setting the power target to down:

```
2023-08-24 13:45:55 [INFO ] [POWERMANAGER] Executing job: POWEROFF 5800
```

And the node seems to receive the event:

```
2023-08-24 15:46:00 | [+] noded: 2023-08-24T13:46:00Z info got power target change event node=5800
```

The node's clock seems to be off by two hours. Compare the wakeup event from the farmerbot:

```
2023-08-24 13:03:06 [INFO ] [POWERMANAGER] Executing job: POWERON 5800
```

With the timestamps from the first grouping of logs submitted by the node today below. I've selected logs related to ntp in case there's something of note there (ntpd times out twice more in the available logs):

```
2023-08-24 15:10:04 | [-] ntp: ntpd: timed out waiting for 185.238.130.233, reach 0x00, next query in 32s
2023-08-24 15:10:04 | [-] ntp: ntpd: setting time to 2023-08-24 13:08:29.135673 (offset -1.158783s)
2023-08-24 15:10:04 | [-] ntp: ntpd: bad address '0.pool.ntp.org'
2023-08-24 15:10:04 | [-] ntp: ntpd: bad address '1.pool.ntp.org'
2023-08-24 15:10:04 | [-] ntp: ntpd: bad address '2.pool.ntp.org'
2023-08-24 15:10:04 | [-] ntp: ntpd: bad address '3.pool.ntp.org'
2023-08-24 15:10:04 | [-] ntp: ntpd: bad address '0.pool.ntp.org'
2023-08-24 15:10:04 | [-] ntp: ntpd: bad address '1.pool.ntp.org'
2023-08-24 15:10:04 | [-] ntp: ntpd: bad address '2.pool.ntp.org'
2023-08-24 15:10:04 | [-] ntp: ntpd: bad address '3.pool.ntp.org'
```

There are also a number of errors about connecting to TF Chain, for example:

```
2023-08-24 15:46:05 | [+] powerd: 2023-08-24T13:46:05Z error failed to connect to endpoint, retrying error="error connecting to substrate at 'wss://03.tfchain.grid.tf/': dial tcp: lookup 03.tfchain.grid.tf on [fe80::8efd:deff:fe1c:63f7%zos]:53: server misbehaving"
2023-08-24 15:40:50 | [+] noded: 2023-08-24T13:40:50Z error failed to reactivate account error="error connecting to substrate at 'wss://03.tfchain.grid.tf/': dial tcp: lookup 03.tfchain.grid.tf on [fe80::8efd:deff:fe1c:63f7%zos]:53: dial udp [fe80::8efd:deff:fe1c:63f7%zos]:53: connect: invalid argument"
```

That said, it does look like the event is coming through as seen above, so this wouldn't explain why the node didn't shut down.

muhamadazmy:
This to make sure that the state is always synced
even if handling of an event fails!

Also make sure error processing an event is logged

Fixes #2029

### DIFF
--- a/pkg/events/redis.go
+++ b/pkg/events/redis.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
+	"time"
 
 	"github.com/centrifuge/go-substrate-rpc-client/v4/types"
 	"github.com/gomodule/redigo/redis"
@@ -242,6 +243,12 @@ func (r *RedisConsumer) consumer(ctx context.Context, stream string, ch reflect.
 			messages, err := r.pop(con, group, stream)
 			if err != nil {
 				logger.Error().Err(err).Msg("failed to get events from")
+				select {
+				case <-ctx.Done():
+					return
+				case <-time.After(2 * time.Second):
+					continue
+				}
 			}
 
 			for _, message := range messages {
--- a/pkg/power/power.go
+++ b/pkg/power/power.go
@@ -274,17 +274,28 @@ func (p *PowerServer) setNodePowerState(cl *substrate.Substrate, up bool) error 
 
 func (p *PowerServer) recv(ctx context.Context) error {
 	log.Info().Msg("listening for power events")
-	stream, err := p.consumer.PowerTargetChange(ctx)
+
+	if err := p.syncSelf(); err != nil {
+		return errors.Wrap(err, "failed to synchronize power status")
+	}
+
+	subCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	stream, err := p.consumer.PowerTargetChange(subCtx)
 	if err != nil {
-		return errors.Wrapf(errConnectionError, "failed to connect to zbus events: %s", err)
+		return errors.Wrap(err, "failed to connect to zbus events")
 	}
 
 	for event := range stream {
 		if err := p.event(&event); err != nil {
-			log.Error().Err(err).Msg("failed to process power event")
+			return errors.Wrap(err, "failed to process power event")
 		}
 	}
 
+	// if we reach here it means stream was ended. this can only happen
+	// if and only if the steam was over and that can only be via a ctx
+	// cancel.
 	return nil
 }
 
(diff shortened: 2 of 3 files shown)


pkg/events/redis.go
```diff
@@ -242,6 +243,12 @@ func (r *RedisConsumer) consumer(ctx context.Context, stream string, ch reflect.
 			messages, err := r.pop(con, group, stream)
 			if err != nil {
 				logger.Error().Err(err).Msg("failed to get events from")
+				select {
+				case <-ctx.Done():
+					return
+				case <-time.After(2 * time.Second):
```
xmonader, line 249: What is the significance of 2 seconds in here? why not `default: continue` ?

pkg/power/power.go
```diff
@@ -274,43 +270,50 @@ func (p *PowerServer) setNodePowerState(cl *substrate.Substrate, up bool) error
...
-			return err
+		if err != nil {
+			log.Error().Err(err).Msg("failed to process power events")
 		}
 
 		select {
 		case <-ctx.Done():
 			return ctx.Err()
-		case <-time.After(10 * time.Second):
+		case <-time.After(5 * time.Second):
```
xmonader, line 316: Also wonder how we choose the numbers 10 and 5 in general for this

pkg/events/redis.go
muhamadazmy: it's just a prevention to avoid getting stuck in a busy loop. I was thinking of differentiate also between normal error and connection errors and in the later case the entire loop should exit because that's a fatal error.

pkg/power/power.go
muhamadazmy: There is no specific rule, the recv() function in general should never return, but if it does then it's probably a connection error to the chain, so there is no need to retry immediately, instead it should wait a little bit longer. seems i did 10 seconds before but now i just feel it's too long. The number of seconds to wait is just an arbitrary not too long, not to short wait to give the system time to recover

Issue #2000, opened 2023-07-11
zos upgrade service does not leverage on local 0-fs cache


muhamadazmy:
the zos upgrade service is implemented in Go. to download the system binaries from the flists it uses a 0-fs implementation in go embedded in the process itself, this way it doesn't have to use or rely on an external binary during upgrade.

ZOS on the otherhand uses a rust implementation of 0-fs (rfs) to serve workload and those mounts runs as separate process. rfs uses the local disk cache to make sure workload flist files are cached locally. The go implementation of 0-fs also used a cache but it's not compatible with the Go implementation. This mean the go and rust implementation can't use the same cache directory. 

To allow upgrade to use cache we can have one of 2 fixes:
- Enable the cache for upgrade on a different location under on the same cache disk
- Implement the cache in a similar format as used by the rust implementation and then both implementation can use the same cache directory

Note that both require some work

muhamadazmy:
@MarioBassem can we have some updates on this one ?

PR #2032, opened 2023-08-28
add block cache for upgrade service

mariobassem:
## Related issues:

- #2000


pkg/upgrade/cache.go
```diff
@@ -0,0 +1,264 @@
...
+	"golang.org/x/crypto/blake2b"
+
+	"golang.org/x/sync/errgroup"
+)
+
+const (
+	// DefaultDownloadWorkers define the default number of workload to use to downloads data blocks
+	DefaultDownloadWorkers = 4
+	//DefaultBlockSize is the default block size
+	DefaultBlockSize = 512 //KB
```
xmonader: Please add unit suffix e.g DefaultBlockSizeKB

xmonader: also noticed these constants are also used internally in functions and there's no way to override them, please make them unexported

pkg/upgrade/cache.go
```diff
@@ -0,0 +1,264 @@
...
+)
+
+// Downloader allows to get some data blocks using a pool of workers
+type Downloader struct {
+	cache     string
```
xmonader: is that cacheDir?

pkg/upgrade/cache.go
```diff
@@ -0,0 +1,264 @@
...
+	workers   int
```
xmonader: change to numWorkers

pkg/upgrade/cache.go
```diff
@@ -0,0 +1,264 @@
...
+	storage   storage.Storage
+	blocks    []meta.BlockInfo
+	blockSize uint64
```
xmonader: add unit here please, Is it byte, kb, .. etc

pkg/upgrade/cache.go
```diff
@@ -0,0 +1,264 @@
...
+)
+
+const (
+	// DefaultDownloadWorkers define the default number of workload to use to downloads data blocks
+	DefaultDownloadWorkers = 4
+	//DefaultBlockSize is the default block size
+	DefaultBlockSize = 512 //KB
+)
+
+// Downloader allows to get some data blocks using a pool of workers
```
xmonader: This is a concrete type, I believe the documentation needs to be more specific

pkg/upgrade/cache.go
```diff
@@ -43,11 +43,11 @@ type DownloaderOption interface {
 	apply(d *Downloader)
```
xmonader: it's not done like that, please make sure to remove the interface, 

you can have a new type e.g DownloaderOption that's a mutation method func(d  *Downloader)

and create function e.g WtihNumWorkers e.g

```go
func WithNumWorkers(n int) func(*Downloader) {
            return func(d *Downloader){ 
                    d.numWorkers = n
            
            }
}

```


Please don't introduce interfaces that're not needed

xmonader (changes requested):

pkg/upgrade/cache.go
xmonader: we use that in order not to break the API, while this is all internal it's fine to have the numworker in the factory function, or even if the type is simple an depends on the zero value you can directly use it and have the fields exported without tons of chores.

muhamadazmy (dismissed):

xmonader (approved):

Issue #2035, opened 2023-08-29
on nodes with public config, an iperf server need to run inside the public name space


muhamadazmy:
A node with public config need to make sure an `iperf` server instance is always running inside the public namespace. This simply can be done during the public config setup to also ensure a zinit service is auto-configured and monitored to start iperf service inside `public` namespace

PR #2042, opened 2023-08-30
add iperf service

rawdaGastan:
### Issues

- https://github.com/threefoldtech/zos/issues/2035

### DIFF
--- a/pkg/network/iperf/iperf.go
+++ b/pkg/network/iperf/iperf.go
@@ -2,7 +2,6 @@ package iperf
 
 import (
 	"os/exec"
-	"time"
 
 	"github.com/pkg/errors"
 	"github.com/threefoldtech/zos/pkg/zinit"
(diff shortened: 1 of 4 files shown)


muhamadazmy (changes requested):
I think you over engineered this a little bit. 
- The zinit service for the iperf just need to be "added" to zinit on public.EnsurePublicSetup.
- There is no need to do `iperf` stop or start, if the service exists then nothing to be done. We never need to restart it even if public config changes
- If public config is removed, the node is rebooted and on next boot it will have no public config so the service will not be added either so nothing to do.
- In zinit a (monitor) does automatically start so a call to start or start wait is totally not needed. It's up to the zinit to start and maintain it, not you.
- I would add `networkd` in the After of the zinit. 
- Finally and the most important is that `iperf` is not started inside the public namespace instead the command in the zinit file should be `ip netns exec public iperf -s` 
-

pkg/network/iperf/iperf.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/pkg/errors"
+	"github.com/threefoldtech/zos/pkg/zinit"
+)
+
+const (
+	zinitService = "iperf"
+)
+
+// IPerfServer represent an iperf server
+type IPerfServer struct {
```
muhamadazmy: I will replace all of this with one function `Ensure` that checks if the zinit unit file for iperf exists, and if not add one and call monitor and that's it. Other functions are not needed (and can be then added later if needed, but don't see why would thay) 
then a caller can directly call `iperf.Ensure` where iperf is the package name (Ensure is a function not an instance method) no need to have an empty struct

pkg/network/iperf/iperf.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/pkg/errors"
+	"github.com/threefoldtech/zos/pkg/zinit"
+)
+
+const (
+	zinitService = "iperf"
+)
+
+// Ensure creates an iperf zinit service and monitors it
+func Ensure(z *zinit.Client) error {
```
muhamadazmy, line 18: The Ensure makes sure that the service exist, right? what if it's already there. 
The right way is u first need to check the status, add it if it does not exist

muhamadazmy (changes requested):

pkg/network/iperf/iperf.go
```diff
@@ -0,0 +1,39 @@
...
+	if _, err := z.Status(zinitService); err == nil {
+		return nil
+	}
+
+	_, err := exec.LookPath("iperf")
+	if err != nil {
+		return err
+	}
+
+	cmd := `ip netns exec public iperf -s`
```
muhamadazmy: as discussed, may be it's better to add the port explicitly in the command line. (from a constant) that same constant need to be reserved in networkd as i showed you over telegram

muhamadazmy (approved):

Issue #2036, opened 2023-08-29
[perf] research cpu performance tools


muhamadazmy:
The tool need to do some shallow tests that does not affect the performance of running workloads but give an estimation of CPU pefroamnce.

During building of the peformance report (after running the perf tool) the reporter must also include the amount of workloads that was hosted on the node during test running

AbdelrahmanElawady:
decided not to go with `perf` binary instead just stress test some of the available CPUs and report load with the number of running workloads.
waiting on #2046

muhamadazmy:
You already can start by creating the required binary package

PR #2045, opened 2023-09-03
Add perf binary

AbdelrahmanElawady:
## Issues

- #2036


muhamadazmy:
We probably not going with `perf` it's not what we want for that case. I will close for now until we are done with investigations

Issue #2037, opened 2023-08-29
perf tests framework inside noded


muhamadazmy:
Noded can have an extendable performance tests modules where they can be scheduled to run. Each test can then update it's cashed results in redis.

The perf framework then can expose rmb handler for performance queries that can return, all test results or specific tests.

PR #2046, opened 2023-09-03
performance monitor package

Omarabdul3ziz:
related issues:
- https://github.com/threefoldtech/zos/issues/2037


xmonader (changes requested):
make sure to enable linting on your editor before making this ready for review

cmds/modules/noded/main.go
```diff
@@ -193,6 +194,17 @@ func action(cli *cli.Context) error {
 		return hypervisor, nil
 	})
 
+	log.Info().Msg("Start Perf scheduler")
```
xmonader: use lowercase in the message

cmds/modules/noded/main.go
```diff
@@ -193,6 +194,17 @@ func action(cli *cli.Context) error {
...
+	performanceMonitor := perf.NewPerformanceMonitor("/var/run/redis.sock")
```
xmonader: prefer to use shorter var names e.g perfMon

pkg/perf/cache.go
```diff
@@ -0,0 +1,49 @@
...
+type GetRequest struct {
+	TestName   string `json:"test_name"`
+	TestNumber uint64 `json:"test_number"`
+}
+
+// CacheResult set result in redis
+func (pm *PerformanceMonitor) CacheResult(ctx context.Context, resultKey string, resultData TestResultData) error {
+	data, err := json.Marshal(resultData)
+	if err != nil {
+		return errors.Wrap(err, "Error marshaling data to JSON")
```
xmonader: errors should be lowercased

pkg/perf/cache.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	return pm.RedisClient.Set(resultKey, data, 10*time.Second).Err()
```
xmonader: extract the expiration into an unexported constant `resultTTL`, also 10 is too low imo,

pkg/perf/tasks.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/rs/zerolog/log"
+)
+
+// TestLogging simple helper method that ensure scheduler is working
+func TestLogging(ctx context.Context) (interface{}, error) {
```
xmonader: let's remove this one, there're other logs we can see.

pkg/perf/monitor.go
```diff
@@ -0,0 +1,100 @@
...
+	"encoding/json"
+	"time"
+
+	"github.com/go-redis/redis"
+	"github.com/jasonlvhit/gocron"
+	"github.com/pkg/errors"
+)
+
+// TaskMethod is the task method signature
+type TaskMethod func(ctx context.Context) (interface{}, error)
```
xmonader: TaskAction or TaskFunc

pkg/perf/monitor.go
```diff
@@ -0,0 +1,100 @@
...
+
+type Task struct {
+	Key            string
+	Method         TaskMethod
+	Interval       time.Duration // interval in seconds
```
xmonader: IntervalSeconds

pkg/perf/monitor.go
```diff
@@ -0,0 +1,100 @@
...
+	ExecutionCount uint64
```
xmonader: What does this mean?

pkg/perf/monitor.go
```diff
@@ -0,0 +1,100 @@
...
+	pm.Tasks = append(pm.Tasks, Task{
+		Key:            taskName,
+		Method:         task,
+		Interval:       interval,
+		ExecutionCount: 0,
+	})
+}
+
+// InitScheduler adds all the test to the scheduler queue
+func (pm *PerformanceMonitor) InitScheduler() {
```
xmonader: remove this

pkg/perf/monitor.go
```diff
@@ -0,0 +1,100 @@
...
+		err := pm.Scheduler.Every(uint64(task.Interval)).Seconds().Do(func() error {
+			testResult, err := task.Method(ctx)
+			if err != nil {
+				return errors.Wrapf(err, "failed running test: %s", task.Key)
+			}
+
+			task.ExecutionCount++
+			err = pm.CacheResult(ctx, task.Key, TestResultData{
+				TestName:   task.Key,
+				TestNumber: task.ExecutionCount,
```
xmonader: okay now I understand the point of TestNumber, Please remove it from the test data completely, we only care about the latest test results. it doesn't mean much given it gets invalidated every 10 seconds according to the current code, so there's no point in storing that.

pkg/perf/monitor.go
```diff
@@ -0,0 +1,100 @@
...
+
+// InitScheduler adds all the test to the scheduler queue
+func (pm *PerformanceMonitor) InitScheduler() {
+	pm.AddTask("TestLogging", 5, TestLogging)
+}
+
+// RunScheduler adds the tasks to the corn queue and start the scheduler
+func (pm *PerformanceMonitor) RunScheduler(ctx context.Context) error {
+	for _, task := range pm.Tasks {
+		err := pm.Scheduler.Every(uint64(task.Interval)).Seconds().Do(func() error {
```
xmonader: I don't see the added value of using go-cron, simple goroutines and error groups should have been more than enough?

pkg/perf/monitor.go
```diff
@@ -0,0 +1,100 @@
...
+	})
+}
+
+// InitScheduler adds all the test to the scheduler queue
+func (pm *PerformanceMonitor) InitScheduler() {
+	pm.AddTask("TestLogging", 5, TestLogging)
+}
+
+// RunScheduler adds the tasks to the corn queue and start the scheduler
+func (pm *PerformanceMonitor) RunScheduler(ctx context.Context) error {
```
xmonader: the name RunScheduler shouldn't be introduced, a simple Run or Start should have been enough, now whoever is using this API will need to know it's using go-cron scheduler

pkg/perf/cache.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"context"
+	"encoding/json"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+// TestResultData the result test schema
+type TestResultData struct {
```
xmonader: Now after checking the rest of the files, this should be TaskResult struct only task name and the result

pkg/perf/cache.go
```diff
@@ -0,0 +1,49 @@
...
+	TestName   string
+	TestNumber uint64
+	Result     interface{}
+}
+
+// GetRequest the get request struct
+type GetRequest struct {
```
xmonader: this struct can be removed completely and use the task name always to get the task result from the cache

qemu/Makefile
```diff
@@ -14,3 +14,16 @@ start:
...
 	bash vm.sh -n node1 -c "runmode=test farmer_id=$(FARMERID)"
+
+auth: 
+	@echo "Copying your public ssh to machine rootfs"
+	mkdir -p overlay/root/.ssh
+	cp ~/.ssh/id_rsa.pub overlay/root/.ssh/authorized_keys
+
+net:
+	@echo "Creating a virtual natted network"
+	bash ../docs/development/net.sh
```
muhamadazmy: Can the script be instead moved here, and docs then refers to it in this location instead of the other way around?

pkg/perf/monitor.go
```diff
@@ -0,0 +1,100 @@
...
+// PerformanceMonitor holds the module data
+type PerformanceMonitor struct {
+	Scheduler   *gocron.Scheduler
+	RedisClient *redis.Client
+	Tasks       []Task
+}
+
+// NewPerformanceMonitor returns PerformanceMonitor instance
+func NewPerformanceMonitor(redisAddr string) *PerformanceMonitor {
+	redisClient := redis.NewClient(&redis.Options{
```
muhamadazmy: We already using a `redis` package in zos so no need to add a new dependency. Please move to `github.com/gomodule/redigo` you can already create a new pool directly by calling this util function here https://github.com/threefoldtech/zos/blob/main_perf_package/pkg/utils/redis.go#L11

where the url is just `unix://path/to/socket`

pkg/perf/monitor.go
```diff
@@ -0,0 +1,100 @@
...
+	Method         TaskMethod
+	Interval       time.Duration // interval in seconds
+	ExecutionCount uint64
+}
+
+// PerformanceMonitor holds the module data
+type PerformanceMonitor struct {
+	Scheduler   *gocron.Scheduler
+	RedisClient *redis.Client
+	Tasks       []Task
```
muhamadazmy: Why these fields are public? they open the door to bad design since you will end up accessing an `internal` implementation details to an object. A better approach is that all fields become private and then functionality is either defined by interfaces or public methods. 

Some fields can become public of course, but usually in Data only types. Not operational types. For example there is noway someone will ever need to access the redis client on the Monitor structure because that is implementation details of how the Monitor caches the result!

pkg/perf/monitor.go
muhamadazmy: it's  wrong to use `Duration` in seconds. Duration is represented internally in nano seconds! Means all Golang std and other libraries that accepts Duration it's ALWAYS in nano-seconds and you can't break this contract.

then a duration of 5 seconds is ` dur = 5 * time.Second`  dur in this case. 

If you want a type that represents time in Seconds then use a `uint16` or any unassigned numerical type. or create a new type `type Seconds uint32`

Another way to do this is make duration a private field, then the New function will then take `uint32` (in seconds) and does the following
```go
type Test {
dur Duration
}

func NewTest(seconds uint32) Test {
   return Test {
    dur: seconds * time.Seconds 
   }
}
```
```

File: pkg/perf/monitor.go [hunk 18]
```diff
@@ -0,0 +1,100 @@
...
+
+	return &PerformanceMonitor{
+		Scheduler:   scheduler,
+		RedisClient: redisClient,
+		Tasks:       []Task{},
+	}
+}
+
+// AddTask a simple helper method to add new tasks
+func (pm *PerformanceMonitor) AddTask(taskName string, interval time.Duration, task TaskMethod) {
```
muhamadazmy: The point behind using `gocron` is not giving it an `interval` but instead a `cron` specs instead. I don't want to specify how often does it run but instead `when` it runs using the cron syntax.

A task can be secduled to run every sunday at 12pm, or 3 times per day on 0 8 and 16, etc.. So specifying an interval is not the right way.

I also don't like that the task is `TaskMethod` is just a function not an interface. This already forces you to create internal types to maintain the "task" why not instead use an interface as

```
type Task interface {
  ID() string
  Cron() string 
  Run(ctx context.Context) (any, error)
}
```

pkg/perf/cache.go
```diff
@@ -0,0 +1,49 @@
...
+	TestName   string `json:"test_name"`
+	TestNumber uint64 `json:"test_number"`
+}
+
+// CacheResult set result in redis
+func (pm *PerformanceMonitor) CacheResult(ctx context.Context, resultKey string, resultData TestResultData) error {
```
muhamadazmy: Why is this a `public` method. I don't expect anyone should ever call this method by himself. this should an internal method. I see many attributes and methods that are `public` for no obvious reason. It make the code subtle to abuse and bad usage. And allows bad design

pkg/perf/cache.go
muhamadazmy: Was gonna say that data should never expire, but instead have the `timestamp` of when the test was run. but never expire. Some tests run every 6 hours our once per day. Hence expiration is not needed. so `0` is the right value

pkg/perf/monitor.go
```diff
@@ -0,0 +1,100 @@
...
+			return errors.Wrapf(err, "failed scheduling the job: %s", task.Key)
+		}
+	}
+
+	pm.Scheduler.Start()
+	return nil
+}
+
+// Get handles the request to get data from cache
+func (pm *PerformanceMonitor) Get(ctx context.Context, payload []byte) (interface{}, error) {
```
muhamadazmy: This is so wrong! Why a function in a library knows about an API request format ? A library should only expose a library interface (Say a Get(test)) and that' it

An public API interface (http, rmb, or whatever) then can handle user requests and then just make a proper call to the library

pkg/perf/cache.go
```diff
@@ -0,0 +1,49 @@
...
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+// TestResultData the result test schema
+type TestResultData struct {
+	TestName   string
+	TestNumber uint64
+	Result     interface{}
```
muhamadazmy: fields missing json tags

cmds/modules/noded/main.go
```diff
@@ -193,6 +194,17 @@ func action(cli *cli.Context) error {
...
+	performanceMonitor.InitScheduler()
+	err = performanceMonitor.RunScheduler(ctx)
+	if err != nil {
+		log.Error().Err(err).Msg("fails in scheduler")
+	}
+	bus.WithHandler("zos.perf.get", func(ctx context.Context, payload []byte) (interface{}, error) {
+		return performanceMonitor.Get(ctx, payload)
```
muhamadazmy: handling of user input is part of the `public` API and then it should handled here not in the performanceMonitor library or object. 

This breaks `separation of concerns` and `isolation` of your design.

pkg/perf/monitor.go
muhamadazmy: I already commented on the AddTask method regarding exactly that

muhamadazmy (changes requested):

cmds/modules/noded/main.go
```diff
@@ -193,6 +195,27 @@ func action(cli *cli.Context) error {
 		return hypervisor, nil
 	})
 
+	log.Info().Msg("start perf scheduler")
+	perfMon, err := perf.NewPerformanceMonitor("unix://var/run/redis.sock")
```
muhamadazmy: Shouldn't be hard coded, specially that we already have a command line argument for that

cmds/modules/noded/main.go
```diff
@@ -193,6 +195,27 @@ func action(cli *cli.Context) error {
...
+	if err != nil {
+		return errors.Wrap(err, "failed to create a new perfMon")
+	}
+	defer perfMon.Close()
+
+	err = perfMon.Run(ctx)
+	if err != nil {
+		log.Error().Err(err).Msg("failed to run the scheduler")
+	}
```
muhamadazmy:
```suggestion
	
	if err := perfMon.Run(ctx); err != nil {
		log.Error().Err(err).Msg("failed to run the scheduler")
	}
```

Also i think it's better if we return the error not just log

cmds/modules/noded/main.go
```diff
@@ -193,6 +195,27 @@ func action(cli *cli.Context) error {
...
+	bus.WithHandler("zos.perf.get", func(ctx context.Context, payload []byte) (interface{}, error) {
+		var taskName string
+		err := json.Unmarshal(payload, &taskName)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to unmarshal payload: %v", payload)
+		}
+
+		return perfMon.GetCache(taskName)
```
muhamadazmy: I think the API need to support returning all available tests, or specific one by name. so this function need to support not having a `test-name` and that name it call be results from all. 

Also the `GetCache` is not a good name. Again this exposes the details of an implementation. The user of the API does not care where the result is stored or if the test is executed on request. Hence this should only called `Get`

pkg/perf/cache.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// setCache set result in redis
+func (pm *PerformanceMonitor) setCache(ctx context.Context, taskName string, result TaskResult) error {
+	data, err := json.Marshal(result)
+	if err != nil {
+		return errors.Wrap(err, "failed to marshal data to JSON")
+	}
+
+	_, err = pm.redisConn.Do("SET", taskName, data)
```
muhamadazmy: Using the task name alone as a key is not a good idea. The same redis instance is used by other apps on the system to store and queue data. Hence you need to create some sort of a `namespacing` for your data. for example prefix all keys with `perf.` so for example a `cpu` test key will be `perf.cpu` instead of `cpu` so it has less chance of having collisions with other apps

Having a prefix also makes it easier for range scan. For example if u wanna return ALL tests u can scan KEYS with `perf.*` to return all results for all tests

pkg/perf/cache.go
```diff
@@ -0,0 +1,43 @@
...
+	return err
+}
+
+// GetCache get data from redis
+func (pm *PerformanceMonitor) GetCache(taskName string) (TaskResult, error) {
```
muhamadazmy: As commented before, the GetCache is not correct. The performance framework internal details (how tests are run or how results are retrieved) is only an implementation details. The PerformanceMonitor struct then API should be `Get` for a result.

In other words, why a user of this framework know that the result is cached for example

pkg/perf/monitor.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// NewPerformanceMonitor returns PerformanceMonitor instance
+func NewPerformanceMonitor(redisAddr string) (*PerformanceMonitor, error) {
+	redisPool, err := utils.NewRedisPool(redisAddr)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed creating new redis pool")
+	}
+
+	redisConn := redisPool.Get()
```
muhamadazmy: This is not how u use the redis Pool. The idea of a Pool of connection that u Get a connection from the pool, use it and then close it immediately when you are done using it. 

Closing a connection will return the connection to the pool (the pool might not close the connection) but keep it alive for the next one to need a connection (by calling the pool.Get()) 

So what u are doing is breaking the pooling contract by hording a connection forever. This is also really bad because if the connection is interrupted (or lost because of any reason) there is NO way that u renew the connection. The pool on the other hand tests the connection before it give u one when u do a Get, and renew it if needed.

In summary, what u need to do is to keep the pool on you structure (not a connection) and in you setCache and Get methods what u need to do is

```go
con := pm.pool.Get()
defer con.Clonse()

// use the con to get or set
```

pkg/perf/cache.go
```diff
@@ -0,0 +1,43 @@
...
+	"context"
+	"encoding/json"
+
+	"github.com/pkg/errors"
+)
+
+// TaskResult the result test schema
+type TaskResult struct {
+	TaskName     string      `json:"task_name"`
+	RunTimestamp string      `json:"run_timestamp"`
```
muhamadazmy: Please change this to an actual unix timpestamp uint64. Storing time as a string is the most inefficient way to store time.

pkg/perf/cache.go
```diff
@@ -0,0 +1,43 @@
...
+	Result       interface{} `json:"result"`
+}
+
+// setCache set result in redis
+func (pm *PerformanceMonitor) setCache(ctx context.Context, taskName string, result TaskResult) error {
```
muhamadazmy: Why do u pass taskName when the TaskResult already have this?

muhamadazmy (changes requested):

cmds/modules/noded/main.go
```diff
@@ -193,6 +195,30 @@ func action(cli *cli.Context) error {
...
+		err := json.Unmarshal(payload, &taskName)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to unmarshal payload: %v", payload)
+		}
+
+		if taskName == "" {
+			return perfMon.GetAll()
+		}
+
+		return perfMon.Get(taskName)
```
muhamadazmy: While this will work perfectly will in Go . I don't like the fact that the output type changes based on the input (one returns a single object, the other returns a list) which makes it annoying to call from languages that supports static type.

This can be solved in 2 possible ways
- have 2 functions `zos.perf.get` and `zos.perf.get_all` 
- Return a unified type, say list of tests which can be a list of a single test if specific name is given, or all.

A 3rd solution would be always return ALL? and then client can filter the test on client side is also a possible option

pkg/perf/cache.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/pkg/errors"
+)
+
+const (
+	moduleName = "perf"
+)
+
+// TaskResult the result test schema
+type TaskResult struct {
+	TaskName  string      `json:"task_name"`
```
muhamadazmy: The type is already name TaskResult, so I can perfectly call this `Name`
```suggestion
	Name  string      `json:"name"`
```

pkg/perf/cache.go
```diff
@@ -0,0 +1,103 @@
...
+			if err != nil {
+				continue
+			}
+
+			var result TaskResult
+
+			err = json.Unmarshal(value.([]byte), &result)
+			if err != nil {
+				return res, errors.Wrap(err, "failed to unmarshal data from json")
+			}
```
muhamadazmy: Why not just call `pm.Get()`

muhamadazmy (changes requested):

pkg/perf/monitor.go
```diff
@@ -0,0 +1,86 @@
...
+		ok, err := pm.exists(task.ID())
+		if err != nil {
+			return errors.Wrapf(err, "failed to find key %s", task.ID())
+		}
+
+		if !ok {
+			if err := pm.runTask(ctx, task); err != nil {
+				return errors.Wrapf(err, "failed to run task: %s", task.ID())
+			}
+		}
```
muhamadazmy: Shouldn't this be run asynchronously so you don't block registration and running of other tests. They all can run in parallel

muhamadazmy (changes requested):
Looks good except one small comment

pkg/perf/monitor.go
```diff
@@ -0,0 +1,86 @@
...
+	redisPool, err := utils.NewRedisPool(redisAddr)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed creating new redis pool")
+	}
+
+	scheduler := gocron.NewScheduler(time.UTC)
+
+	return &PerformanceMonitor{
+		scheduler: scheduler,
+		pool:      *redisPool,
```
muhamadazmy: You can't do that here, this fails already in CI. redisPool is already a pointer right? and that's is exactly what u need to keep on the PerfomanceMonitor structure

pkg/perf/monitor.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/gomodule/redigo/redis"
+
+	"github.com/pkg/errors"
+	"github.com/threefoldtech/zos/pkg/utils"
+)
+
+// PerformanceMonitor holds the module data
+type PerformanceMonitor struct {
+	scheduler *gocron.Scheduler
+	pool      redis.Pool
```
muhamadazmy:
```suggestion
	pool      *redis.Pool
```
check the other comment to see why

pkg/perf/monitor.go
muhamadazmy: The reason this is done because when u do a `deref` (with `*`) on a point value it takes a copy. Pools (at least this one) is not designed to be copied other wise u will end up with 2 pools that don't share internal state. Instead you need always to use a pointer to the same pool hence copying the value is not good.

It's not caught by the compiler but by the CI tests

muhamadazmy (changes requested):

pkg/perf/monitor.go
```diff
@@ -0,0 +1,105 @@
...
+					routineErr = errors.Wrapf(err, "failed to run task: %s", task.ID())
+				}
+			}
+			errChan <- routineErr
+		}(task)
+
+	}
+
+	wg.Wait()
+	close(errChan)
```
muhamadazmy: I am wondering actually if we need to `wait` until the tests are done ? why do they need to be handled any different than when the `cron` starts them.

This will slow down the start of the main service (blocking on Run()) for a very long time on first run. What i wanted to do is that if there are no cached results u `start` the tests in the background to have some results, and then leave it to the shceulder for later runs. Very simple, there should no be any blocking in Run in this case, if an perf test failed u should only see a log in the daemon logs but that's it.

muhamadazmy (approved):

Issue #2050, opened 2023-09-07
implement a network test against public nodes


muhamadazmy:
Now since public nodes are running iperf server by default, we can then implement a perf test from all nodes by choosing random public nodes.
- We also probably need to keep a list of fixed public nodes maintained by us for speed tests
- Select few random ones
- The test should follow the same schedule as described by the story card.

PR #2052, opened 2023-09-11
Main perf tests

rawdaGastan:
### Issues

- https://github.com/threefoldtech/zos/issues/2050

### example output

[result.txt](https://github.com/threefoldtech/zos/files/12727833/result.txt)


rawdaGastan:
waiting for 
- https://github.com/threefoldtech/zos/pull/2046

pkg/graphql/graphql.go
```diff
@@ -0,0 +1,68 @@
+// Package graphql for grid graphql support
+package graphql
```
muhamadazmy: Did we really needed to implement this from scratch ? can't we use an existing graphql client ?

cmds/modules/noded/main.go
```diff
@@ -202,6 +205,41 @@ func action(cli *cli.Context) error {
 		return errors.Wrap(err, "failed to create a new perfMon")
 	}
 
+	if exists := iperf.Exists(zinit.Default()); exists {
```
muhamadazmy: As discussed before running a tests against public nodes happens from ALL nodes (even the ones without public config). The idea is:
- Public nodes running iperf running in test mode 
- All nodes run test against the nodes that has iperf server running 
- Nodes measure their up/down speed against the random public nodes.

cmds/modules/noded/main.go
```diff
@@ -202,6 +205,41 @@ func action(cli *cli.Context) error {
...
+		freeFarmNodes, err := g.ListPublicNodes(0, 1, true, false)
+		if err != nil {
+			return errors.Wrap(err, "failed to list freefarm nodes from graphql")
+		}
+
+		nodes, err := g.ListPublicNodes(12, 0, true, false)
+		if err != nil {
+			return errors.Wrap(err, "failed to list nodes from graphql")
+		}
+
```
muhamadazmy: This is too much to be done by noded. Noded all what it cares about is to register a certain test against the perf framework. 

The test ON RUNNING, will pick the nodes to test against. means the nodes can be different each time the test runs

cmds/modules/noded/main.go
```diff
@@ -202,6 +205,41 @@ func action(cli *cli.Context) error {
...
+		nodes = append(nodes, freeFarmNodes...)
+
+		for _, node := range nodes {
+			perfMon.AddTask(&perf.TCPTask{
+				TaskID:    "TCPTask",
+				Schedule:  "* */5 * * * *",
+				ClientIP:  strings.SplitN(node.PublicConfig.Ipv4, "/", 2)[0],
+				Bandwidth: "1M",
```
muhamadazmy: While it's correct, but I really don't like this way to construct types. Instead you should create a `NewTest` function that takes no argument (in this case) because all this stuff should be defined by the test itself, and should be modified by the test plugin (for example iperf test has specific fixed id and schedule)

Also we should have one single test for `iperf` not 2. It's a single test plugin that will do both tcp and udb (sequentially) and store the result of both tests. 

I think a task id can be only `iperf`. I don't know why it's called a Task here when it's not

pkg/network/iperf/iperf.go
```diff
@@ -14,9 +14,18 @@ const (
...
 
+// Exists checks if the iperf service is running
+func Exists(z *zinit.Client) bool {
+	if _, err := z.Status(zinitService); err == nil {
+		return true
+	}
+
+	return false
+}
+
```
muhamadazmy: according to my previous comment this is totally not needed

pkg/perf/monitor.go
```diff
@@ -29,7 +29,7 @@ func NewPerformanceMonitor(redisAddr string) (*PerformanceMonitor, error) {
 
 	return &PerformanceMonitor{
 		scheduler: scheduler,
-		pool:      *redisPool,
+		pool:      redisPool,
```
muhamadazmy: revert that because i already asked Omar to fix it on his branch

cmds/modules/noded/main.go
muhamadazmy: The tests also need to run against MANY nodes per test not a single node. I thought this is clear. so a test can take quite some time before it fiinishes (running multiple tests against multiple nodes for tcp and udp

pkg/perf/tcp_task.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/rs/zerolog/log"
+	"github.com/threefoldtech/zos/pkg/network/iperf"
+)
+
+// TCPTask is the task for iperf tcp tests
+type TCPTask struct {
+	TaskID    string
+	Schedule  string
+	Bandwidth string
+	ClientIP  string
```
muhamadazmy: These all need to be private fields (based on previous commet)

Also as explained, we need to run multiple tests against multiple nodes per run. hence the random list of nodes plust few fixed ones so u have like total of +10 nodes. then tests can run against these in sequence for bot tcp and udp, build and then return the report

pkg/perf/tcp_task.go
```diff
@@ -0,0 +1,50 @@
...
+	var stderr bytes.Buffer
+	cmd.Stderr = &stderr
+	cmd.Stdout = &out
+
+	err = cmd.Run()
+	if err != nil {
+		log.Error().Err(err).Msgf("failed to run iperf tcp task: %s", stderr.String())
+	}
+
+	return out.String(), nil
```
muhamadazmy: This is not a good structure. The data from the output need to be parse and put in a structure that describe both upload and download speeds (decent structure) not just raw output from the command.

The structure should also then have information about what node, what Ip the test was run against, did it work or not, what are the speeds, and any other useful information from the test.

pkg/perf/udp_task.go
```diff
@@ -0,0 +1,50 @@
...
+	"context"
+	"fmt"
+	"os/exec"
+
+	"github.com/rs/zerolog/log"
+	"github.com/threefoldtech/zos/pkg/network/iperf"
+)
+
+// UDPTask is the task for iperf udp tests
+type UDPTask struct {
```
muhamadazmy: This should become one with the other Test because u run both tcp and udp against each selected node and aggreate all results in one decent structure

muhamadazmy (changes requested):

pkg/perf/iperf_task.go
```diff
@@ -0,0 +1,123 @@
...
+	NodeID        uint32                       `json:"node_id"`
+	NodeIpv4      string                       `json:"node_ip"`
+	TestType      string                       `json:"test_type"`
+	Error         error                        `json:"error"`
+	CpuReport     goIperf.CpuUtilizationReport `json:"cpu_report"`
+}
+
+// NewIperfTest creates a new iperf test
+func NewIperfTest() IperfTest {
+	return IperfTest{taskID: "iperf", schedule: "* */5 * * * *"}
```
muhamadazmy: We need to run every 6 hours according to the requirements

muhamadazmy (changes requested):

pkg/perf/iperf_task.go
```diff
@@ -0,0 +1,123 @@
...
+
+	<-iperfClient.Done
+
+	return IperfResult{
+		UploadSpeed:   iperfClient.Report().End.SumSent.BitsPerSecond,
+		DownloadSpeed: iperfClient.Report().End.SumReceived.BitsPerSecond,
+		CpuReport:     iperfClient.Report().End.CpuReport,
+		NodeIpv4:      clientIP,
+		TestType:      string(iperfClient.Proto()),
+		Error:         err,
```
muhamadazmy: That's the wrong error, use the one from the Report

pkg/perf/iperf_task.go
```diff
@@ -0,0 +1,123 @@
...
+
+	return results, nil
+}
+
+func (t *IperfTest) runIperfTest(ctx context.Context, clientIP string, tcp bool) IperfResult {
+	iperfClient := goIperf.NewClient(clientIP)
+	iperfClient.SetBandwidth("1M")
+	iperfClient.SetPort(iperf.IperfPort)
+	iperfClient.SetJSON(true)
+	iperfClient.SetIncludeServer(true)
```
muhamadazmy: Not sure if this is needed

pkg/perf/iperf_task.go
```diff
@@ -0,0 +1,123 @@
...
+
+	_, err = exec.LookPath("iperf")
+	if err != nil {
+		return nil, err
+	}
+
+	var results []IperfResult
+	for _, node := range nodes {
+		clientIP := strings.SplitN(node.PublicConfig.Ipv4, "/", 2)[0]
+		clientIPv6 := strings.SplitN(node.PublicConfig.Ipv6, "/", 2)[0]
```
muhamadazmy: Possible panic because of index out of range, always validate the output that it's of the right length or use the https://pkg.go.dev/net#ParseCIDR method

pkg/perf/graphql_nodes.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// PublicConfig includes the public config information for the node
+type PublicConfig struct {
+	Ipv4 string `graphql:"ipv4"`
+	Ipv6 string `graphql:"ipv6"`
+}
+
+// ListPublicNodes returns a list of public nodes
+func (g *GraphQl) ListPublicNodes(ctx context.Context, n int, farmID, excludeFarmID uint32, ipv4, ipv6 bool) ([]Node, error) {
```
muhamadazmy: what is `n`? can u give it a proper name expand the docs

muhamadazmy (approved):

Issue #2056, opened 2023-09-15
limit number of cpus reserved in VM to max number of CPUs on host


muhamadazmy:
Since VMs uses vCPUs which are not mapped to actual cores. it's totally okay to reserve more vCPUs that number of cores that the node physically has. 
While this works fine, it might be better to avoid doing this and limit the number of CPUs to physical number of cpus on the host. 

Note: that this limitation is done per VM so a node with 8 cores can still has 2 VMs each with 8 VCPUs.

A-Harby:
Verified, 

Testes on TS client and Playground by deploying a VM with a number of CPU exceeding the node cores.

![image](https://github.com/threefoldtech/zos/assets/109473873/db666d63-2ddf-46f6-bf10-3c67253d88ec)
![image](https://github.com/threefoldtech/zos/assets/109473873/ae88888b-1c09-491c-a34f-f0674ae723b0)


Created test on TS Client suite with full steps:
https://app.testlodge.com/a/26076/projects/40893/suites/194096?expand_section=306578#case_4099871

A-Harby:
Verified, 

Testes on TS client and Playground by deploying a VM with a number of CPU exceeding the node cores.

![image](https://github.com/threefoldtech/zos/assets/109473873/db666d63-2ddf-46f6-bf10-3c67253d88ec)
![image](https://github.com/threefoldtech/zos/assets/109473873/ae88888b-1c09-491c-a34f-f0674ae723b0)


Created test on TS Client suite with full steps:
https://app.testlodge.com/a/26076/projects/40893/suites/194096?expand_section=306578#case_4099871

PR #2057, opened 2023-09-17
validate cpus according to actual number of node cpus

rawdaGastan:
### Issues

- https://github.com/threefoldtech/zos/issues/2056

### DIFF
--- a/pkg/vm.go
+++ b/pkg/vm.go
@@ -243,7 +243,7 @@ func (n *NetMetric) Nu() float64 {
 }
 
 // MachineMetric is a container for metrics from multiple networks
-// currently only groped as private (wireguard + yggdrasil), and public (public Ips)
+// currently only grouped as private (wireguard + yggdrasil), and public (public Ips)
 type MachineMetric struct {
 	Private NetMetric
 	Public  NetMetric
--- a/pkg/vm.go
+++ b/pkg/vm.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"path/filepath"
 
+	"github.com/shirou/gopsutil/cpu"
 	"github.com/threefoldtech/zos/pkg/gridtypes"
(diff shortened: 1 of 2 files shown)


pkg/vm.go
```diff
@@ -175,8 +176,16 @@ func (vm *VM) Validate() error {
 		return fmt.Errorf("invalid memory must not be less than 250M")
 	}
 
-	if vm.CPU == 0 || vm.CPU > 32 {
-		return fmt.Errorf("invalid cpu must be between 1 and 32")
+	n, err := cpu.Counts(true)
```
xmonader: can't we use runtime.NumCPU? https://pkg.go.dev/runtime#NumCPU, if so, let's not introduce a new dependency like gopsutil

pkg/vm.go
```diff
@@ -234,7 +243,7 @@ func (n *NetMetric) Nu() float64 {
 }
 
 // MachineMetric is a container for metrics from multiple networks
-// currently only groped as private (wiregaurd + yggdrasil), and public (public Ips)
+// currently only groped as private (wireguard + yggdrasil), and public (public Ips)
```
xmonader: groped -> grouped

pkg/vm.go
xmonader: n -> numCpus or ncpus or cpus

muhamadazmy: > can't we use runtime.NumCPU? https://pkg.go.dev/runtime#NumCPU, if so, let's not introduce a new dependency like gopsutil

this is not the same, according to the docs, the runtime.NumCPU is logical number of cpus that can be used by the `current` process. this is a totally Go runtime thing, not related to the total number of CPUs available on the system

muhamadazmy: But should definitely rename the variable from `n` to something else

muhamadazmy (changes requested):

muhamadazmy (approved):

Issue #2054, opened 2023-09-14
splash: add kernel version somewhere


maxux:
Would be interesting to get the kernel version (`uname -r`) somewhere on the TTY UI to have a better view of which version is running when we do some update.

A-Harby:
Verified, Devnet.
![image](https://github.com/threefoldtech/zos/assets/109473873/e530d08d-38a4-4c74-bfa5-b422e0470056)

PR #2058, opened 2023-09-17
add kernel version to zui

rawdaGastan:
### Issues

- https://github.com/threefoldtech/zos/issues/2054

### UI

Screenshot from 2023-09-17 17-39-20

### DIFF
--- a/cmds/modules/zui/header.go
+++ b/cmds/modules/zui/header.go
@@ -96,9 +96,13 @@ func headerRenderer(ctx context.Context, c zbus.Client, h *widgets.Paragraph, r 
 }
 
 func int8ToStr(arr []int8) string {
+	if len(arr) == 0 {
+		return ""
+	}
+
 	b := make([]byte, 0, len(arr))
 	for _, v := range arr {
-		if v == 0x00 {
+		if v == 0 {
 			break
 		}
 		b = append(b, byte(v))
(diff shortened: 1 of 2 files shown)


cmds/modules/zui/header.go
```diff
@@ -77,10 +79,33 @@ func headerRenderer(ctx context.Context, c zbus.Client, h *widgets.Paragraph, r
...
+
+	b := make([]byte, 0, len(arr))
+	for _, v := range arr {
+		if v == 0 {
+			break
+		}
+		b = append(b, byte(v))
+	}
+	return string(b)
+}
```
muhamadazmy: Since we know that int8 is a byte we can actually do this without copying the data. In C this would be a simple cast but this is not supported in Go

But Go still provide unsafe pointers type so this can be done like this instead
```suggestion
func int8ToStr(arr []int8) string {
        // get a pointer to the slice (this is like void* in C)
        ptr := unsafe.Pointer(&arr)
        // recast the pointer to a byte array pointer then get the value (deref) of it
        // so bytes is now a []byte not &[]byte
        bytes := *(*[]byte)(ptr)
        // create string
        return string(bytes)
        // this can be all done in one liner
        return string(*(*[]byte)(unsafe.Pointer(&x)))
}
```

muhamadazmy: test it before u update the PR

muhamadazmy (changes requested):

muhamadazmy (approved):

PR #2066, opened 2023-10-03
Add CPU benchmark task

AbdelrahmanElawady:
## Description

Add benchmark task to run 4 times a day and reports the workload numbers during the benchmark.

## Issues

- #2036

### DIFF
--- a/pkg/perf/cpubench_task.go
+++ b/pkg/perf/cpubench_task.go
@@ -1,0 +1,75 @@
+package perf
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os/exec"
+
+	"github.com/threefoldtech/zos/pkg/stubs"
+)
+
+// CPUBenchmarkTask defines CPU benchmark task data.
+type CPUBenchmarkTask struct {
+	taskID   string
+	schedule string
+}
+
+// CPUBenchmarkResult holds CPU benchmark results with the workloads number during the benchmark.
+type CPUBenchmarkResult struct {
+	Single    float64 `json:"single"`
+	Multi     float64 `json:"multi"`
+	Threads   int     `json:"threads"`
+	Workloads int     `json:"workloads"`
+}
+
+var _ Task = (*CPUBenchmarkTask)(nil)
+
+// NewCPUBenchmarkTask returns a new CPU benchmark task.
+func NewCPUBenchmarkTask() CPUBenchmarkTask {
+	return CPUBenchmarkTask{
+		taskID:   "CPUBenchmark",
+		schedule: "0 0 */6 * * *",
+	}
+}
+
+// ID returns task ID.
+func (c *CPUBenchmarkTask) ID() string {
+	return c.taskID
+}
+
+// Cron returns task cron schedule.
+func (c *CPUBenchmarkTask) Cron() string {
+	return c.schedule
+}
+
+// Run executes the CPU benchmark.
+func (c *CPUBenchmarkTask) Run(ctx context.Context) (interface{}, error) {
+	cpubenchOut, err := exec.CommandContext(ctx, "cpubench", "-j").CombinedOutput()
+	if err != nil {
+		return nil, fmt.Errorf("failed to execute cpubench command: %w", err)
+	}
+	cpubenchData := struct {
+		Single  float64 `json:"single"`
+		Multi   float64 `json:"multi"`
+		Threads int     `json:"threads"`
+	}{}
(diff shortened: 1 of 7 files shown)


pkg/perf/cpubench_task.go
```diff
@@ -0,0 +1,75 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+	"os/exec"
+
+	"github.com/threefoldtech/zos/pkg/stubs"
+)
+
+// CPUBenchmarkTask defines CPU benchmark task data.
+type CPUBenchmarkTask struct {
```
xmonader, line 16: please add more explanation on the fields, e.g taskID is that a sequential string?a uuid?or else?
schedule: i'm assuming that's a cron string? let's clear the ambiguity please

pkg/perf/cpubench_task.go
```diff
@@ -0,0 +1,75 @@
...
+	taskID   string
+	schedule string
+}
+
+// CPUBenchmarkResult holds CPU benchmark results with the workloads number during the benchmark.
+type CPUBenchmarkResult struct {
```
xmonader, line 24: Is the Single/Multi process or a thread? can we add a few more info and even change the name to a include the full name (or that's because of the json marshalling?)

pkg/perf/cpubench_task.go
```diff
@@ -0,0 +1,75 @@
...
+	Workloads int     `json:"workloads"`
+}
+
+var _ Task = (*CPUBenchmarkTask)(nil)
+
+// NewCPUBenchmarkTask returns a new CPU benchmark task.
+func NewCPUBenchmarkTask() CPUBenchmarkTask {
+	return CPUBenchmarkTask{
+		taskID:   "CPUBenchmark",
+		schedule: "0 0 */6 * * *",
```
xmonader: can we move that cron string to a const and be used here?

pkg/perf/cpubench_task.go
```diff
@@ -0,0 +1,75 @@
...
+	return c.schedule
+}
+
+// Run executes the CPU benchmark.
+func (c *CPUBenchmarkTask) Run(ctx context.Context) (interface{}, error) {
+	cpubenchOut, err := exec.CommandContext(ctx, "cpubench", "-j").CombinedOutput()
+	if err != nil {
+		return nil, fmt.Errorf("failed to execute cpubench command: %w", err)
+	}
+	cpubenchData := struct {
```
xmonader: Why do we need another anonymous struct while we have CPUBenchmarkResult?

AbdelrahmanElawady: as the JSON output does not have the `workloads` field and I don't know what the behavior of `json` package with missing fields.

xmonader: Hmm I think you can just do `json:"-"` for the field 

https://go.dev/play/p/K6b-FOSs6e3

AbdelrahmanElawady: I can't use that because it will be ignored when querying the task data. I can just remove the anonymous struct and `json` ,I believe, will set it to the default value: 0

pkg/perf/cpubench_task.go
AbdelrahmanElawady: I can change the name of the fields but the data returned from `zos.perf.get_all` will still be single, multi and threads, is that fine?

pkg/perf/cpubench_task.go
```diff
@@ -9,27 +9,32 @@ import (
 	"github.com/threefoldtech/zos/pkg/stubs"
 )
 
+const cpuBenchmarkTaskID = "CPUBenchmark"
+const cpuBenchmarkCronScheule = "0 0 */6 * * *"
```
xmonader: Schedule*

pkg/perf/monitor.go
```diff
@@ -86,3 +94,13 @@ func (pm *PerformanceMonitor) Run(ctx context.Context) error {
...
 	return nil
 }
+
+type zbusClient struct{}
+
+func withZbusClient(ctx context.Context, client zbus.Client) context.Context {
+	return context.WithValue(ctx, zbusClient{}, client)
+}
+
+func getZbusClient(ctx context.Context) zbus.Client {
```
muhamadazmy:
```suggestion
func GetZbusClient(ctx context.Context) zbus.Client {
```

this should be public because external tasks can still use it

muhamadazmy (changes requested):

muhamadazmy (approved):

PR #2067, opened 2023-10-03
adds flag to disable gpu

ashraffouda:


pkg/kernel/kernel.go
```diff
@@ -41,6 +42,11 @@ func (k Params) IsDebug() bool {
 	return k.Exists(Debug)
 }
 
+// GPUDisabled checks if gpu is diabled
```
xmonader, line 45: it says GPUDisabled while the function is called GPUEnabled?

cmds/modules/provisiond/api.go
```diff
@@ -71,6 +73,9 @@ func setupGPURmb(router rmb.Router, store provision.Storage) {
 		return gpus, nil
 	}
 	gpus.WithHandler("list", func(ctx context.Context, payload []byte) (interface{}, error) {
+		if kernel.GetParams().IsGPUDisabled() {
+			return nil, fmt.Errorf("gpu is disabled on this node")
```
muhamadazmy: Why return an error? this can just return an empty list. (no gpus)

muhamadazmy: The client doesn't need to know if the farmer disable the GPU feature, all what he needs to know is that the node has no GPUs for renting

muhamadazmy (changes requested):

muhamadazmy (approved):

Issue #2069, opened 2023-10-03
node validation of farm public ips


muhamadazmy:
We have a big issues with public ips available on the farm is that until this moment there is no way we can validate if those IPs are routable from/to the internet to that farm or not.

A miminaml `<syntax>` validation is done on the IPs to make sure the values provided by the farmer are valid public ips and gateway config is "sane" but we can't grantee if the IPs are topologically valid or are owned by this farm. 

Hence we agreed that a `single` node in each farm (can be the node with lowest ID and online inside the farm) to run a scheduled tests of all "usable" IPs in that farm say every 6 hours.

- Test is run on a single node to avoid conflicts in case multiple nodes decided to test all IPs at the same time (since we probably use internal cron)
- The tests need to only test `free` IPs that are not assigned to any contract to avoid disturbing active workloads
- The test need to be fast enough to not hog the IP for long time, since the IP will still show up as free on the chain and hence can be used.

## The test goes like this:
- On start of the test suite, a `Public` tap will be created by asking networkd to prepare one, this can be used via the stub. The tap name can have a unique hardcoded name like `testtap` or something similar 
- once the tap is created the tap is moved to a special namespace to make it isolated from the rest of the system. This setup can stay as is for the entire life of the system, hence the code need to create it only once and assure it exists only on the service restart.
- by default the tap should not assigned ANY ip (again because IPs can get used) and only IPs are assigned when tests are running.
- when test starts, the node gets the farm object with all its IPs.
- for each IP a test result is generated this can be
  - IP allocated (so we don't run test, but is mentioned in the report)
  - IP free, we run the test, and store the result
- If an IP is free, the IP is assigned to the tab (inside the namespace) and the gw is set correctly (also inside the namespace) then a ping test is done against some known external IPs (guess what, the ping runs inside the namespace)
- The update of the ping test is stored on the report for that IP.

Once all IPs are tested, the current tap ip is remove and the tap is set to `down` but not removed so it's ready to test next time.

This can integrate well in the current `perf` framework that we have, because this expose a way to fetch the tests from the node directly, and integrates well in the node cron.

muhamadazmy:
Actually the `tap` might not work! since it can't be used as a virtual device (unlike veth). A tap is instead need to be used via software but not given IPs on its own (tbh I am not 100% sure) 

[...]

PR #2070, opened 2023-10-03
Add public IPs validation task

AbdelrahmanElawady:
## Description

Add public IPs validation task while ensuring only runs on one node from the farm.

## Steps

- On node startup create a test namespace and macvlan once for the task to use. Do not create them if they already exist.
- Decide if the node should run the task or another one in the farm based on the node ID. The node with the least ID and with power target as up should run it. The other will log why they shouldn't run the task and exit with no errors.
- Get public IPs set on the farm.
- Inside the test namespace and using the macvlan link, assign each IP and its routes to the link and check reliable source for the real public IP. If they match then the public IP is valid.
- Delete all IPs and routes between each IP validation and before running the task to ensure they are available to be used and no problems happen between task runs.

## Issues

#2069


xmonader:
Blocked on docs

pkg/network/public/public.go
```diff
@@ -319,6 +321,11 @@ func EnsurePublicSetup(nodeID pkg.Identifier, inf *pkg.PublicConfig) (*netlink.B
 		return nil, errors.Wrap(err, "failed to get current public bridge uplink")
 	}
 
+	err = ensureTestNamespace(br)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create test namespace: %w", err)
+	}
+
```
muhamadazmy:
```suggestion
	if err := ensureTestNamespace(br); err != nil {
		return nil, errors.Wrap(err, "failed to create test namespace")
	}

```

pkg/network/public/public.go
```diff
@@ -349,6 +356,15 @@ func EnsurePublicSetup(nodeID pkg.Identifier, inf *pkg.PublicConfig) (*netlink.B
 	return br, netlink.LinkSetUp(br)
 }
 
+func ensureTestNamespace(publicBrdige *netlink.Bridge) error {
```
muhamadazmy, line 391: I don't see where u do `ensure` here. You just blindly create the namespace and the macvlan. This code will fail when it tries to run again or breaks the setup. 

The point of an ensure function is to validate and build the setup if not there. Hence you need to be check (validate) how good the setup is, and then create, or change it to match the requires setup.

This function can be called multiple times during the runtime for example because of a system update

pkg/perf/pubip_task.go
```diff
@@ -0,0 +1,238 @@
...
+				continue
+			}
+			unusedIPs[publicIP.IP] = false
+
+			ip, ipNet, routes, err := getIPWithRoute(publicIP)
+			if err != nil {
+				log.Err(err).Send()
+				continue
+			}
+			err = macvlan.Install(mv, nil, ipNet, routes, netNS)
```
muhamadazmy: I don't believe you need to pass netNS because u are already inside the namespace. it's then better to pass `nil` to avoid unpredeceted errors

pkg/perf/pubip_task.go
```diff
@@ -0,0 +1,238 @@
...
+		return nil, nil
+	}
+
+	farm, err := sub.GetFarm(uint32(farmID))
+	if err != nil {
+		return nil, fmt.Errorf("failed to get farm with id %d: %w", farmID, err)
+	}
+	unusedIPs := make(map[string]bool)
+	err = netNS.Do(func(nn ns.NetNS) error {
+		mv, err := macvlan.GetByName(testMacvlan)
```
muhamadazmy: This is too much code in the Do closure, i prefer u create another function that has the logic then do `netNS.Do(p.logic)` this way the indentation of the code won't also be too much

pkg/perf/pubip_task.go
```diff
@@ -0,0 +1,238 @@
...
+	registrar := stubs.NewRegistrarStub(cl)
+	var nodeID uint32
+	for {
+		nodeID, err = registrar.NodeID(ctx)
+		if err == nil {
+			break
+		}
+		log.Err(err).Msg("failed to get node id")
+		time.Sleep(10 * time.Second)
+	}
```
muhamadazmy: There is a package `backoff` that we use you should use here instead of this.

muhamadazmy (changes requested):

pkg/perf/pubip_task.go
```diff
@@ -0,0 +1,238 @@
...
+	defer req.Body.Close()
+
+	body, err := io.ReadAll(req.Body)
+	if err != nil {
+		return "", err
+	}
+	return string(body), nil
+}
+
+func deleteIPAndRoutes(publicIP substrate.PublicIP, routes []*netlink.Route, macvlan netlink.Link) error {
```
muhamadazmy: I really think you should just delee all Ips and routes blindly, so u start by listing the ips on this link and ALL routes inside the namespace and delete them.

This function should also be called before the test to make sure there are no left over ips on the device because of a crash

AbdelrahmanElawady:
Also, I think the directory of the tasks should not be called `perf` as it is concerned with running tasks and monitoring them. Maybe `tasks` is better? or moving tasks not related to "performance" to another directory?

cmds/modules/noded/main.go
```diff
@@ -208,6 +208,9 @@ func action(cli *cli.Context) error {
 	cpuBenchmarkTask := perf.NewCPUBenchmarkTask()
 	perfMon.AddTask(&cpuBenchmarkTask)
 
+	pubIPTask := perf.NewPublicIPValidationTask()
+	perfMon.AddTask(pubIPTask)
```
muhamadazmy: Why do u need the intermediate variable `pubIPTask` ?
```suggestion
	perfMon.AddTask(perf.NewPublicIPValidationTask())
```

pkg/network/public/public.go
```diff
@@ -26,6 +26,8 @@ import (
 const (
 	toZosVeth                   = "tozos" // veth pair from br-pub to zos
 	publicNsMACDerivationSuffix = "-public"
+	testMacvlan                 = "pubtestmacvlan"
```
muhamadazmy: weird name, this can just be `pub` or something. it's inside a namespace anyway so no chances of name collision.

pkg/network/public/public.go
```diff
@@ -349,6 +355,24 @@ func EnsurePublicSetup(nodeID pkg.Identifier, inf *pkg.PublicConfig) (*netlink.B
 	return br, netlink.LinkSetUp(br)
 }
 
+func ensureTestNamespace(publicBrdige *netlink.Bridge) error {
+	netNS, err := namespace.GetByName(testNamespace)
+	if err != nil {
```
muhamadazmy: if you check `GetByName` implementation you will see that it returns os.ErrNotExist if the namespace does not exist. So that's the only valid case you should try to create the namespace, then any other error type should be returned

```go
netNS, err := GetByName(name)
if os.IsNotExist(err) {
   err = Create(name)
}

if err != nil {
   return err
}
```

pkg/perf/pubip_task.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+func getRealPublicIP() (string, error) {
+	// for testing now, should change to cloudflare
+	req, err := http.Get("https://api.ipify.org/")
+	if err != nil {
+		return "", err
+	}
+	defer req.Body.Close()
+
```
muhamadazmy: You need to check the returned status from the req in case it's not `200`

muhamadazmy (changes requested):
Please also create a sub pkg under perf to include all your files related to this task for organization purposes

pkg/perf/publicip/publicip_task.go
```diff
@@ -0,0 +1,322 @@
...
+
+		realIP, err := getRealPublicIP()
+		if err != nil {
+			p.farmIPsReport[publicIP.IP] = IPReport{
+				State:  SkippedState,
+				Reason: FetchRealIPFailed,
+			}
+			log.Err(err).Msg("failed to get node real IP")
+			continue
+		}
```
muhamadazmy: This is not always correct, if you failed to get the real ip this can duo to different reasons. and they all need to be handled differently:
- http.Get fail, this can be duo to name look failure, or failed to establish connection this is hugely mean that the IP is invalid (is not routable to the internet) 
- http.Get succeeded but the response is not with success status (!= 200) this means that u have internet but the check service is donw, in that case we are not sure about the IP, we can mark it as skipped, for now.
- http.Get succeed but the the returned value does not match the actual IP we are testing means those IPs are natted, so they are also invalid.

the best way to do this is that getRealPublicIP need to return (wrapped) errors with more details on what happened exactly, and then based on that choose the proper state

pkg/network/public/public.go
```diff
@@ -382,6 +388,27 @@ func EnsurePublicSetup(nodeID pkg.Identifier, vlan *uint16, inf *pkg.PublicConfi
...
 
+func ensureTestNamespace(publicBrdige *netlink.Bridge) error {
+	netNS, err := namespace.GetByName(testNamespace)
+	if errors.Is(err, os.ErrNotExist) {
+		netNS, err = namespace.Create(testNamespace)
+		if err != nil {
+			return fmt.Errorf("failed to create namespace %s: %w", testNamespace, err)
+		}
+	}
+	if err != nil {
```
muhamadazmy: this is probably what u want. since u already check the error of the create

```suggestion
	} else if err != nil {
```

muhamadazmy (changes requested):

muhamadazmy (approved):

Issue #2075, opened 2023-10-11
update the deprecated set-output


xmonader:
https://dev.to/cicirello/how-to-patch-the-deprecated-set-output-in-github-workflows-and-in-container-actions-9co#updated-workflow-step-outputs

muhamadazmy:
I actually tried this when they first deprecated it and it did not work! (the CI failed) so i reverted it back until github actually makes it work

PR #2078, opened 2023-10-15
Fix deprecated set output

Eslam-Nawara:
### Description

replace deprecated `set-output` with environment variable `$GITHUB_OUTPUT` 

### Related issues
- #2075

### DIFF
--- a/bins/bins-extra.sh
+++ b/bins/bins-extra.sh
@@ -269,7 +269,7 @@ exclude_libs() {
 
 github_name() {
     # force github print
-    echo "::set-output name=name::${1}"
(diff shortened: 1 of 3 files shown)


bins/bins-extra.sh
```diff
@@ -269,7 +269,7 @@ exclude_libs() {
 
 github_name() {
     # force github print
-    echo "::set-output name=name::${1}"
+    echo "name=${1}" >> $env:GITHUB_OUTPUT
```
muhamadazmy, line 272: Why this is `$env:GITHUB_OUTPUT` and not `$GITHUB_OUTPUT` ?

muhamadazmy (approved):

Issue #2071, opened 2023-10-04
install and support rfs v2 with backward compatibility support


muhamadazmy:
Right now zos uses rfs with v1 (named g8ufs on the zos) we need to also have (and support) v2 on the node. So we ALWAYS need to maintain the 2 binary versions 
- v1 (already named g8ufs)
- v2 (we can stick with rfs name)

The flistd daemon need to be able to decide which one of the 2 binaries to use based on the `flist` name. Right now the best way to do so is by just checking the `flist` extension of the flist to be mounted.
- v1 flist extension is always (.flist) as done by the hub.
- v2 flist is a completely new format (named fungistore) hence the extension used is `.fl` 

flistd then can detect the type, and based on that either use the old binary, or the new one. Note that they both has different argument lists.

khaledyoussef24:
where can i verify or check the images of v2 `.fl` ?

khaledyoussef24:
- verified 
- followed and read the docs of https://github.com/threefoldtech/rfs

- it was clear and followed them and `ran rfs pack -m output.fl -s 00-80=dir:///tmp/store0 -s 81-ff=dir:///tmp/store1 ~/Documents`
![image](https://github.com/threefoldtech/zos/assets/110984055/b93d8f8f-faa9-490b-9acd-cf9ed1a08f09)

- and the output file created successfully 
![image](https://github.com/threefoldtech/zos/assets/110984055/2b215483-9ecf-43e0-b2a3-08d229f2b435)

- then mounting the `.fl` file to another directory
![image](https://github.com/threefoldtech/zos/assets/110984055/022c6919-5181-4985-b342-fcf6a0239085)

- then checking the directory from another terminal to check the mounted docs
![image](https://github.com/threefoldtech/zos/assets/110984055/9170914e-265c-40b8-878f-b0e060aa4a4e)


- the mounted docs copied successfully 
![image](https://github.com/threefoldtech/zos/assets/110984055/6f7127c2-546c-460c-a553-73bf8de06e47)


- the data found successfully

khaledyoussef24:
verified deployed a vm using terraform using v2 '.fl'
and it worked fine and was reachable
it is not supported on playground yet
the example used:
![image](https://github.com/threefoldtech/zos/assets/110984055/8462ab28-caae-45f7-9c47-fb7d45088160)
deployed 
![image](https://github.com/threefoldtech/zos/assets/110984055/342c8bd4-6145-427f-9fe7-f331096550aa)
[...]

PR #2079, opened 2023-10-15
Main rfs

rawdaGastan:
### Issues

- https://github.com/threefoldtech/zos/issues/2071

### DIFF
--- a/pkg/flist/flist.go
+++ b/pkg/flist/flist.go
@@ -207,12 +207,19 @@ func (f *flistModule) mountRO(url, storage string) (string, error) {
 	}
 
 	logPath := filepath.Join(f.log, hash) + ".log"
+
+	flistExt := filepath.Ext(flistPath)
 	var args []string
+
+	if flistExt == ".flist" {
+		args = append(args,
+			"--storage-url", storage,
+		)
+	}
 
 	args = append(args,
 		"--cache", f.cache,
 		"--meta", flistPath,
-		"--storage-url", storage,
 		"--daemon",
 		"--log", logPath,
 		// this is always read-only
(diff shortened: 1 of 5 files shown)


pkg/flist/mounts.go
```diff
@@ -12,6 +12,7 @@ import (
 )
 
 const (
+	fsTypeRfs     = "fuse.rfs"
```
muhamadazmy: this is not correct, this require a change to `rfs` actually because it still identity itself as `g8ufs` (that's something that is hard coded in the code) of course we can change that in rfs itself but I don't see why would we do that right now. 

the idea was to be able to tell `rfs` from overlay so as long as we can do that it's fine. so in other words there should be no difference in reading the `rfs` from `g8ufs` since they both use the pid as source mount

pkg/flist/mounts.go
```diff
@@ -116,7 +117,7 @@ func (f *flistModule) resolve(path string) (g8ufsInfo, error) {
 		return g8ufsInfo{}, err
 	}
 
-	if info.FSType == fsTypeG8ufs {
+	if info.FSType == fsTypeG8ufs || info.FSType == fsTypeRfs {
```
muhamadazmy: this means this is not deeded since both are defined as fuse.g8ufs

muhamadazmy (changes requested):

muhamadazmy:
For testing the idea is that if u use `zdb` as a store the IP used while creating the flist must be reachable from ZOS. so if u are running zdb locally make sure:
- zdb is listening on all IPs.
- while creating the flist, u have to use the `ip` of the machine not `localhost` or `127.0.0.1`

pkg/flist/flist.go
```diff
@@ -207,25 +207,33 @@ func (f *flistModule) mountRO(url, storage string) (string, error) {
...
+			// this is always read-only
+			"--ro",
+			mountpoint,
+		)
+		cmd = f.commander.Command("g8ufs", args...)
+	} else if flistExt == ".fl" {
+		sublog.Info().Strs("args", args).Msg("starting rfs daemon")
+		args = append([]string{"mount"}, append(args, mountpoint)...)
+		cmd = f.commander.Command("rfs", args...)
+	}
```
muhamadazmy, line 234: can we add an `else` case here so we return an error if the url extension is unknown

muhamadazmy (approved):

PR #2082, opened 2023-10-17
Rework release

muhamadazmy:
This is a big change to how we build and release ZOS. The idea is that we create a "consistent" and "reproducible" builds and release. Before this, zos building creates multiple scattered flists, releases has to have specific suffix so the build system knows where to place the files from a certain build. 

This made it hard to move a qa release to testnet without creating a new release on the same head. This can also cause errors during the update. 

This new mechanism uses a new hub feature that was developed specifically for this story. On the hub, all packages related to a build will get tagged with the same release or sha. Later on, a link from the network name will be created to that specific tag. So all packages that belong to a specific tag will be installed together. That simple

this required the following changes:
- all github actions needed to adapt to the new feature, tagging flists on building 
- under tf-autobuild a tag will appear (looks like a directory) that lists all links to tagged packages. the name of that tag can be a short sha of the commit (from main) or a decent version say `v1.2.3` 
- under tf-zos a link to the specific tag associated with that network is created 
  -  for example `tf-zos/production -> tf-autobuild/v1.2.3`
- ZOS required other changes to make this work
  - bootstrap binary need to understand new system, it also needed to be backward compatible (if link from tf-zos does not exist) it needs to fallback to old bootstrap style.
  - identityd needs to switch to use the new update system

Things to consider during the switch:
- When creating a release for "A" network the current nodes in that network still uses the old style, hence we need to first create a "link" on the hub to the new zos build, then the system will have to do 2 updates, the first to bring latest binary then the second to actually update with new system


.github/workflows/publish.yaml
```diff
@@ -0,0 +1,107 @@
+# the publish workflow builds zos binaries
+# and tag them correctly
+# the build is triggered when pushing to any
+# branch, but the tagging of the build only happens
+# either on main or when a tag is created
+name: Release
+on:
+  push:
+    branches:
+      - "*"
```
AbdelrahmanElawady, line 10: I am curious on why we are publishing on every push on any branch. I understand for tags and maybe on main, but why all branches?

bootstrap/bootstrap/src/config.rs
```diff
@@ -49,17 +63,16 @@ fn version() -> Result<Version> {
...
                     bail!("unknown version: {}", m);
                 }
             },
-            None => Version::V2,
+            None => Version::V3,
         },
         // version was not provided in cmdline
         // so default is v2
-        None => Version::V2,
+        None => Version::V3,
```
AbdelrahmanElawady: the comment should be V3?

bootstrap/bootstrap/src/hub.rs
```diff
@@ -34,53 +43,101 @@ impl Repo {
...
         };
-        info.url = format!("{}/{}/{}", self.base, self.name, flist.as_ref());
+        info.url = format!("{}/{}/{}", HUB, self.name, flist.as_ref());
         Ok(info)
     }
 }
 
 impl Flist {
+    // tag_link return the target repo and tag name for
+    // a taglink flist
```
AbdelrahmanElawady: if this is intended to be docstring should it be `///`?

cmds/identityd/main.go
```diff
@@ -165,176 +137,14 @@ func main() {
 		log.Info().Msg("received a termination signal")
 	})
 
-	if bootMethod != upgrade.BootMethodFList {
-		log.Info().Msg("node is not booted from an flist. upgrade is not supported")
-		<-ctx.Done()
+	err = upgrader.Run(ctx)
+	if errors.Is(err, upgrade.ErrRestartNeeded) {
```
AbdelrahmanElawady, line 141: I am not familiar with how the updates happen but how exactly does exiting begin the needed restart? or is it just restarting this service?

pkg/upgrade/hub/hub.go
```diff
@@ -0,0 +1,347 @@
...
+
+	defer response.Body.Close()
+	defer func() {
+		_, _ = io.ReadAll(response.Body)
+	}()
+
+	if response.StatusCode != http.StatusOK {
+		return info, fmt.Errorf("failed to get flist (%s/%s) info: %s", repo, name, response.Status)
+	}
+	defer response.Body.Close()
```
AbdelrahmanElawady: isn't the second defer here will read from a closed body?

.github/workflows/publish.yaml
muhamadazmy: This indeed will trigger the build for all branches, but later in the workflow we have checks for specific patterns to make sure tagging of flists only happens on main and a tagged branch. 

This is to make sure all builds are working fine on that branch. We can later improve on this and only build on "development" branches if the workflows or the binaries has modified.

cmds/identityd/main.go
muhamadazmy: That restarts the service itself only, not the system. It's needed to force the service to run new code. This allows code to ask the service to restart without being "seen" as an error

pkg/upgrade/hub/hub.go
muhamadazmy: Yeah, you are correct. I didn't see that one. good catch

bootstrap/bootstrap/src/bootstrap.rs
```diff
@@ -57,26 +59,15 @@ fn boostrap_zos(cfg: &config::Config) -> Result<()> {
...
-    })?;
-
-    result
-}
-
 /// update stage make sure we are running latest
 /// version of bootstrap
 pub fn update(cfg: &config::Config) -> Result<()> {
-    let result = WorkDir::run(WORKDIR, || -> Result<()> {
+    let result = WorkDir::run(WORKDIR, || {
```
Omarabdul3ziz, line 71: I don't fully understand how errors are handled in Rust, but inferring the return type for closure will allow to return anyhow propagated errors coming from `update_bootstrap()?`
Is it necessary or will this happen by default?

qemu/overlay
```diff
@@ -0,0 +1 @@
+overlay.normal
```
Omarabdul3ziz, line 1: why change the directory to `overlay.normal` and link to it with overlay?

Omarabdul3ziz (commented):
I just have two parts I don't fully understand

bootstrap/bootstrap/src/bootstrap.rs
```diff
@@ -57,26 +59,15 @@ fn boostrap_zos(cfg: &config::Config) -> Result<()> {
...
         update_bootstrap(cfg.debug)?;
```
mariobassem: debug is always false because `update()` caller will return if debug is true, no?
hence, name will always be "bootstrap:latest.flist", right?
https://github.com/threefoldtech/zos/blob/124e113e40e09c00c9fa9aca9e2fc9816f837810/bootstrap/bootstrap/src/bootstrap.rs#L90-L97

bootstrap/bootstrap/src/bootstrap.rs
muhamadazmy: If you check the `run` function signature you will see that it takes a closure (FnOnce) that return Result<T>. Where T is a generic type.

The generic is usually inferred from the code, if the compiler fail to infer it, you will have to define it yourself (like bfeore)

so if I passed a clousre `|| { 10  }` the compiler will assume the return type is a `usize` and then the return type is Result<usize>. 

The code has enough clause so the compiler knows what I return, hence I don't have to define it as before. which is `Result<(), anyhow::Error>`

qemu/overlay
muhamadazmy: During the development i usually create multiple copies of `overlay` that has different binaries and setups. I normally never push those other `overlay` directories. but over time it became a hassle to keep renaming the directories and know which ones has what.

I eventually decided to just create a symlink `overly` and suffix each directory with a name. 
```sh
overlay.normal # probably should have named it `overlay.default`
overlay.bootstrap 
#etc...
overlay -> overlay.normal
```

this way i can switch by just changing the symlink to which overlay i want to use

bootstrap/bootstrap/src/bootstrap.rs
muhamadazmy: Actually very good catch. that's absolutely correct. I think there were code before that made sense of this but is gone now. 

I think the best way is remove the debug flag. in the update function.

muhamadazmy: Anyway, the logic was that if starting bootstrap in debugging avoid updating yourself usually to test local code so it skips this stage. 
Means we never need to use the development build

bootstrap/bootstrap/src/hub.rs
```diff
@@ -34,53 +43,101 @@ impl Repo {
...
             s => bail!("failed to get flist info: {}", s),
         };
-        info.url = format!("{}/{}/{}", self.base, self.name, flist.as_ref());
+        info.url = format!("{}/{}/{}", HUB, self.name, flist.as_ref());
         Ok(info)
     }
 }
 
 impl Flist {
+    /// tag_link return the target repo and tag name for
```
mariobassem, line 117: should we add in the doc that this method panics if the kind is not `TagLink`?

mariobassem: or should we not panic? :D

muhamadazmy: I will add that to the docs that it will panic. The panic is okay in this case because this means u have a development error. (there is no way u can call this with wrong type unless u have wrong logic) hence the panic

PR #2084, opened 2023-10-18
Vlan support

muhamadazmy:
Add support for new boot params 
- `vlan:priv` (optional) sets the private subnet vlan
- `vlan:pub` (optional) sets the public subnet vlan (only in single node setup)

Please read docs for more details.


pkg/kernel/kernel.go
```diff
@@ -38,6 +38,19 @@ func (k Params) Get(key string) ([]string, bool) {
...
+	all, found := k.Get(key)
+	if !found {
+		return "", false
+	}
+
+	if len(all) == 0 {
+		return "", false
+	}
+
+	return all[len(all)-1], true
```
xmonader, line 54: does this mean it returns the last item in case of the item passed multiple times? if so, let's please document the behavior

xmonader (dismissed):

pkg/kernel/kernel.go
muhamadazmy: yes, if the same param is provided multiple times via a cmdline it's overridden by the last value provided. This way u can override hardcoded params in the image with whatever values u provide during booting

PR #2089, opened 2023-10-23
Add iperf output types and use it instead of go-iperf

AbdelrahmanElawady:
## Description

`go-iperf` didn't have proper cleanup for temp directories created. So, it was easier to just exec the command and parse it directly instead of using it.

## Changes

- Add iperf output types
- Run iperf directly during the task
- Remove go-iperf

### DIFF
--- a/pkg/perf/iperf_task.go
+++ b/pkg/perf/iperf_task.go
@@ -2,10 +2,11 @@ package perf
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
 	"net"
 	"os/exec"
 
-	goIperf "github.com/BGrewell/go-iperf"
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog/log"
 	"github.com/threefoldtech/zos/pkg/environment"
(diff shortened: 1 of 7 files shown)


pkg/perf/iperf_task.go
```diff
@@ -104,35 +105,46 @@ func (t *IperfTest) Run(ctx context.Context) (interface{}, error) {
...
-		iperfClient.SetProto(goIperf.PROTO_UDP)
+		opts = append(opts, "--length", "16B", "--udp")
 	}
-
-	err := iperfClient.Start()
-	if err != nil {
-		log.Error().Err(err).Msgf("failed to start iperf client with ip '%s'", clientIP)
+	output, err := exec.CommandContext(ctx, "iperf", opts...).CombinedOutput()
+	exitErr := &exec.ExitError{}
+	if err != nil && !errors.As(err, &exitErr) {
```
xmonader, line 131: Actually I'm not sure I follow, we ignore errors other than exitErr? while that actually means we failed to execute the binary?

pkg/perf/iperf_types.go
```diff
@@ -0,0 +1,134 @@
+package perf
```
xmonader: What is the point of having these here while they exist in go-iperf code?

AbdelrahmanElawady: because go-iperf creates a temp directory in `init` so by importing go-iperf it would cause the same problem we are avoiding each time the task runs

pkg/perf/iperf_task.go
AbdelrahmanElawady: we ignore exit errors because `iperf` exits with a non-zero code in case it failed to run and prints the error reason in `error` field in JSON output. For example:

```json
{
  "start": {
    "connected": [],
    "version": "iperf 3.14",
    "system_info": "Linux zero-os 5.10.55-Zero-OS #2 SMP PREEMPT Wed Aug 9 13:31:25 UTC 2023 x86_64"
  },
  "intervals": [],
  "end": {},
  "error": "unable to connect to server - server may have stopped running or use a different port, firewall issue, etc.: Network unreachable"
}
```

muhamadazmy (changes requested):
In general this looks good, but my only comment is that the `perf` package is getting very cluttered. Can u move the iperf task into a sub package
```
perf
  /iperf/
```
since the registration of the tasks happens in noded/main this should not cause any cyclic dependency

> NOTE: please also move other tasks to their own separate sub packages to perf.
so we will also have `cpu` and eventually public-ips

muhamadazmy (approved):

Issue #2081, opened 2023-10-16
unify the bin package (rfs and zufs)  


muhamadazmy:
since both binaries are installed from the same repo, please merge the 2 into one package using 
```
RFS_VERSION_V1="<v1 versino>"
RFS_VERSION_V2="<v2 versino>"
```
and so on

PR #2091, opened 2023-10-24
unify rfs and zufs pkgs

Eslam-Nawara:
## Related Issues:
- #2081


bins/packages/rfs/rfs.sh
```diff
@@ -1,22 +1,31 @@
...
-    download_file ${RFS_LINK} ${RFS_CHECKSUM} rfs-${RFS_VERSION}
+    download_file ${RFS_LINK_V1} ${RFS_CHECKSUM_V1} rfs-${RFS_VERSION_V1}
+    download_file ${RFS_LINK_V2} ${RFS_CHECKSUM_V2} rfs-${RFS_VERSION_V2}
 }
 
 prepare_rfs() {
     echo "[+] prepare rfs"
-    github_name "rfs-${RFS_VERSION}"
+    github_name "rfs-${RFS_VERSION_V1}"
+    github_name "0-fs-${RFS_VERSION_V2}"
```
muhamadazmy: You can only produce one github_name. That will be `rfs-${RFS_VERSION_V2}`

basically that is gonna be the name and version of the produced pacakge.
this one will have then 2 binaries, rfs and g8ufs

muhamadazmy (changes requested):

bins/packages/rfs/rfs.sh
```diff
@@ -1,22 +1,31 @@
-RFS_VERSION="2.0.0"
-RFS_CHECKSUM="d6e1c3036d8b0d7d6df4d67fe17a0502"
-RFS_LINK="https://github.com/threefoldtech/rfs/releases/download/v${RFS_VERSION}/rfs"
+RFS_VERSION_V1="1.1.1"
+RFS_VERSION_V2="2.0.0"
```
muhamadazmy: Since we are at it, can we change this to latest "2.0.0" .Note that changing this will also require changing the RFS_CHECKSUM_V2 hash. You can simply do this by downloading the binary from the release 2.0.1 and do an `md5sum rfs` and then use that value.

muhamadazmy (changes requested):
Much better, but we can also maybe update v2 to latest 2.0.1. check the comment

muhamadazmy (approved):

PR #2094, opened 2023-10-27
RFS new format integration. Avoid hub specific calls

muhamadazmy:
- Also update rfs to latest version to support s3 store
- Run rfs inside public namespace, or ndmz if public is not setup

This way the filesystem will use the public network (which is mostly faster) and will be able to use yggdrasil as well

### DIFF
--- a/pkg/flist/flist.go
+++ b/pkg/flist/flist.go
@@ -42,7 +42,7 @@ var (
 	ErrTransportEndpointIsNotConencted = errors.New("transport endpoint is not connected")
 	ErrZFSProcessNotFound              = errors.New("0-fs process not found")
 	ErrHashNotSupported                = errors.New("hash not supported by flist host")
-	ErrHashInvalid                     = errors.New("invalid hash length")
+	ErrHashInvalidLen                  = errors.New("invalid hash length")
 )
 
 // Hash type
@@ -623,7 +623,7 @@ func (f *flistModule) FlistHash(url string) (string, error) {
 
 	hashStr := strings.TrimSpace(string(hash))
 	if len(hashStr) != md5HexLength {
-		return "", ErrHashInvalid
+		return "", ErrHashInvalidLen
 	}
 
 	return hashStr, nil
--- a/pkg/flist/flist.go
+++ b/pkg/flist/flist.go
@@ -15,17 +15,23 @@ import (
 	"syscall"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/flist/flist.go
```diff
@@ -35,10 +41,19 @@ var (
 	ErrNotMountPoint                   = errors.New("path is not a mountpoint")
 	ErrTransportEndpointIsNotConencted = errors.New("transport endpoint is not connected")
 	ErrZFSProcessNotFound              = errors.New("0-fs process not found")
+	ErrHashNotSupported                = errors.New("hash not supported by flist host")
+	ErrHashInvalid                     = errors.New("invalid hash length")
 )
 
+// Hash type
+type Hash string
```
xmonader, line 49: do we need Hash and Path types? it seems they're only converted from a string and to string in the code

pkg/flist/flist.go
```diff
@@ -35,10 +41,19 @@ var (
 	ErrNotMountPoint                   = errors.New("path is not a mountpoint")
 	ErrTransportEndpointIsNotConencted = errors.New("transport endpoint is not connected")
 	ErrZFSProcessNotFound              = errors.New("0-fs process not found")
+	ErrHashNotSupported                = errors.New("hash not supported by flist host")
+	ErrHashInvalid                     = errors.New("invalid hash length")
```
xmonader: I suggest to change it to `ErrInvalidHashLen`, as it could be eg Malformed hash or something else

pkg/flist/flist.go
muhamadazmy: The type alias just to make it clear when u reading the code and also give hints during compile time so u can't mix them by mistake. I did this because the method returns path and hash, if i use string `return (string, string, error)` it will never be clear which is which until u read docs or the code each time. But if i return (Hash, Path, error) it's clear and can't be mixed by accident, and only convert them back to string when needed

pkg/flist/flist.go
muhamadazmy: Yeah, will do.

pkg/flist/flist.go
xmonader: You're right, I was thinking of changing the function signature to use named returns so would be `hash, path string, err error`

xmonader (dismissed):

xmonader (approved):

PR #2100, opened 2023-10-31
Fix setting the gateway and caching the request

AbdelrahmanElawady:
## Description

There was an issue with the way the routes was set using the gateway which set the valid public IPs as invalid. Also, http.Get caches the request (or uses the host ns) and actually makes the request when it shouldn't even reach the validation service. So, changed that too.


pkg/perf/publicip/publicip_task.go
```diff
@@ -163,11 +163,13 @@ func (p *publicIPValidationTask) validateIPs(publicIPs []substrate.PublicIP) err
 				State:  InvalidState,
 				Reason: PublicIPDataInvalid,
 			}
+			log.Err(err).Send()
```
muhamadazmy: What does this do?

muhamadazmy (approved):

Issue #1974, opened 2023-06-02
Discussion: improvements to VM detection


scottyeager:
I have seen a couple cases so far where `virt-what` misfires and detects bare metal hardware as a VM. In the first case, I noticed that the behavior had been corrected in a later release of `virt-what` and recommended upgrading what we use in Zos. Unfortunately, the newer version also included a new check that misidentifies another farmer's hardware as a VM.

While I've filed a [bug report](https://bugzilla.redhat.com/show_bug.cgi?id=2211980) upstream to correct the new misfire, and we can compensate the farmer for now from the treasury, this got me thinking about how we can improve VM detection.

Some observations and notes:
* Both bugs in `virt-what` were essentially the same type, using a manufacturer tag in the DMI info as the sole check when the manufacturer does both VM software and hardware. This suggests a pattern, whereby new checks are introduced that are initially too broad and then later refined after the issue is noticed
* `virt-what` is therefore maintained in such a way that doesn't emphasize rigor in the checks before release (it's focus seems to be providing a convenient way to check for many possible VM types, rather than a high degree of correctness)
* As far as [minting is concerned](https://github.com/threefoldtech/minting_v3/blob/4fa0c51b03a391478f0e6bbda4c8f622efeb24bf/minting/src/main.rs#L361), "once a VM, always a VM" (this is simple to change, but for now it's a fact)

Some potential approaches:
* Try to anticipate which updates to `virt-what` might cause new misfires (introducing new checks rather than refining existing ones)
* Don't change a node's type from bare metal to VM unless some other change happens too (DMI changes, hardware stats change, etc). Allow a change from VM to bare metal though, in case some misfire is fixed. Some undetected VMs might continue farming after they would otherwise be detected, but at least we don't have real bare metal nodes suddenly tagged as VM and not minting

xmonader:
wasn't virt-what updated recently ? https://github.com/threefoldtech/zos/blob/baf8d9ee77fe9d20bd3ab2f83c73c5e133ff462d/bins/packages/virtwhat/virtwhat.sh#L26

Eslam-Nawara:
### Work Completed:
- [x] make sure virt-what is up-to-date
- [x] prevent node type update to vm if the node was registered as hw

PR #2105, opened 2023-11-06
improve VM detection

Eslam-Nawara:
## Related Issues

- #1974


muhamadazmy (changes requested):
please forma the code properly. and make the formatter part of the editor save routine

muhamadazmy (approved):

Issue #2115, opened 2023-11-15
inconsistency in the RMB calls for the new added test calls


AhmedHanafy725:
- some use camel case and others use snake case (used to get snake case from zos calls)

```
{ name: 'PublicIPValidation' },
{ name: 'CPUBenchmark' },
{ name: 'iperf' }
```

- to get one of the tests only, we can use `zos.perf.get` with the test name as the payload (payload: `'"iperf"'`), I think it would look nicer if the payload becomes `{"name": "iperf"}` and to be consistent with the rest zos calls

muhamadazmy:
I suggest we use `Kebab case` instead of camel case it looks better in `json` keys
- `public-ip-validation` 
- `cpu-benchmark` 
- `iperf`

Also can you add a `Description` method to the task interface that has description of the test, and explain the results. Means that all tasks need to return a meaningful description

khaledyoussef24:
verified by running the rmb call get.perf.all
names changed from
```

{ name: 'PublicIPValidation' },
{ name: 'CPUBenchmark' },
{ name: 'iperf' }

```

to

```
    "name": "public-ip-validation",
    "name": "iperf",
    "name": "cpu-benchmark",
```

description is also added to the response of the rmb call

Issue #2114, opened 2023-11-15
Can't get the result of the public ip validation


AhmedHanafy725:
Env: devnet
nodeId: 12

### Steps

- Get public ip validation test result through RMB call `zos.perf.get_all`.

### Result 

no result was found on node 12 (the node with the smallest ID and up state) as node 11 is down (the last `updateAt` for node 11 `Tuesday, November 14, 2023 12:54:12 PM GMT`)

```ts
{ name: 'PublicIPValidation', timestamp: 1700028002, result: null },
```

### Expectations

as node 11 has been down since yesterday, I was expecting to see the public ip validation result on node 12.

muhamadazmy:
![image](https://github.com/threefoldtech/zos/assets/10920323/f3a6ea68-c796-4de7-974e-04b710185919)

Seems that node 12 still thinks that there is another node that should do the tests. Maybe it doesn't take into account if the node is not up anymore.

@AbdelrahmanElawady can you please take a look why that is

muhamadazmy:
https://github.com/threefoldtech/zos/pull/2120

muhamadazmy:
@Omarabdul3ziz please update this issue.

Omarabdul3ziz:
updated the node status decision part, to use graphql client instead of directly calling the chain.

khaledyoussef24:
issue is not producible any more 

```
 "description": "Runs on the least NodeID node in a farm to validate all its IPs.",
    "name": "public-ip-validation",
    "result": {
      "185.206.122.16/24": {
        "reason": "ip is already assigned to a contract",
        "state": "skipped"
      },
      "185.206.122.17/24": {
        "reason": "ip is already assigned to a contract",
        "state": "skipped"
      },
      "185.206.122.18/24": {
        "reason": "ip is already assigned to a contract",
        "state": "skipped"
      },
      "185.206.122.19/24": {
        "reason": "ip is already assigned to a contract",
        "state": "skipped"
      },
      "185.206.122.20/24": {
        "reason": "ip is already assigned to a contract",
        "state": "skipped"
      },
      "185.206.122.21/24": {
        "reason": "ip is already assigned to a contract",
        "state": "skipped"
      },
      "185.206.122.22/24": {
        "reason": "ip is already assigned to a contract",
        "state": "skipped"
      },
      "185.206.122.23/24": {
        "reason": "ip is already assigned to a contract",
        "state": "skipped"
      },
      "185.206.122.24/24": {
[...]

Issue #2112, opened 2023-11-15
CPU benckmark result doesn't tell if is it good or bad


AhmedHanafy725:
Env: devnet

### Steps

- Get CPU benchmark test result through RMB call `zos.perf.get_all`.

### Result

The result says some numbers but what do these numbers represent? Or what are the good ranges?

```
{
  name: 'CPUBenchmark',
  timestamp: 1700028004,
  result: { multi: 59.723, single: 2.06, threads: 24, workloads: 0 }
}
```

muhamadazmy:
I suggest that a test Task should also have a "description" that explain what do the values mean

Omarabdul3ziz:
basically, the test runs the binary from this [repo](https://github.com/threefoldtech/cpu-benchmark-simple) and as I understand from the code there, it just runs a CRC64 computing process and calculates the timespent. so these numbers represent the time in seconds the less time it takes the better.

verbose logs:
![image](https://github.com/threefoldtech/zos/assets/32527347/0a16b704-bc06-4afc-ac05-4860f2e1bcef)

I did some research for any reference of running `crc64` on a set of cpus but couldn't find any.

muhamadazmy:
We need explanation (docs) on what that score mean. What is a good score and what is a bad score. etc.. create an issue on the cpu-benchmark repo and assign to maxime to give an explanation

Omarabdul3ziz:
check maxime comments on the issue https://github.com/threefoldtech/cpu-benchmark-simple/issues/1#issuecomment-1820051375

khaledyoussef24:
so after checking maxime comment 
the results of the bench cpu performance refer to the efficency of know cpu 
so 
for node 12 on devnet we get

```
   "description": "Measures the performance of the node CPU by reporting the timespent of computing a task in seconds.",
    "name": "cpu-benchmark",
    "result": {
      "multi": 60.999,
      "single": 2.433,
      "threads": 24,
      "workloads": 67
    }, 
```

which is nearly to the same performance as 
2x Intel(R) Xeon(R) CPU L5640 @ 2.27GHz (Q1'2010)

with specs 
```
Single thread score: 2.504
Multi threads score: 72.452 [24 threads]
```

so it's more clear after reading the examples maxime provided 

can this example be added to the docs so it will be more clear for those who run the test ?

ramezsaeed:
Please add the clarification and existing example to the docs, so users can use to compare results.

khaledyoussef24:
[...]

PR #2119, opened 2023-11-19
fixes for perf pacakge

Omarabdul3ziz:
### Description
- add a description field in the task report
- use graphql instead of substrate to decide the node status
- define payload type for `perf.get` call
- unify tasks names to kabab case

### related issues
- https://github.com/threefoldtech/zos/issues/2112
- https://github.com/threefoldtech/zos/issues/2114
- https://github.com/threefoldtech/zos/issues/2115

### DIFF
--- a/pkg/perf/README.md
+++ b/pkg/perf/README.md
@@ -10,8 +10,8 @@ Perf tests are monitored by `noded` service from zos modules.
 
 ```go
 type LoggingTask struct {
-	taskID   	string
-	schedule 	string // should be in cron syntax with seconds ("* 0 * * * *")
+	taskID      string
+	schedule    string // should be in cron syntax with seconds ("* 0 * * * *")
 	description string
 }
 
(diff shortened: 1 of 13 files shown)


pkg/perf/README.md
```diff
@@ -33,7 +38,8 @@ func (t *LoggingTask) Run(ctx context.Context) (interface{}, error) {
 
 ```go
 perfMon.AddTask(&perf.LoggingTask{
-    TaskID:   "LoggingTask",
-    Schedule: "* 0 * * * *", // when minutes is 0 (aka: every hour)
+    taskID:   "LoggingTask",
+    schedule: "* 0 * * * *", // when minutes is 0 (aka: every hour)
+	description: "Simple task that logs the time every hour"
```
muhamadazmy: Please USE a code formatter in your editor

pkg/perf/cpubench/cpubench_task.go
```diff
@@ -10,15 +10,20 @@ import (
 	"github.com/threefoldtech/zos/pkg/stubs"
 )
 
-const cpuBenchmarkTaskID = "CPUBenchmark"
-const cpuBenchmarkCronSchedule = "0 0 */6 * * *"
+const (
+	cpuBenchmarkTaskID       = "CPUBenchmark"
```
muhamadazmy: since this will end up in json, i actually don't like CamelCase as json keys. It's better if we then use kebab-case or snake_case
I suggest `cpu-benchmark` it looks better in json output

muhamadazmy (changes requested):

muhamadazmy (approved):

Issue #2111, opened 2023-11-15
iperf tests run on offline nodes (should run on online nodes)


AhmedHanafy725:
Env: Devnet
nodeId: 12

### Steps 
- Get iperf test result through RMB call `zos.perf.get_all`.

### Result

All the tests are failing. So the user would say the issue is in node 12 but the real issue is some tests are triggered against offline nodes (like 46, 50, 91) or up nodes but always get this error message `the server is busy running a test. try again later`

```ts
{
  name: 'iperf',
  timestamp: 1700029067,
  result: [
    {
      cpu_report: {
        host_system: 0,
        host_total: 0,
        host_user: 0,
        remote_system: 0,
        remote_total: 0,
        remote_user: 0
      },
      download_speed: 0,
      error: 'unable to connect to server - server may have stopped running or use a different port, firewall issue, etc.: Host is unreachable',
      node_id: 46,
      node_ip: '162.205.240.144',
      test_type: 'tcp',
      upload_speed: 0
    },
    {
      cpu_report: {
        host_system: 0,
        host_total: 0,
        host_user: 0,
        remote_system: 0,
        remote_total: 0,
        remote_user: 0
      },
      download_speed: 0,
      error: 'unable to connect to server - server may have stopped running or use a different port, firewall issue, etc.: Operation timed out',
      node_id: 46,
      node_ip: '2600:1700:c1e0:1150:f4d0:96ff:fec9:ecca',
      test_type: 'tcp',
      upload_speed: 0
    },
    {
      cpu_report: {
        host_system: 0,
        host_total: 0,
        host_user: 0,
        remote_system: 0,
        remote_total: 0,
        remote_user: 0
      },
      download_speed: 0,
      error: 'unable to connect to server - server may have stopped running or use a different port, firewall issue, etc.: Host is unreachable',
      node_id: 46,
      node_ip: '162.205.240.144',
      test_type: 'udp',
      upload_speed: 0
    },
    {
      cpu_report: {
        host_system: 0,
        host_total: 0,
        host_user: 0,
        remote_system: 0,
        remote_total: 0,
        remote_user: 0
      },
      download_speed: 0,
      error: 'unable to connect to server - server may have stopped running or use a different port, firewall issue, etc.: Operation timed out',
      node_id: 46,
      node_ip: '2600:1700:c1e0:1150:f4d0:96ff:fec9:ecca',
      test_type: 'udp',
      upload_speed: 0
    },
    {
      cpu_report: {
[...]

PR #2120, opened 2023-11-19
filter up nodes

rawdaGastan:
### Issues

- https://github.com/threefoldtech/zos/issues/2111


pkg/perf/iperf/graphql_nodes.go
```diff
@@ -57,7 +57,9 @@ func (g *GraphQl) ListPublicNodes(ctx context.Context, nodesNum int, farmID, exc
 		excludeFarmCond = fmt.Sprintf(", farmID_not_eq: %d", excludeFarmID)
 	}
 
-	whereCond := fmt.Sprintf(`where: { publicConfig: {%s} %s %s }`, pubCond, farmCond, excludeFarmCond)
+	nodeUpReportInterval := time.Minute * 40
+	nodeUpInterval := time.Now().Unix() - 2*int64(nodeUpReportInterval.Seconds())
+	whereCond := fmt.Sprintf(`where: { updatedAt_gte: %d, AND: {power_isNull: true, OR: {power: {state_eq: Up, target_eq: Up}}}, publicConfig: {%s} %s %s }`, nodeUpInterval, pubCond, farmCond, excludeFarmCond)
```
muhamadazmy, line 62: I just have a question. is updatedAt is actually updated on node uptime ?

muhamadazmy: I also noticed that some nodes `power` object is completely not set in graphql ( == null) because there were never an event to set it up or down by a farmer bot.

A node that it's power state/traged were never updated has no data. The chain on the other hand is programmed to return the default when u do NodeGetPower if there is no value set.

rawdaGastan: I checked the proxy status filter for it
- https://github.com/threefoldtech/tfgrid-sdk-go/blob/development/grid-proxy/pkg/nodestatus/nodestatus.go#L37

muhamadazmy (dismissed):

pkg/perf/graphql/graphql_nodes.go
```diff
@@ -57,7 +57,9 @@ func (g *GraphQl) ListPublicNodes(ctx context.Context, nodesNum int, farmID, exc
 		excludeFarmCond = fmt.Sprintf(", farmID_not_eq: %d", excludeFarmID)
 	}
 
-	whereCond := fmt.Sprintf(`where: { publicConfig: {%s} %s %s }`, pubCond, farmCond, excludeFarmCond)
+	nodeUpReportInterval := time.Minute * 40
+	nodeUpInterval := time.Now().Unix() - 2*int64(nodeUpReportInterval.Seconds())
+	whereCond := fmt.Sprintf(`where: { updatedAt_gte: %d, AND: {power_isNull: true, OR: {power: {state_eq: Up, target_eq: Up}}}, publicConfig: {%s} %s %s }`, nodeUpInterval, pubCond, farmCond, excludeFarmCond)
```
muhamadazmy, line 62: Just one question before approval, does the updateAt is updated when the node uptime report received ? I want to make sure this is a safe check that does what we think it does.

muhamadazmy (approved):

Issue #2128, opened 2023-11-20
integer divide by zero error on storage module


Omarabdul3ziz:
## Describe the bug
faced this issue on the logs when i tried to run more than one zos node on the same machine.

## To Reproduce
not sure if this is the only way to trigger this behavior but maybe running two zos nodes at the same time

## Screenshots
![image](https://github.com/threefoldtech/zos/assets/32527347/eefe0961-7614-486b-9aa4-6f9ba1ac913c)

PR #2129, opened 2023-11-20
set default value of 5GB for cache size it is zero

Omarabdul3ziz:
### Description
set default value of 5GB for cache size it is zero

### Related Issues
- https://github.com/threefoldtech/zos/issues/2128

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -658,10 +658,13 @@ func (s *Module) checkAndResizeCache(cache filesystem.Volume, sizeMultiplier uin
 		return errors.Wrap(err, "failed to check cache usage")
 	}
 	if usage.Size == 0 {
-		usage.Size = 5
+		usage.Size = max(sizeMultiplier, usage.Excl*2)
+		if err := cache.Limit(usage.Size); err != nil {
+			return err
+		}
 	}
-	log.Debug().Msgf("cache usage %+v", usage)
 
+	log.Debug().Msgf("cache usage %+v", usage)
 	percent := usage.Excl * 100 / usage.Size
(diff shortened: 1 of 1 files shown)


xmonader:
Did you try to mount other disks in the new VM? other than the ones used by the first one? check vm.sh first please this seems like hiding a problem

pkg/storage/storage.go
```diff
@@ -657,7 +657,11 @@ func (s *Module) checkAndResizeCache(cache filesystem.Volume, sizeMultiplier uin
 	if err != nil {
 		return errors.Wrap(err, "failed to check cache usage")
 	}
+	if usage.Size == 0 {
+		usage.Size = 5
```
muhamadazmy: 5 what? :D 
When i said set the size of the cache to 5. I meant 5gib. and that's not just a random number, it's the sizeMultiplier. 

But anyway, this is not exactly i wanted it to be done. The idea is if the cache.Size is not set (no limit) I wanted to make sure the `limit` is actually set first, before you even try to adjust.

```go
if useage.Size == 0 {
   usge.Size = max(sizeMultiplier, usage.Excl * 2)
   if err := cache.Limit(usage.Size); err != nil {
       return err
   }
}
// ... rest of the function
```

muhamadazmy: To test this, make sure your cache volume has no quota set then restart storaged. then the cache should have some limit

muhamadazmy (changes requested):

muhamadazmy (approved):

Issue #2109, opened 2023-11-14
 Clean up VM host key before first boot


muhamadazmy:
Related to https://github.com/threefoldtech/tf-images/issues/165

Many flists has the host key built in the image, this causes a problem because now all instances created from that the same flist will end up having the same hostkey which is not secure!

According to @scottyeager it would be better if zos takes care on cleaning up the host keys from the VM rootfs before it boots for the **first time** 

After booting, when the image generate a new key the key should not be touched. The image scripts should also take care to generate the keys when they don't exist

Of course  a good image should not include the host key in the base image, but it seems it's a common mistake that people leave that in there by accident

scottyeager:
Hi @muhamadazmy, thanks for the issue.

To be clear, my suggestion is to handle both the deletion of host keys and the generation of fresh ones automatically on first boot. This is to mirror the functionality provided by `clout-init` in our full VMs.

If we only delete the host keys then we'll break SSH in many existing images and force all maintainers (including ourselves) to update. I'd rather not take this halfway step unless we can agree on the full approach.

Let's pause this for further discussion.

muhamadazmy:
I will definitely not do generation of host keys @scottyeager ! Again, already i am not fully convinced that deletion of the key is the right thing to do.

ZOS should be completely transparent to workload types, deletion (or generation) means that zos assumes something regarding the workload and/or operation of the workload which is wrong in the firsplace.

scottyeager:
Let's go ahead with the deletion feature here, and I'll coordinate with our image maintainers to make sure the images generate host keys as needed.

That said, I don't agree with the assessments about what is "wrong" behavior for Zos. It exists to serve users, not philosophical ideals!

My suggestion, again, is simply that we align behavior for our micro VMs with the industry standards for full VMs, where the possibility of host keys sometimes getting shipped in public images is just a fact of life, not a "mistake", and the erasure and regeneration of such keys is considered sane default behavior.

But even though we don't agree on that, doing the part we agree on here can at least help protect all grid users without requiring compliance from third party image authors, so I don't want to hold that up any further.

Eslam-Nawara:
Investigation and Solution:

- currently, I'm investigating where to add the update and trying to create a zos node to test the updates but it gets stuck for some reason with no reported error

Eslam-Nawara:
Work in Progress (WIP): 
- created a local node to test the update against 
- started updating zos to delete vm host keys on first boot

Eslam-Nawara:
[...]

PR #2138, opened 2023-11-27
clean up host keys before first boot

Eslam-Nawara:
### Description

remove host keys on vm run

### Related Issues

- #2109

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -70,6 +70,7 @@ require (
 	github.com/coreos/go-iptables v0.6.0 // indirect
 	github.com/cosmos/go-bip39 v1.0.0 // indirect
 	github.com/cpuguy83/go-md2man/v2 v2.0.2 // indirect
+	github.com/dave/jennifer v1.3.0 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/decred/dcrd/crypto/blake256 v1.0.1 // indirect
 	github.com/decred/dcrd/dcrec/secp256k1/v3 v3.0.0 // indirect
--- a/go.mod
+++ b/go.mod
@@ -70,7 +70,6 @@ require (
 	github.com/coreos/go-iptables v0.6.0 // indirect
 	github.com/cosmos/go-bip39 v1.0.0 // indirect
 	github.com/cpuguy83/go-md2man/v2 v2.0.2 // indirect
-	github.com/dave/jennifer v1.3.0 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/decred/dcrd/crypto/blake256 v1.0.1 // indirect
 	github.com/decred/dcrd/dcrec/secp256k1/v3 v3.0.0 // indirect
--- a/go.sum
+++ b/go.sum
@@ -106,6 +106,7 @@ github.com/d2g/dhcp4 v0.0.0-20170904100407-a1d1b6c41b1c/go.mod h1:Ct2BUK8SB0YC1S
 github.com/d2g/dhcp4client v1.0.0/go.mod h1:j0hNfjhrt2SxUOw55nL0ATM/z4Yt3t2Kd1mW34z5W5s=
 github.com/d2g/dhcp4server v0.0.0-20181031114812-7d4a0a7f59a5/go.mod h1:Eo87+Kg/IX2hfWJfwxMzLyuSZyxSoAug2nGa1G2QAi8=
 github.com/d2g/hardwareaddr v0.0.0-20190221164911-e7d9fbe030e4/go.mod h1:bMl4RjIciD2oAxI7DmWRx6gbeqrkoLqv3MV0vzNad+I=
+github.com/dave/jennifer v1.3.0 h1:p3tl41zjjCZTNBytMwrUuiAnherNUZktlhPTKoF/sEk=
 github.com/dave/jennifer v1.3.0/go.mod h1:fIb+770HOpJ2fmN9EPPKOqm1vMGhB+TwXKMZhrIygKg=
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
--- a/go.sum
+++ b/go.sum
@@ -106,7 +106,6 @@ github.com/d2g/dhcp4 v0.0.0-20170904100407-a1d1b6c41b1c/go.mod h1:Ct2BUK8SB0YC1S
 github.com/d2g/dhcp4client v1.0.0/go.mod h1:j0hNfjhrt2SxUOw55nL0ATM/z4Yt3t2Kd1mW34z5W5s=
 github.com/d2g/dhcp4server v0.0.0-20181031114812-7d4a0a7f59a5/go.mod h1:Eo87+Kg/IX2hfWJfwxMzLyuSZyxSoAug2nGa1G2QAi8=
 github.com/d2g/hardwareaddr v0.0.0-20190221164911-e7d9fbe030e4/go.mod h1:bMl4RjIciD2oAxI7DmWRx6gbeqrkoLqv3MV0vzNad+I=
-github.com/dave/jennifer v1.3.0 h1:p3tl41zjjCZTNBytMwrUuiAnherNUZktlhPTKoF/sEk=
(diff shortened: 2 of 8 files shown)


pkg/vm/manager.go
```diff
@@ -524,6 +525,13 @@ func (m *Module) Run(vm pkg.VM) (pkg.MachineInfo, error) {
 		return pkg.MachineInfo{}, m.withLogs(m.logsPath(vm.Name), err)
 	}
 
+	// clean up host keys before boot
+	command := exec.Command("rm", "-r", "/etc/ssh/ssh_host_*")
```
muhamadazmy: - Why are you executing a full subprocess command `rm` to delete file, not using the ? a remove is basically a system call that you can execute progrmatically like in go it's available under `os.Remove`  and family of functions
- The `*` is only wild card if used i the shell. In other words when u run `rm *` in a shell (like bash) it expands the * to all matches FIRST before it execute the command with an exact argument list. It means what you are doing here is basically calling the rm command with an exact filename `ssh_host_*` 
- Second, this happens every time the vm is started not on first boot 
- This removed the host keys of ZOS itself !! not the VM !!  

What you need to do in the `provision` daemon, and after mounting the flist is that you check if it has those files and delete them, IF AND ONLY IF you are booting this VM for the first time

muhamadazmy: Not that you should but if you really really really wants to use a `wildcard expansion` then what you need to do is `exec.Command("bash", "-c", "rm whatever*")` to make sure the bash expands the * before.

Which is ALSO not needed because go provide you with https://pkg.go.dev/path/filepath#Glob function

pkg/primitives/vm/utils.go
```diff
@@ -459,3 +480,10 @@ func mergeEnvs(base, new map[string]string) map[string]string {
 
 	return base
 }
+
+func (p *Manager) volumeExist(volName string) bool {
+	ctx := context.Background()
+	storage := stubs.NewStorageModuleStub(p.zbus)
+	_, err := storage.VolumeLookup(ctx, volName)
+	return err == nil
+}
```
muhamadazmy: Why this is a separate function, i see this doesn't make it better because:
- U are using the wrong context? this context since u simply creating a new empty context
- You recreate the storage stubs and u also recreate that in the caller.

pkg/primitives/vm/utils.go
```diff
@@ -156,6 +156,27 @@ func (p *Manager) prepContainer(
 		machine.Entrypoint = "/corex --ipv6 -d 7 --interface eth0"
 	}
 
+	// clean up host keys
+	firstBootVolumeName := fmt.Sprintf("firstBoot:%s", wl.ID.String())
```
muhamadazmy: What is this ? why do u create a new volume ? 

Look @Eslam-Nawara i think you completely got this completely wrong

The code above, create a volume where the flist will be mounted. (line 106) above.
You can think that this storage volume will hold the files from the mounted flist. Good ? The function storage.VolumeCreate will create a new one if the volume does not exist already this happen on the first boot of the vm.

This means, if I detect that this volume didn't exist before this call, it means i am booting this vm for the first time hence i can safely decide to delete the files **from that volume** (after the mount call)

What you do here is that u create completely out of context a new volume that u call firstBootVolume !!! 
Then what u do is u still immediately delete all files from the **zos  host** not from the Volume

What you will u need to do eventually is to delete the files from the the 
```go
filepath.Join(volume.Path, "etc", "ssh", "ssh_host_*")
```

muhamadazmy (changes requested):
This is completely wrong, please check comments

pkg/primitives/vm/utils.go
```diff
@@ -102,11 +102,18 @@ func (p *Manager) prepContainer(
...
-	volume, err := storage.VolumeCreate(ctx, volName, rootfsSize)
-	if err != nil {
-		return errors.Wrap(err, "failed to create vm rootfs")
-	}
+
+    volName := fmt.Sprintf("rootfs:%s", wl.ID.String())
+    volume, err := storage.VolumeLookup(ctx, volName)
+    isFirstBoot := false
+    if err != nil {
+        isFirstBoot = true
```
rawdaGastan: I think you should check on `os.ErrNotExist`

pkg/primitives/vm/utils.go
```diff
@@ -103,6 +100,12 @@ func (p *Manager) prepContainer(
 	// via the flist, so we have control over when to decomission this volume.
 	// remounting in RW mode
 	volName := fmt.Sprintf("rootfs:%s", wl.ID.String())
+
+	isFirstBoot, err := storage.VolumeExists(ctx, volName)
```
muhamadazmy: `isFirstBoot` is a weird name for that variable because, If true, it means it's NOT first boot, right ? So now instead of cleaning the keys on actually the first boo, the cleaning happens on any boot that is not the frist boot.

Instead i would name this `volumeExists` instead. It's relevant to the actual value that it holds.

pkg/primitives/vm/utils.go
```diff
@@ -121,11 +124,22 @@ func (p *Manager) prepContainer(
 		ReadOnly:        false,
 		PersistedVolume: volume.Path,
 	})
-
 	if err != nil {
 		return errors.Wrapf(err, "failed to mount flist: %s", wl.ID.String())
 	}
 
+	// clean up host keys
+	if isFirstBoot {
```
muhamadazmy: Check the comment above, because now this works if the volume exists, which means **not** the first boot (since the volume existed before)

pkg/primitives/vm/utils.go
```diff
@@ -121,11 +124,22 @@ func (p *Manager) prepContainer(
...
+		files, err := filepath.Glob(filepath.Join(mnt, "etc", "ssh", "ssh_host_*"))
+		if err != nil {
+			log.Fatal().Err(err)
```
muhamadazmy: Are you aware that `Fatal` will cause the entire process to panic and crash? any call to `log.Fatal()` will immediately exist the process after logging the error because there is no way you can continue after a **fatal** error.

This error imho is completely ignorable, it's not an error that should be even logged as an error because this does not affect the operation of the system. Instead I would log this as a debug message and them move on

Also according to docs of `filepath.Glob`

> Glob ignores file system errors such as I/O errors reading directories. The only possible returned error is ErrBadPattern, when pattern is malformed.

Which means this will never return an error unless your `*` is wrong which is not the case here. (if files or path doesn't exist it won't mind)

```suggestion
			log.Debug().Err(err).Msg("failed to list ssh host keys for a vm image")
```

one last thing, A log with No `Msg()` call will not print anything. The only way for this error to print is to terminate it with a Message like in the suggestion above.

pkg/primitives/vm/utils.go
```diff
@@ -121,11 +124,22 @@ func (p *Manager) prepContainer(
...
+		}
+
+		for _, file := range files {
+			os.Remove(file)
```
muhamadazmy:
```suggestion
			if err := os.Remove(file); err != nil {
				log.Debug().Err(err).Str("file", file).Msg("failed to delete host key file")
			}
```

muhamadazmy (changes requested):

go.mod
```diff
@@ -70,6 +70,7 @@ require (
 	github.com/coreos/go-iptables v0.6.0 // indirect
 	github.com/cosmos/go-bip39 v1.0.0 // indirect
 	github.com/cpuguy83/go-md2man/v2 v2.0.2 // indirect
+	github.com/dave/jennifer v1.3.0 // indirect
```
muhamadazmy: Why is that was added? Please remove.

go.sum
```diff
@@ -106,6 +106,7 @@ github.com/d2g/dhcp4 v0.0.0-20170904100407-a1d1b6c41b1c/go.mod h1:Ct2BUK8SB0YC1S
 github.com/d2g/dhcp4client v1.0.0/go.mod h1:j0hNfjhrt2SxUOw55nL0ATM/z4Yt3t2Kd1mW34z5W5s=
 github.com/d2g/dhcp4server v0.0.0-20181031114812-7d4a0a7f59a5/go.mod h1:Eo87+Kg/IX2hfWJfwxMzLyuSZyxSoAug2nGa1G2QAi8=
 github.com/d2g/hardwareaddr v0.0.0-20190221164911-e7d9fbe030e4/go.mod h1:bMl4RjIciD2oAxI7DmWRx6gbeqrkoLqv3MV0vzNad+I=
+github.com/dave/jennifer v1.3.0 h1:p3tl41zjjCZTNBytMwrUuiAnherNUZktlhPTKoF/sEk=
```
muhamadazmy: That of course as well

muhamadazmy (changes requested):

muhamadazmy (approved):
Thank you

Issue #2134, opened 2023-11-24
Stop uptime reports for nodes with read only cache disk


scottyeager:
Recently I've seen a couple more cases like #1983, where the node logs indicate read only filesystem and no workloads are accepted. I'll include details below for reference.

The thing is that these nodes are still generating uptime reports, and thus minting. Since we only want to mint for nodes that provide usable capacity, I wonder if we can detect this condition in Zos and stop uptime reporting when it happens.

### Node 949
```
[+] provisiond: 2023-11-17T21:28:40Z fatal exiting error="error running integrity checks: unlinkat /var/cache/modules/provisiond/metrics-diff.bolt: read-only file system"
[+] provisiond: 2023-11-17T21:28:40Z error integrity check failed, resetting rrd db error="exit status 1" stderr="failed to create metrics database: failed to open bolt db: open /var/cache/modules/provisiond/metrics-diff.bolt: read-only file system\n"
```

### Node 1941
```
provisiond: 2023-11-17T22:11:41Z error capacity reported stopped unexpectedly error="failed to create rrd slot: failed to apply retnetion policy: write /var/cache/modules/provisiond/metrics-diff.bolt: read-only file system"
```

ramezsaeed:
we should show a relevant message in zui, and stop sending uptime reports until farmers fix that.

scottyeager:
Yes, there's a spot for cache disk status on zui. It can say "Read only" in red.

AbdelrahmanElawady:
for nodes that managed to get the update on mainnet, uptime reports stopped:
![image](https://github.com/threefoldtech/zos/assets/60783742/8c5d6a7f-ecb6-4319-a758-b0402f15d316)
however, the majority didn't get the latest version due to disk issues. which should be addressed here: #2228 #2227

PR #2143, opened 2023-11-28
Add healthcheck to report to zui for errors

AbdelrahmanElawady:
### Description

Add healthcheck task to execute multiple checks (only cache for now) and report to zui for errors. It also set flags that can be checked by Power daemon to stop uptime reports.

### Changes

- Add and register zui module on zbus to allow other modules to set some errors in the ui.
- Add healthcheck Task that perform checks on the node to make sure everything is OK.
- Add another part in the ui to display errors with the node state.
- Stop uptime reports if cache is unusable (readonly or limited).

### Related Issues

- https://github.com/threefoldtech/zos/issues/2134

### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation

### DIFF
--- a/cmds/modules/zui/header.go
+++ b/cmds/modules/zui/header.go
@@ -76,7 +76,7 @@ func headerRenderer(ctx context.Context, c zbus.Client, h *widgets.Paragraph, r 
 			}
 
 			cache := green("OK")
-			if app.CheckFlag(app.LimitedCache) {
+			if app.CheckFlag(app.LimitedCache) || app.CheckFlag(app.ReadonlyCache) {
 				cache = red("no ssd disks detected")
 			}
 
--- a/cmds/modules/zui/header.go
+++ b/cmds/modules/zui/header.go
@@ -76,8 +76,10 @@ func headerRenderer(ctx context.Context, c zbus.Client, h *widgets.Paragraph, r 
 			}
 
 			cache := green("OK")
-			if app.CheckFlag(app.LimitedCache) || app.CheckFlag(app.ReadonlyCache) {
+			if app.CheckFlag(app.LimitedCache) {
 				cache = red("no ssd disks detected")
+			} else if app.CheckFlag(app.ReadonlyCache) {
+				cache = red("cache is read-only")
 			}
 
 			var utsname syscall.Utsname
--- a/go.mod
+++ b/go.mod
@@ -70,6 +70,7 @@ require (
 	github.com/coreos/go-iptables v0.6.0 // indirect
 	github.com/cosmos/go-bip39 v1.0.0 // indirect
 	github.com/cpuguy83/go-md2man/v2 v2.0.2 // indirect
+	github.com/dave/jennifer v1.3.0 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/decred/dcrd/crypto/blake256 v1.0.1 // indirect
 	github.com/decred/dcrd/dcrec/secp256k1/v3 v3.0.0 // indirect
--- a/go.mod
+++ b/go.mod
@@ -70,7 +70,6 @@ require (
 	github.com/coreos/go-iptables v0.6.0 // indirect
 	github.com/cosmos/go-bip39 v1.0.0 // indirect
 	github.com/cpuguy83/go-md2man/v2 v2.0.2 // indirect
-	github.com/dave/jennifer v1.3.0 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/decred/dcrd/crypto/blake256 v1.0.1 // indirect
 	github.com/decred/dcrd/dcrec/secp256k1/v3 v3.0.0 // indirect
--- a/go.sum
+++ b/go.sum
@@ -106,6 +106,7 @@ github.com/d2g/dhcp4 v0.0.0-20170904100407-a1d1b6c41b1c/go.mod h1:Ct2BUK8SB0YC1S
 github.com/d2g/dhcp4client v1.0.0/go.mod h1:j0hNfjhrt2SxUOw55nL0ATM/z4Yt3t2Kd1mW34z5W5s=
 github.com/d2g/dhcp4server v0.0.0-20181031114812-7d4a0a7f59a5/go.mod h1:Eo87+Kg/IX2hfWJfwxMzLyuSZyxSoAug2nGa1G2QAi8=
 github.com/d2g/hardwareaddr v0.0.0-20190221164911-e7d9fbe030e4/go.mod h1:bMl4RjIciD2oAxI7DmWRx6gbeqrkoLqv3MV0vzNad+I=
+github.com/dave/jennifer v1.3.0 h1:p3tl41zjjCZTNBytMwrUuiAnherNUZktlhPTKoF/sEk=
 github.com/dave/jennifer v1.3.0/go.mod h1:fIb+770HOpJ2fmN9EPPKOqm1vMGhB+TwXKMZhrIygKg=
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
--- a/go.sum
+++ b/go.sum
@@ -106,8 +106,6 @@ github.com/d2g/dhcp4 v0.0.0-20170904100407-a1d1b6c41b1c/go.mod h1:Ct2BUK8SB0YC1S
 github.com/d2g/dhcp4client v1.0.0/go.mod h1:j0hNfjhrt2SxUOw55nL0ATM/z4Yt3t2Kd1mW34z5W5s=
 github.com/d2g/dhcp4server v0.0.0-20181031114812-7d4a0a7f59a5/go.mod h1:Eo87+Kg/IX2hfWJfwxMzLyuSZyxSoAug2nGa1G2QAi8=
 github.com/d2g/hardwareaddr v0.0.0-20190221164911-e7d9fbe030e4/go.mod h1:bMl4RjIciD2oAxI7DmWRx6gbeqrkoLqv3MV0vzNad+I=
-github.com/dave/jennifer v1.3.0 h1:p3tl41zjjCZTNBytMwrUuiAnherNUZktlhPTKoF/sEk=
-github.com/dave/jennifer v1.3.0/go.mod h1:fIb+770HOpJ2fmN9EPPKOqm1vMGhB+TwXKMZhrIygKg=
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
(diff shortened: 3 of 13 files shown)


AbdelrahmanElawady:
Here is how it looks on the ui:
image_2023-11-28_18-19-15
it was tested using `chattr +i /var/cache` to make sure creating the file is not possible.

still missing:
- checking the error flag in the uptime goroutine.
- wrapping the errors in zui if the errors are large.
- some docs.

scottyeager:
This view is a bit confusing, since it shows `cache disk: OK` above and an error on cache below. I think it's better to write the error in place of the `OK` message for the cache. If future health checks create the need for a separate errors section, it can be added later.

AbdelrahmanElawady:
It is showing OK because a check for Readonly cache hasn't been added yet, it will be added with the other remaining changes. A separate section for errors in ZOS node is more helpful to get more details of what is wrong without cluttering the first section (it already has a lot of information) and to support multiple health checks. But that's my opinion.

scottyeager:
Okay, that makes sense, @AbdelrahmanElawady. In that case it looks good to me.

cmds/modules/zui/header.go
```diff
@@ -76,7 +76,7 @@ func headerRenderer(ctx context.Context, c zbus.Client, h *widgets.Paragraph, r
 			}
 
 			cache := green("OK")
-			if app.CheckFlag(app.LimitedCache) {
+			if app.CheckFlag(app.LimitedCache) || app.CheckFlag(app.ReadonlyCache) {
```
muhamadazmy: I suggest we show a different message if the cache is `read-only`. Because now `no ssd disks detected` is an abovious lie 😆

go.mod
```diff
@@ -70,6 +70,7 @@ require (
 	github.com/coreos/go-iptables v0.6.0 // indirect
 	github.com/cosmos/go-bip39 v1.0.0 // indirect
 	github.com/cpuguy83/go-md2man/v2 v2.0.2 // indirect
+	github.com/dave/jennifer v1.3.0 // indirect
```
muhamadazmy: I don't know why this keeps getting added to go.mod but it's not needed. it must be zbus that add this dependency .

Please remove

muhamadazmy: you should notice when the go.mod is modified when u didn't modify it intentionally

go.sum
```diff
@@ -106,6 +106,7 @@ github.com/d2g/dhcp4 v0.0.0-20170904100407-a1d1b6c41b1c/go.mod h1:Ct2BUK8SB0YC1S
 github.com/d2g/dhcp4client v1.0.0/go.mod h1:j0hNfjhrt2SxUOw55nL0ATM/z4Yt3t2Kd1mW34z5W5s=
 github.com/d2g/dhcp4server v0.0.0-20181031114812-7d4a0a7f59a5/go.mod h1:Eo87+Kg/IX2hfWJfwxMzLyuSZyxSoAug2nGa1G2QAi8=
 github.com/d2g/hardwareaddr v0.0.0-20190221164911-e7d9fbe030e4/go.mod h1:bMl4RjIciD2oAxI7DmWRx6gbeqrkoLqv3MV0vzNad+I=
+github.com/dave/jennifer v1.3.0 h1:p3tl41zjjCZTNBytMwrUuiAnherNUZktlhPTKoF/sEk=
```
muhamadazmy: Same here

pkg/perf/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,97 @@
...
+func (h *healthcheckTask) Description() string {
+	return description
+}
+
+// Run executes the health checks.
+func (h *healthcheckTask) Run(ctx context.Context) (interface{}, error) {
+	log.Debug().Msg("starting health check task")
+	for k := range h.errors {
+		// reset errors on each run
+		h.errors[k] = make([]string, 0)
```
muhamadazmy: doing 
```suggestion
		h.errors[k] = nil
```
is faster and has the same effect in your case.

muhamadazmy: remeber that a `nil` array is a valid `empty` array. a call to `a = append(a, v)` will initialize the array automatically if needed.

pkg/perf/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,97 @@
...
+	cl := perf.GetZbusClient(ctx)
+	zui := stubs.NewZUIStub(cl)
+
+	for label, data := range h.errors {
+		zui.PushErrors(ctx, label, data)
+	}
+	return h.errors, nil
+}
+
+func cacheCheck(ctx context.Context) (string, error) {
```
muhamadazmy: returning the label is weird imho. The better approach is that the `checks` on the task itself can instead be a map of the (label, checkFunc)
```go
checks := map[string]checkFunc{
   "cache", cacheCheck
}
```
which means, your funcion can just return an error, and the loop will know what label to use

pkg/power/uptime.go
```diff
@@ -71,6 +72,23 @@ func (u *Uptime) uptime(ctx context.Context) error {
...
+		}
+		if app.CheckFlag(app.LimitedCache) {
+			log.Error().Msg("node is running on limited cache")
+			unusable = true
+		}
+		if unusable {
+			log.Error().Msg("node is not usable skipping uptime reports")
+		} else if err := report(); err != nil {
+			return err
+		}
```
muhamadazmy: why it's done like this but no like
```suggestion
		if app.CheckFlag(app.ReadonlyCache) {
			log.Error().Msg("node cache is read only")
		} else if app.CheckFlag(app.LimitedCache) {
			log.Error().Msg("node is running on limited cache")
		} else if err := report(); err != nil {
			return err
		}
```

Also I think that's the wrong place, because the SendNow function can sitll be called from other places (for exasmple node shutdown) when it actually shouldn't.

Maybe the better place to palce these checks in the SendNow fuction.

pkg/zui.go
```diff
@@ -0,0 +1,7 @@
+package pkg
+
+//go:generate zbusc -module zui -version 0.0.1 -name zui -package stubs github.com/threefoldtech/zos/pkg+ZUI stubs/zui_stub.go
+
+type ZUI interface {
+	PushErrors(label string, errors []string)
```
muhamadazmy:
```suggestion
	PushErrors(label string, errors []string) error
```

Just to make the linter happy about the genreate stubs

pkg/zui/zui.go
```diff
@@ -0,0 +1,87 @@
...
+			return
+		default:
+			m.mu.Lock()
+			labels := make([]labelData, len(m.labels))
+			copy(labels, m.labels)
+			m.mu.Unlock()
+			display(labels, m.paragraph, m.render)
+			// in case nothing got displayed
+			<-time.After(2 * time.Second)
+		}
```
muhamadazmy:
```suggestion
		select {
		case <-ctx.Done():
			return
		case <-time.After(2 * time.Second):
			m.mu.Lock()
			labels := make([]labelData, len(m.labels))
			copy(labels, m.labels)
			m.mu.Unlock()
			display(labels, m.paragraph, m.render)
		}
```

your code means that during the wait (for the 2 seconds) you will not be able to detect the cacncelled context

pkg/zui/zui.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func display(labels []labelData, p *widgets.Paragraph, render Signaler) {
+	p.Text = "[No Errors!](fg:green)"
+
+	for _, label := range labels {
+		for _, e := range label.errors {
+			p.Text = fmt.Sprintf("%s: [%s](fg:red)", label.label, e)
+			render.Signal()
+			<-time.After(2 * time.Second)
```
muhamadazmy: You need to select between the sleep and the ctx.Done() here as well otherwise that will block even if the ctx is cancelled.

muhamadazmy (changes requested):

pkg/perf/healthcheck/healthcheck.go
AbdelrahmanElawady: what I had in mind is that there could be multiple `checkFunc` checking the same module instead of one big function responsible for just `cache` or `network`. So, in case of `cache` one will check the read-only thing other will maybe check reported capacity and so on. should it be just one function responsible for the whole module (or capacity)?

pkg/power/uptime.go
AbdelrahmanElawady: the assumption was that both flags (and other flags added) can occur at the same time so we need to log to the farmer every problem with the node. or should we just report one?

pkg/perf/healthcheck/healthcheck.go
muhamadazmy: I get what you mean, and the idea itself is good. But i still find it weird that the function has to return the label.

Other possible way to get the same effect is 
- `map[string][]checkFunc`  
- `checkFunc() []error` where the checkFunc can return a list of errors which means it internally can run mutliple checks and return a list of possible health errors.

```
health() []error {
   var issues []error
   if err := cacheCheck() err = !nil {
        issues = append(issues, err)
   }
   // more checks

  return issues
}
```

pkg/power/uptime.go
muhamadazmy: regardless if both can happen on the same time or not (they won't) but i don't think the logs need to be this verbose here (in other words it doesn't really matter here what is the reason we skipping the uptime) because:
- The healthcheck will still log the proper rason 
- The Ui will still show the proper reason.

This makes me think actually this can be even more comact by doing `if is limite-cache or is reasony, then skip`

pkg/perf/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,116 @@
...
+func errorsToStrings(errs []error) []string {
+	s := make([]string, len(errs))
+	for i, err := range errs {
+		s[i] = err.Error()
+	}
+	return s
+}
+
+func cacheCheck(ctx context.Context) []error {
+	errors := make([]error, 0)
```
muhamadazmy:
```suggestion
	var errors []error
```
This is better because it does not allocate unless needed

pkg/perf/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,116 @@
...
+		err := zui.PushErrors(ctx, label, data)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return h.errors, nil
+}
+
+func errorsToStrings(errs []error) []string {
+	s := make([]string, len(errs))
```
muhamadazmy: if you really wanna be more efficent u do it like this 
```suggestion
	s := make([]string, 0, len(errs)) // set the capacity instead
	for _, err := range errs {
		s = append(s, err.Error())
	}
	return s
```

muhamadazmy: I intitlize the array with exact capacity as the len of th error list, then just use append. it's more go style than than using indexies

pkg/zui/zui.go
```diff
@@ -0,0 +1,90 @@
...
+
+func (m *module) renderErrors(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(2 * time.Second):
+			m.mu.Lock()
+			labels := make([]labelData, len(m.labels))
+			copy(labels, m.labels)
```
muhamadazmy, line 67: Are we sure copy requires destination to have the same size?

muhamadazmy: nevermind, it is required

muhamadazmy (approved):

Issue #2142, opened 2023-11-28
CPU benchmark performance test results needs docs


khaledyoussef24:
## Describe the bug

CPU benchmark performance test results needs docs because it is not clear

#2112

after reading https://github.com/threefoldtech/cpu-benchmark-simple/issues/1#issuecomment-1820051375
it become more clear so if it is added to docs results will be clear.

khaledyoussef24:
verified now an explanation for the results is provided in `docs/tasks/cpubench.md `
![image](https://github.com/threefoldtech/zos/assets/110984055/fcb5d106-9a5c-48ae-a537-986f6aeaf477)

testlodge:
https://app.testlodge.com/a/26076/projects/40893/suites/234919

Issue #2139, opened 2023-11-27
iperf test download speed giving 0


khaledyoussef24:
## Describe the bug

after running the iperf test on different nodes, it is always giving me 0 download speed
```
  "download_speed": 0,
  "error": "the server is busy running a test. try again later",`## To Reproduce

```

muhamadazmy:
I think the problem might be that using cron, all tests end up running at exactly the same time. I thought this might affect the test results but not that.

The solution might be that we introduce some jitter in the scheduler.

muhamadazmy:
I think the best way to do this is for the Task interface to provide a Jitter function that returns Seconds which is the max number of seconds the job will sleep before actual execution 

The code execution will then `sleep rand(0, task.Jitter())` before actually running the task.
A Jitter value of 0 should not introduce any sleep

A-Harby:
I just ran it again, and it gave the same error: `The server is busy running a test. try again later`.

Devnet, Node TwinID is 21.
![image](https://github.com/threefoldtech/zos/assets/109473873/e82843ae-966e-4c94-886f-be4937a180ca)

muhamadazmy:
Okay, seems the jitter did not have the right effect (although i am sure it helped) but seems we have 2 issues here that need to be worked on:

- Iperf server that we host on public node need to support multiple (but still limited) number of concurrent tests. I think this should supported by the server arguments (please check manual and let's modify options)
- The second thing is that for each test of these if we get this exact error, the client need to retry (with some exponential waits and jitter) for maximum of may be 10 retires before it give up and move to the next public node)

Omarabdul3ziz:
the mentioned pr does some retries with backoffs in case the server is busy.

exploring the possibility of enabling the iperf3 server to handle multiple connections found the following:
- Iperf3 is designed as a single-threaded application. 
  there were suggestions on the repository to enable the server to accept multiple requests per connection but weren't approved by the dev team.
- one of the proposals suggested the following approach:
  the server would handle the initial request. subsequent requests would fork the iperf server process on different ports.
  we can do something similar by running multiple instances of the server on different ports and randomly selecting one until an available port can handle the request?
- also there were some comments that the single-threaded design is specific to iperf3, while iperf2 is multi-threaded. I will try to validate this first before we consider using iperf2 instead. is there any specific need to use iperf3?

Omarabdul3ziz:
- verified that `iperf2` can handle multiple clients (running both servers on the same machine)
[...]

PR #2149, opened 2023-12-04
add jitter for the task

Omarabdul3ziz:
### Description
- jitter is the maximum number of seconds a task can wait before actually running.
- remove the task struct fields and return the constant values directly from the methods
- update docs

### Related Issues
- https://github.com/threefoldtech/zos/issues/2139
- https://github.com/threefoldtech/zos/issues/2142

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/docs/tasks/cpubench.md
+++ b/docs/tasks/cpubench.md
@@ -8,6 +8,7 @@ The `CPUBenchmark` task is designed to measure the performance of the CPU. it ut
 
 - Name: `cpu-benchmark`
 - Schedule: 4 times a day
+- Jitter: 10 min
 
 ### Details
 
--- a/docs/tasks/cpubench.md
+++ b/docs/tasks/cpubench.md
@@ -8,7 +8,7 @@ The `CPUBenchmark` task is designed to measure the performance of the CPU. it ut
 
 - Name: `cpu-benchmark`
 - Schedule: 4 times a day
-- Jitter: 10 min
+- Jitter: 0
 
 ### Details
 
@@ -17,6 +17,7 @@ The `CPUBenchmark` task is designed to measure the performance of the CPU. it ut
 - Lower time = better performance: for a single threaded benchmark, a lower execution time indicates better performance.
 
 ### Result sample
+
 ```json
 {
   "description": "Measures the performance of the node CPU by reporting the timespent of computing a task in seconds.",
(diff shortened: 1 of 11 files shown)


docs/tasks/iperf.md
```diff
@@ -8,6 +8,7 @@ The `iperf` package is designed to facilitate network performance testing using
 
 - Name: `iperf`
 - Schedule: 4 times a day
+- Jitter: 10 min
```
muhamadazmy: i will increase any network tests to 20 min jitter

docs/tasks/cpubench.md
```diff
@@ -8,6 +8,7 @@ The `CPUBenchmark` task is designed to measure the performance of the CPU. it ut
 
 - Name: `cpu-benchmark`
 - Schedule: 4 times a day
+- Jitter: 10 min
```
muhamadazmy: this runs completely local so this can be 0 actually

pkg/perf/publicip/publicip_task.go
```diff
@@ -62,6 +63,7 @@ func NewTask() perf.Task {
 		taskID:      taskID,
 		schedule:    taskSchedule,
 		description: taskDescription,
+		jitter:      10 * 60,
```
muhamadazmy: Why this has to be defined as an attribute? you can always return this in the jitter function itself.

pkg/perf/publicip/publicip_task.go
```diff
@@ -77,8 +79,11 @@ func (p *publicIPValidationTask) Description() string {
 	return p.description
 }
 
-func (p *publicIPValidationTask) Run(ctx context.Context) (interface{}, error) {
+func (p *publicIPValidationTask) Jitter() uint32 {
+	return p.jitter
```
muhamadazmy:
```suggestion
	return 10 * 60 //10 minutes
```

pkg/perf/publicip/publicip_task.go
```diff
@@ -62,6 +63,7 @@ func NewTask() perf.Task {
 		taskID:      taskID,
 		schedule:    taskSchedule,
 		description: taskDescription,
```
muhamadazmy: i don't get what is the point of defining constants, then basically __copy__ them to the struct. The functions (say Description()) can instead return the constant directly without the need to store (copy) this to the actual task object.

muhamadazmy (changes requested):

muhamadazmy (approved):

Issue #2153, opened 2023-12-06
introduce diagnostics call


xmonader:
Introducing the `zos.system.diagnostics` Command

## Motivation

Currently, we depend on a centralized mechanism to gather detailed logs/monitoring the system and present the health status of system components. which probably not ideal to query or to develop toolings for while we could just get the information from the node  in a brief way 

## Proposed Solution

Introduce the `zos.system.diagnostics` command to provide a consolidated view of system component health. This command should retrieve and display the status of each component, including:
```
{
  systemok: true/false,
  components: 
    {
    	 name: {name, status, error}, .. 
    
    }
}
```

- **SystemOK:** A global boolean flag indicating whether the whole system is considered healthy or not. (if the system is unhealthy no uptime reports should be sent)
- **Component Name:** The identifier for the system component.
- **Status:** The current operational state of the component, such as "OK", "NO"
- **Error:** A string containing the error message that should be communicated

> Note: feel free to remove the repetition of the component name in the value record

This information would enable clients to quickly assess the overall system health, make decisions to include/exclude that node from their planning, notifying the user in the dashboard or even adding more visualization to the node details

- **Proactive Issue Detection:** Early identification of potential system issues, facilitating timely intervention and preventing system outages.
- **Enhanced Troubleshooting:** Simplified troubleshooting process by providing a centralized source of system component status information.
- **Improved System Stability:** Reduced downtime and improved system reliability due to proactive issue detection and resolution (when probably monitored)
- **Uptime Reports:** It shouldn't send uptime in cases where SystemOK is false

muhamadazmy:
Many don't know that for each zbus server there is a built in "Status" call that can:
- Get indication if service is running (if it answers or not)
- Returns number of workers 
- Return state of the worker (free, busy, etc...)

zos already has a built in command line tool `zbusdebug` that prints the status of all services and their helath. 

it should be failry easy to implement an rmb handler that can basically run the same code and return diangoses of the services.

It can also may be expanded later to include other checks as well

Omarabdul3ziz:
update: https://github.com/threefoldtech/zos/pull/2160 in review

PR #2160, opened 2023-12-09
add rmb diagnostics call

Omarabdul3ziz:
### Description
- add new rmb call `zos.system.diagnostics` returns the output of `zbusdebug` command
- add new perf task to check the node reaches for grid services

### Related Issues
- https://github.com/threefoldtech/zos/issues/2153
- https://github.com/threefoldtech/zos/issues/2163

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/client/node.go
+++ b/client/node.go
@@ -77,6 +77,7 @@ import (
 	"net"
 
 	"github.com/threefoldtech/tfgrid-sdk-go/rmb-sdk-go"
+	"github.com/threefoldtech/zbus"
 	"github.com/threefoldtech/zos/pkg"
 	"github.com/threefoldtech/zos/pkg/capacity/dmi"
 	"github.com/threefoldtech/zos/pkg/gridtypes"
@@ -96,6 +97,19 @@ type Version struct {
 type Interface struct {
 	IPs []string `json:"ips"`
 	Mac string   `json:"mac"`
+}
+
+type moduleStatus struct {
+	Status zbus.Status `json:"status,omitempty"`
+	Err    error       `json:"error,omitempty"`
+}
+
+// Diagnostics show the health of zbus modules
+type Diagnostics struct {
+	// SystemStatusOk is the overall system status
+	SystemStatusOk bool `json:"system_status_ok"`
+	// Modules is a list of modules with their objects and workers
+	Modules map[string]moduleStatus `json:"modules"`
 }
 
 type ExitDevice struct {
@@ -325,5 +339,11 @@ func (n *NodeClient) SystemHypervisor(ctx context.Context) (result string, err e
 		return
 	}
 
+	return
+}
+
+func (n *NodeClient) SystemDiagnostics(ctx context.Context) (result Diagnostics, err error) {
+	const cmd = "zos.system.diagnostics"
+	err = n.bus.Call(ctx, n.nodeTwin, cmd, nil, &result)
 	return
 }
--- a/client/node.go
+++ b/client/node.go
@@ -77,9 +77,9 @@ import (
 	"net"
 
 	"github.com/threefoldtech/tfgrid-sdk-go/rmb-sdk-go"
-	"github.com/threefoldtech/zbus"
 	"github.com/threefoldtech/zos/pkg"
 	"github.com/threefoldtech/zos/pkg/capacity/dmi"
+	"github.com/threefoldtech/zos/pkg/diagnostics"
 	"github.com/threefoldtech/zos/pkg/gridtypes"
 )
 
@@ -97,19 +97,6 @@ type Version struct {
 type Interface struct {
 	IPs []string `json:"ips"`
 	Mac string   `json:"mac"`
-}
-
-type moduleStatus struct {
-	Status zbus.Status `json:"status,omitempty"`
-	Err    error       `json:"error,omitempty"`
-}
-
-// Diagnostics show the health of zbus modules
-type Diagnostics struct {
-	// SystemStatusOk is the overall system status
-	SystemStatusOk bool `json:"system_status_ok"`
-	// Modules is a list of modules with their objects and workers
-	Modules map[string]moduleStatus `json:"modules"`
 }
 
 type ExitDevice struct {
@@ -342,7 +329,7 @@ func (n *NodeClient) SystemHypervisor(ctx context.Context) (result string, err e
 	return
 }
 
-func (n *NodeClient) SystemDiagnostics(ctx context.Context) (result Diagnostics, err error) {
+func (n *NodeClient) SystemDiagnostics(ctx context.Context) (result diagnostics.Diagnostics, err error) {
 	const cmd = "zos.system.diagnostics"
 	err = n.bus.Call(ctx, n.nodeTwin, cmd, nil, &result)
 	return
--- a/cmds/modules/noded/main.go
+++ b/cmds/modules/noded/main.go
@@ -193,8 +193,12 @@ func action(cli *cli.Context) error {
 		return version, nil
 	})
 
+	diagnosticsMgr, err := diagnostics.NewDiagnosticsManager(msgBrokerCon, redis)
+	if err != nil {
+		return errors.Wrap(err, "failed to create a new Diagnostics manager")
+	}
 	bus.WithHandler("zos.system.diagnostics", func(ctx context.Context, payload []byte) (interface{}, error) {
-		return diagnostics.GetSystemDiagnostics(ctx, redis, msgBrokerCon)
+		return diagnosticsMgr.GetSystemDiagnostics(ctx)
 	})
 
 	bus.WithHandler("zos.system.dmi", func(ctx context.Context, payload []byte) (interface{}, error) {
--- a/cmds/modules/noded/main.go
+++ b/cmds/modules/noded/main.go
@@ -15,9 +15,9 @@ import (
 
 	substrate "github.com/threefoldtech/tfchain/clients/tfchain-client-go"
 	"github.com/threefoldtech/tfgrid-sdk-go/rmb-sdk-go"
-	"github.com/threefoldtech/zos/cmds/modules/zbusdebug"
 	"github.com/threefoldtech/zos/pkg/app"
 	"github.com/threefoldtech/zos/pkg/capacity"
+	"github.com/threefoldtech/zos/pkg/diagnostics"
 	"github.com/threefoldtech/zos/pkg/environment"
 	"github.com/threefoldtech/zos/pkg/events"
 	"github.com/threefoldtech/zos/pkg/monitord"
@@ -193,36 +193,7 @@ func action(cli *cli.Context) error {
 	})
 
 	bus.WithHandler("zos.system.diagnostics", func(ctx context.Context, payload []byte) (interface{}, error) {
-		type moduleStatus struct {
-			Status zbus.Status `json:"status,omitempty"`
-			Err    error       `json:"error,omitempty"`
-		}
-		results := struct {
-			SystemStatusOk bool                    `json:"system_status_ok"`
-			Modules        map[string]moduleStatus `json:"modules"`
-		}{
-			SystemStatusOk: true,
-			Modules:        make(map[string]moduleStatus),
-		}
-
-		for module := range zbusdebug.PossibleModules {
-			ctx, cancel := context.WithTimeout(ctx, 3*time.Second)
-
-			status, err := redis.Status(ctx, module)
-			if err != nil {
-				results.SystemStatusOk = false
-			}
-
-			moduleStatus := moduleStatus{
-				Status: status,
-				Err:    err,
-			}
-			results.Modules[module] = moduleStatus
-
-			cancel()
-		}
-
-		return results, nil
+		return diagnostics.GetSystemDiagnostics(ctx, redis)
 	})
 
 	bus.WithHandler("zos.system.dmi", func(ctx context.Context, payload []byte) (interface{}, error) {
--- a/cmds/modules/noded/main.go
+++ b/cmds/modules/noded/main.go
@@ -206,23 +206,20 @@ func action(cli *cli.Context) error {
 		}
 
 		for module := range zbusdebug.PossibleModules {
+			ctx, cancel := context.WithTimeout(ctx, 3*time.Second)
+
 			status, err := redis.Status(ctx, module)
+			if err != nil {
+				results.SystemStatusOk = false
+			}
+
 			moduleStatus := moduleStatus{
 				Status: status,
 				Err:    err,
 			}
-
-			if err != nil {
-				results.SystemStatusOk = false
-			} else {
-				for _, worker := range status.Workers {
-					if worker.State != zbus.WorkerFree {
-						results.SystemStatusOk = false
-					}
-				}
-			}
-
 			results.Modules[module] = moduleStatus
+
+			cancel()
 		}
 
 		return results, nil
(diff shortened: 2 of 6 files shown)


cmds/modules/noded/main.go
```diff
@@ -40,6 +41,11 @@ const (
 	eventsBlock     = "/tmp/events.chain"
 )
 
+type Diagnostics struct {
```
xmonader: Not the right place, should be in client/node.go (next to zos.statistics.get). cmds is only for commands (binaries)

cmds/modules/noded/main.go
```diff
@@ -191,6 +197,28 @@ func action(cli *cli.Context) error {
...
+			Modules:        make(map[string]zbus.Status),
+		}
+
+		for module := range zbusdebug.PossibleModules {
+			status, err := redis.Status(ctx, module)
+			if err != nil {
+				return nil, errors.Wrapf(err, "failed to get status of module %s", module)
+			}
+			for _, worker := range status.Workers {
+				if worker.State != "free" {
```
xmonader: Should compare against `WorkerFree` instead

cmds/modules/noded/main.go
```diff
@@ -191,6 +197,28 @@ func action(cli *cli.Context) error {
...
+					results.SystemStatusOk = false
```
xmonader: I'm not sure if the comparison against WorkerFree by itself is enough, i think the check should be in status, err := ... and then do if err -> systemOk is false instead (you shouldn't return there)

cmds/modules/noded/main.go
```diff
@@ -191,6 +197,28 @@ func action(cli *cli.Context) error {
...
 
+	bus.WithHandler("zos.system.diagnostics", func(ctx context.Context, payload []byte) (interface{}, error) {
+		results := Diagnostics{
+			SystemStatusOk: true,
+			Modules:        make(map[string]zbus.Status),
+		}
+
+		for module := range zbusdebug.PossibleModules {
+			status, err := redis.Status(ctx, module)
+			if err != nil {
```
xmonader: this should flag the systemok as false and not return ...

cmds/modules/noded/main.go
```diff
@@ -191,6 +192,42 @@ func action(cli *cli.Context) error {
...
+		for module := range zbusdebug.PossibleModules {
+			status, err := redis.Status(ctx, module)
+			moduleStatus := moduleStatus{
+				Status: status,
+				Err:    err,
+			}
+
+			if err != nil {
+				results.SystemStatusOk = false
+			} else {
```
xmonader: no need to check on the workers' status, I double checked with Azmy, being in Free/Busy state doesn't indicate any error

cmds/modules/noded/main.go
```diff
@@ -191,6 +192,42 @@ func action(cli *cli.Context) error {
...
+		results := struct {
+			SystemStatusOk bool                    `json:"system_status_ok"`
+			Modules        map[string]moduleStatus `json:"modules"`
+		}{
+			SystemStatusOk: true,
+			Modules:        make(map[string]moduleStatus),
+		}
+
+		for module := range zbusdebug.PossibleModules {
+			status, err := redis.Status(ctx, module)
```
xmonader: Please make sure the context used is WithTimeout maybe 2 3 seconds

cmds/modules/noded/main.go
```diff
@@ -191,6 +192,39 @@ func action(cli *cli.Context) error {
...
+		}
+		results := struct {
+			SystemStatusOk bool                    `json:"system_status_ok"`
+			Modules        map[string]moduleStatus `json:"modules"`
+		}{
+			SystemStatusOk: true,
+			Modules:        make(map[string]moduleStatus),
+		}
+
+		for module := range zbusdebug.PossibleModules {
```
xmonader: Don't depend on zbusdebug, Copy the possible modules into this file or this function please

cmds/modules/noded/main.go
```diff
@@ -191,6 +192,39 @@ func action(cli *cli.Context) error {
...
+				results.SystemStatusOk = false
+			}
+
+			moduleStatus := moduleStatus{
+				Status: status,
+				Err:    err,
+			}
+			results.Modules[module] = moduleStatus
+
+			cancel()
```
xmonader: cancel should have been deferred, directly after making that context if needed

cmds/modules/noded/main.go
xmonader: Any reason not to do these module status checks concurrently?

Omarabdul3ziz: just thought having them in one place would be easier if gonna change later. but ok can copy them here

no, not a specific reason but the call takes no time.

xmonader: Worst case scenario if each take 3 seconds to reply and all of them don't and we do sequentially that maybe 3 sec * 7 modules = 21 seconds

cmds/modules/noded/main.go
```diff
@@ -191,6 +192,39 @@ func action(cli *cli.Context) error {
 		return version, nil
 	})
 
+	bus.WithHandler("zos.system.diagnostics", func(ctx context.Context, payload []byte) (interface{}, error) {
```
muhamadazmy, line 200: i don't like that the implementation is like this in the main file. The WithHandler can still take a function that is at least defined separately as it's own function (not anonymous) and preferably in a separate file

cmds/modules/noded/main.go
muhamadazmy: You definitely should run all the checks at once (make the calls async) and reply on a single response queue or using a waitgroup

muhamadazmy (changes requested):

pkg/diagnostics/diagnostics.go
```diff
@@ -0,0 +1,121 @@
...
+type Diagnostics struct {
+	// SystemStatusOk is the overall system status
+	SystemStatusOk bool `json:"system_status_ok"`
+	// ZosModules is a list of modules with their objects and workers
+	ZosModules map[string]moduleStatus `json:"modules"`
+	// Online is the state of the grid services reachable from the node
+	Online bool `json:"online"`
+}
+
+func GetSystemDiagnostics(ctx context.Context, msgBrokerCon string) (Diagnostics, error) {
```
muhamadazmy: it's really bad idea to pass around the msgBrokerCon string instead of the zbus.Client itself. the zbus.Client can be used concurrently by multiple stubs with no issue. Creating a Client creates a pool that is kept alive forever. Hence  creating a client each time means you create a new pool each time u call this function!

pkg/diagnostics/diagnostics.go
```diff
@@ -0,0 +1,121 @@
...
+	defer cancel()
+
+	status, err := busClient.Status(ctx, module)
+	return moduleStatus{
+		Status: status,
+		Err:    err,
+	}
+}
+
+func isOnline(ctx context.Context, msgBrokerCon string) bool {
```
muhamadazmy: What does this do ?

pkg/diagnostics/diagnostics.go
```diff
@@ -0,0 +1,121 @@
...
+	client, err := utils.NewRedisPool(msgBrokerCon)
+	if err != nil {
+		return false
+	}
+
+	conn := client.Get()
+	defer conn.Close()
```
muhamadazmy, line 113: you closing the connection but not the pool. The idea of a `Pool` is the reusability so creating a pool ONLY to use one connection then through it away does not sound like a good way to use a pool

Instead Pools are created to live long, and used repeatedly to retrieve (and reuse) connections. So at least here you need to close the pool after you re done .

pkg/diagnostics/diagnostics.go
```diff
@@ -0,0 +1,121 @@
...
+
+	data, err := conn.Do("GET", testNetworkKey)
```
muhamadazmy, line 115: Where is that key coming from ? and what are we trying to do here.

pkg/perf/networkhealth/networkhealth.go
```diff
@@ -0,0 +1,112 @@
...
+func parseUrl(serviceUrl string) string {
+	u, err := url.Parse(serviceUrl)
+	if err != nil {
+		return ""
+	}
+	host := u.Host
+
+	if !strings.Contains(host, ":") {
+		host = fmt.Sprintf("%s:80", host)
+	}
```
muhamadazmy: The default port should be based on the url `scheme` 80 if it's http, and 443 if it's https

pkg/perf/networkhealth/networkhealth.go
```diff
@@ -0,0 +1,112 @@
...
+
+	return host
+}
+
+func isReachable(ctx context.Context, serviceUrl string) error {
```
muhamadazmy:
```suggestion
func isReachable(ctx context.Context, address string) error {
```

you call it serviceUrl while in fact it's an address of format `<host>:<port>`

muhamadazmy (changes requested):

pkg/diagnostics/diagnostics.go
Omarabdul3ziz: we needed to add an extra check on the diagnosis call to check the node connectivity to the grid services.
so i added a new perf task `network-health` that check services (chain, relay, graphql, ...) 
`isOnline` check if all the services `is_reachable` by getting the perf task result from the redis cache.

pkg/perf/networkhealth/networkhealth.go
Omarabdul3ziz: this was just a workaround for the validation on `net.Dialer.DialContext`.
the services url we call are a mix of different schemes (https/wss/redis) but when i directly use the address on the dial it fails `dial tcp: address tfchain.grid.tf: missing port in address`.
so i am just adding the port 80 for all the address

pkg/diagnostics/diagnostics.go
Omarabdul3ziz: a constant perf task `perf.network-health`

client/node.go
```diff
@@ -98,6 +99,19 @@ type Interface struct {
...
+	Err    error       `json:"error,omitempty"`
+}
+
+// Diagnostics show the health of zbus modules
+type Diagnostics struct {
+	// SystemStatusOk is the overall system status
+	SystemStatusOk bool `json:"system_status_ok"`
+	// Modules is a list of modules with their objects and workers
+	Modules map[string]moduleStatus `json:"modules"`
+}
```
xmonader: Why isn't this exported?

pkg/perf/networkhealth/networkhealth.go
```diff
@@ -0,0 +1,112 @@
...
+	"net/url"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/threefoldtech/zos/pkg/environment"
+	"github.com/threefoldtech/zos/pkg/perf"
+)
+
+const requestTimeout = 5 * time.Second
```
xmonader: if that at any point will be parametrized, make it `defaultRequestTimeout`

pkg/perf/networkhealth/networkhealth.go
```diff
@@ -0,0 +1,112 @@
...
+func (t *NetworkHealthTask) Jitter() uint32 {
+	return 0
+}
+
+func (t *NetworkHealthTask) Run(ctx context.Context) (interface{}, error) {
+	env := environment.MustGet()
+	servicesUrl := []string{
+		env.ActivationURL, env.GraphQL, env.FlistURL,
+	}
+	servicesUrl = append(servicesUrl, env.SubstrateURL...)
```
xmonader: can't these two lines be joined?

pkg/utils/redis.go
```diff
@@ -8,42 +8,58 @@ import (
...
-	} else if len(size) > 1 {
-		panic("invalid pool size")
-	}
+type RedisConfig struct {
+	Scheme  string
+	Host    string
+	Options []redis.DialOption
+}
 
+func parseRedisAdd(address string) (RedisConfig, error) {
```
xmonader: Rename to parseRedisAddress, why did you abbreviate that?

pkg/utils/redis.go
```diff
@@ -8,42 +8,58 @@ import (
...
 )
 
-func NewRedisPool(address string, size ...uint32) (*redis.Pool, error) {
-	var poolSize uint32 = 20
-	if len(size) == 1 {
-		poolSize = size[0]
-	} else if len(size) > 1 {
-		panic("invalid pool size")
-	}
+type RedisConfig struct {
```
xmonader: This should be RedisDialParams

pkg/utils/redis.go
xmonader: Also add tests for that

client/node.go
```diff
@@ -98,6 +99,19 @@ type Interface struct {
 	Mac string   `json:"mac"`
 }
 
+type moduleStatus struct {
```
xmonader: Export it

pkg/diagnostics/diagnostics.go
```diff
@@ -0,0 +1,113 @@
...
+	"vmd",
+	"flist",
+	"network",
+	"container",
+	"provision",
+	"gateway",
+	"qsfsd",
+}
+
+type moduleStatus struct {
```
xmonader: Why isn't that exported and reused in client/node.go?

pkg/perf/networkhealth/networkhealth.go
muhamadazmy: that's what i am saying. Dial because it's a tcp dial needs to know an ip and port. You can't assume that ALL services listening on port 80.  because a single IP can has multiple ports listening (multiple services).

For example say the hub. The hub has a web page that is available on port 80 and 443 (http and https) but also has zdb on port 9900. So if u assume port 80 for all your tests (if the port is not in the host name) then u will only be testing the http port.

Instead this function has to use the default port of the services. 

Luckily links with custom ports already have the port in the URL but your code has to be correct, and to be correct you need to use the default port **of the protocol** when guessing the port number

pkg/perf/networkhealth/networkhealth.go
```diff
@@ -0,0 +1,111 @@
...
+func (t *NetworkHealthTask) Description() string {
+	return "Network health check runs periodically to check the connection to most of grid services."
+}
+
+func (t *NetworkHealthTask) Cron() string {
+	return "0 */5 * * * *"
+}
+
+func (t *NetworkHealthTask) Jitter() uint32 {
+	return 0
```
muhamadazmy: we need to add jitter here or it means ALL nodes in the world will try to test the services at exactly the same time, DDoS(ing) our own services!

pkg/perf/networkhealth/networkhealth.go
```diff
@@ -0,0 +1,111 @@
...
+func (t *NetworkHealthTask) ID() string {
+	return "network-health"
+}
+
+func (t *NetworkHealthTask) Description() string {
+	return "Network health check runs periodically to check the connection to most of grid services."
+}
+
+func (t *NetworkHealthTask) Cron() string {
+	return "0 */5 * * * *"
```
muhamadazmy: is 5 min a requirement, may be each 15 min is better

pkg/perf/networkhealth/networkhealth.go
muhamadazmy: I think this should be the number of seconds (check docs) but if we gonna do a test every 15 min the jitter could be around 5 min or so

muhamadazmy (changes requested):

muhamadazmy (approved):

Issue #2146, opened 2023-11-30
Deployment changes not reported when workload description or metadata is changed


scottyeager:
I noticed some unexpected behavior while testing workload updates. The flow is like this:

1. Create a deployment
2. Update the deployment with a change to [metadata or description](https://github.com/threefoldtech/zos/blob/61b7050c709f72a8b1851b282c6fcd937b5a14b4/pkg/gridtypes/workload.go#L135) on some workload
3. Zos accepts without error and executes the change
4. But, when calling `zos.deployment.changes` the new version with the changes is not included in the reply

## Context

This issue came up while attempting to update deployments created in the Playground using `tfgrid-sdk-go` code. Since metadata [isn't supported](https://github.com/threefoldtech/tfgrid-sdk-go/issues/521) in the sdk, it is stripped from deployments converted between Zos and sdk abstractions for workloads, if it exists. Playground attaches metadata to workloads it creates, on the other hand, so an update to remove that metadata is an unavoidable side effect under current implementations.

## Steps to reproduce

While some custom code is needed to see the metadata case, since we can't update deployment in the Playground and we can't set metadata using Terraform, we can reproduce the case of descriptions in a simple way using Terraform.

For brevity, I'll just show the relevant deployment blocks (assume the network is provided). First start with a VM and disk:

```
resource "grid_deployment" "d1" {
  node = 2
  network_name = grid_network.net.name

  disks {
    name = "mydisk1"
    size = 15
    description = ""

  }

  vms {
    name = "vm1"
    flist = "https://hub.grid.tf/tf-official-vms/ubuntu-22.04.flist"
    cpu = 1
    memory = 1024
    entrypoint = "/sbin/zinit init"
    planetary = true
    mounts {
      disk_name = "mydisk1"
      mount_point = "/"
    }
    env_vars = {
      SSH_KEY = file("~/.ssh/id_rsa.pub")
    }
  }
}
```
```
terraform apply
```

Next we're going to put the disk into a detached state by removing the VM from the deployment. To see that Zos is actually destroying the VM, we can check the Zos logs or try some ongoing network communication with the VM like a running ping. Since the deployment will be reverted to it's original state, it's necessary to check while the next apply is happening.

Note that we add a description to the disk:
```
resource "grid_deployment" "d1" {
  node = 2
  network_name = grid_network.net.name

  disks {
    name = "mydisk1"
    size = 15
    description = "test" 
  }
}
```
```
terraform apply
```

[...]

PR #2161, opened 2023-12-10
Update workloads on just version updates

AbdelrahmanElawady:
### Description

Change workloads to be updated on just version changes.

### Related Issues

- #2146 

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/pkg/provision/engine.go
+++ b/pkg/provision/engine.go
@@ -779,14 +779,18 @@ func (e *NativeEngine) updateWorkload(ctx context.Context, wl *gridtypes.Workloa
 	if e.provisioner.CanUpdate(ctx, wl.Type) {
 		result, err = e.provisioner.Update(ctx, wl)
 	} else {
-		// deprecated. We should never update resources by decomission and then provsion
+		// deprecated. We should never update resources by decommission and then provision
 		// the check in Update method should prevent this
 		// #unreachable
 		err = fmt.Errorf("can not update this workload type")
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/provision/provisiner.go
```diff
@@ -209,9 +209,6 @@ func (p *mapProvisioner) Update(ctx context.Context, wl *gridtypes.WorkloadWithI
 	}
 
 	data, err := updater.Update(ctx, wl)
-	if errors.Is(err, ErrNoActionNeeded) {
-		return result, err
-	}
```
muhamadazmy: why did you remove this ?

pkg/provision/provisiner.go
```diff
@@ -231,7 +228,7 @@ func setState(result *gridtypes.Result, err error) {
 	state := gridtypes.StateOk
 	str := ""
 
-	if err != nil {
```
muhamadazmy: it's because you removed the code above that u have to change this.

muhamadazmy (changes requested):
I think you got few things misunderstood 

What we need to do if "no action needed" error is returned from only Update() request that what u need to do is  load the last result (last workload state) and then copy that result to the (new wl) object then save that and return.

Everything else you did in the code is un-necessary and actually breaks things. because now what happens is that the last status of the object can be empty. Because many of the provisioners will return an empty result if nothing need to be done.

muhamadazmy (approved):

Issue #2166, opened 2023-12-13
Improve flist cleanup


AbdelrahmanElawady:
## Description

Some mountpoints of old VMs are kept around in the node cache. For example:

![image](https://github.com/threefoldtech/zos/assets/60783742/941d8eba-da9c-4ccf-8ddf-c33a6a9202e6)

It would be better if we can cleanup this mountpoints instead of cluttering the cache dir

Eslam-Nawara:
WIP:
- added a lock on creating the mount point to prevent the cleaner from removing the mount point before the file system is created

Eslam-Nawara:
## Work Completed: 
- removed the `MountsCleaner` and changed the `cleanUnusedMounts` to be able to remove any folder that is not a mount point

PR #2173, opened 2023-12-19
Add mountpoint cleanup

AbdelrahmanElawady:
### Description

Cleanup mountpoints every 24hrs

### Related Issues

- #2166

### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation

### DIFF
--- a/pkg/flist/cleanup.go
+++ b/pkg/flist/cleanup.go
@@ -15,46 +15,12 @@ import (
 // Cleaner interface, implementer of this interface
 // can start a cleaner job
 type Cleaner interface {
-	// MountsCleaner runs the clean process, MountsCleaner should be
-	// blocking. Caller then can do `go MountsCleaner()` to run it in the background
-	MountsCleaner(ctx context.Context, every time.Duration)
 	// CacheCleaner runs the clean process, CacheCleaner should be
 	// blocking. Caller then can do `go CacheCleaner()` to run it in the background
 	CacheCleaner(ctx context.Context, every time.Duration, age time.Duration)
 }
 
 var _ Cleaner = (*flistModule)(nil)
-
-func (f *flistModule) MountsCleaner(ctx context.Context, every time.Duration) {
-	log := app.SampledLogger()
-
-	for {
-		select {
-		case <-ctx.Done():
-			return
-		case <-time.After(every):
-
-			entries, err := os.ReadDir(f.mountpoint)
-			if err != nil {
-				log.Error().Err(err).Msg("failed to read mountpoint directory")
-			}
-			for _, entry := range entries {
-				path := filepath.Join(f.mountpoint, entry.Name())
-				if err := f.isMountpoint(path); err == nil {
-					continue
-				}
-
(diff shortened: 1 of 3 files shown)


pkg/flist/cleanup.go
```diff
@@ -15,13 +16,44 @@ import (
...
+			}
+		}
+	}
+}
+
+func isMount(ctx context.Context, path string) bool {
+	err := exec.CommandContext(ctx, "mountpoint", path, "-q").Run()
+	return err == nil
+}
+
```
muhamadazmy: flistModule already has a `f.isMountpoint()` method !

pkg/flist/cleanup.go
```diff
@@ -15,13 +16,44 @@ import (
...
-	// blocking. Caller then can do `go MountsCleaner()` to run it in the background
+	MountsCleaner(ctx context.Context, every time.Duration)
+	// CacheCleaner runs the clean process, CacheCleaner should be
+	// blocking. Caller then can do `go CacheCleaner()` to run it in the background
 	CacheCleaner(ctx context.Context, every time.Duration, age time.Duration)
 }
 
 var _ Cleaner = (*flistModule)(nil)
 
+func (f *flistModule) MountsCleaner(ctx context.Context, every time.Duration) {
```
muhamadazmy: This is very unsafe procedures. 

this runs in the background in a continues loop. What prevents this from deleting a mountpoint that was just created, just before the filesystem is actually mounted. 

Hence this procedure can cause race conditions and errors that are impossible to debug.

muhamadazmy (changes requested):
After taking a look at this PR and review the flistd code I realised the following:
- The fix implemented has issues as explained in the comments below
- The flistd already has code that works in a safe way (where there is no way a mount/unmount are in progress) but seems that code is not working as expected. 
- Hence the code here should investigate and fix why the `cleanUnusedMounts` function is not doing what it needs to do

the `flistModule.cleanUnusedMounts` function is invoked after a mount and an unmount operation which make it run inside the zbus workers. this makes it safe because flistd only have one workers hence it's grated that there no other mount operations are in progress.

Please then review the code of this function and fix the issues  that cause the clean operation to not work as expected

muhamadazmy:
I think the issue with the function is that it does only clean up of `ro` mounts that are not used by any mountpoint but does not clean up un-used folders.

It can still go over all possible (as done in the PR) and delete any folder that is not a mount point.
It can also delete all empty ro mount locations that are not mount points

pkg/flist/cleanup.go
```diff
@@ -126,5 +126,22 @@ func (f *flistModule) cleanUnusedMounts() error {
...
+	}
+
+	for _, entry := range entries {
+		path := filepath.Join(f.mountpoint, entry.Name())
+		if err := f.isMountpoint(path); err == nil {
+			continue
+		}
+
+		if err := os.Remove(path); err != nil {
+			log.Error().Err(err).Msg("failed to clean mountpoint")
```
xmonader: include path in the error please

muhamadazmy (approved):

Issue #2177, opened 2023-12-24
Some deployed VMs can't reach the public network


AhmedHanafy725:
env: qa
nodeId: 2

It doesn't always happen, but sometimes the deployed vm has `/etc/resolv.conf` empty.

ramezsaeed:
we need to check node on qa net when started and the version of software running on it.
as a work around u can populate the /etc/resolve.conf manually and it should work.

muhamadazmy:
can you provide information about the flist used to deploy this VM.

AhmedHanafy725:
I was using Caprover flist
https://hub.grid.tf/tf-official-apps/tf-caprover-latest.flist

AbdelrahmanElawady:
It's an issue with cloud-container not continuing network configuration if an error happens which leads to empty `/etc/resolv.conf` file (among maybe other things).
https://github.com/threefoldtech/cloud-container/issues/20

muhamadazmy:
Reopened to be verified first

PR #2195, opened 2024-02-06
Update cloud-container to latest

AbdelrahmanElawady:
### Description

Update cloud-container to latest which includes the fix for resolv.conf file not being set when an error during network setup happens.
It was tested by duplicating the routes entries [here](https://github.com/threefoldtech/zos/blob/main/pkg/vm/manager.go#L227-L230) and it logged the errors and continued the rest of the configuration so resolv.conf had entries.
```
found device with mac: 
found device with mac: 6a:9d:df:5f:9d:e2
found device with mac: aa:84:90:f9:d1:a7
setting up (6a:9d:df:5f:9d:e2)
failed to set route 10.20.0.0/16 via 10.20.2.1 on interface 6a:9d:df:5f:9d:e2: file exists
failed to set route 100.64.0.0/16 via 10.20.2.1 on interface 6a:9d:df:5f:9d:e2: file exists
setting up (aa:84:90:f9:d1:a7)
failed to set route 200::/7 via 301:8afa:1b85:5b6e::1 on interface aa:84:90:f9:d1:a7: file exists
generating host ssh keys
PING 8.8.8.8 (8.8.8.8): 56 data bytes
64 bytes from 8.8.8.8: seq=0 ttl=114 time=43.415 ms
```

### Changes

- latest cloud-container flist will be used

### Related Issues

- #2177 

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/pkg/primitives/vm/vm.go
+++ b/pkg/primitives/vm/vm.go
@@ -19,7 +19,7 @@ import (
 )
 
 const (
-	cloudContainerFlist = "https://hub.grid.tf/tf-autobuilder/cloud-container-32b445e.flist"
+	cloudContainerFlist = "https://hub.grid.tf/tf-autobuilder/cloud-container-6fb64bb.flist"
 	cloudContainerName  = "cloud-container"
 )
 


pkg/primitives/vm/vm.go
```diff
@@ -19,7 +19,7 @@ import (
 )
 
 const (
-	cloudContainerFlist = "https://hub.grid.tf/tf-autobuilder/cloud-container-32b445e.flist"
+	cloudContainerFlist = "https://hub.grid.tf/tf-autobuilder/cloud-container-6fb64bb.flist"
```
xmonader, line 22: Why isn't there a latest link that we can use?

xmonader (approved):

pkg/primitives/vm/vm.go
muhamadazmy: This is to avoid overriding it by accident basically. Now we in completel control which version of zos (network) uses which version of cloud-container.

muhamadazmy (approved):

PR #2199, opened 2024-02-15
Assign ipv6 and gw6 to vm via config

muhamadazmy:
### Description

Instead of relying on slaac to give ipv6 to VMs now it's done explicitly via cidata (cloud init data) provided to the vm

This means cloud-container now also disable router advertisment on interfaces.

This also mean ipv6 is not provided unless requested in the workload. This was a problem before because it was conflicting with other configurable routes for ipv6

### Changes

- Use recent cloud-container that auto disable router advertisement on all nics
- Make sure that assigned IPv6 and gateway are passed to machine via cidata (cloud-init data)

### DIFF
--- a/pkg/network/ndmz/dualstack.go
+++ b/pkg/network/ndmz/dualstack.go
@@ -303,6 +303,39 @@ func (d *dmzImpl) GetIP(family int) ([]net.IPNet, error) {
 	return results, err
 }
 
+// Get gateway to given destination ip
+func (d *dmzImpl) GetDefaultGateway(destination net.IP) (net.IP, error) {
+	netns, err := namespace.GetByName(dmzNamespace)
+	if err != nil {
+		return nil, err
+	}
+
+	defer netns.Close()
+
+	var gw net.IP
+	err = netns.Do(func(_ ns.NetNS) error {
+		routes, err := netlink.RouteGet(destination)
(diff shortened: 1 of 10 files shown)


pkg/network/ndmz/dualstack.go
```diff
@@ -303,6 +303,39 @@ func (d *dmzImpl) GetIP(family int) ([]net.IPNet, error) {
...
+		if err != nil {
+			return err
+		}
+
+		for _, route := range routes {
+			if route.Gw != nil {
+				gw = route.Gw
+				return nil
+			}
+		}
```
AbdelrahmanElawady, line 327: should we check that the route returned here is the default one? or do we just need any route.

AbdelrahmanElawady (approved):
looks good, just a question

pkg/network/ndmz/dualstack.go
muhamadazmy: I think the default gw is the only one that will have `gw` set actually. I tried to check the Src also to be nil but seems that the RouterGet uses your local ip as a source which makes sense

muhamadazmy: I also use a public IP (as destination) so the only one is returned is a default gw with the betst metric

PR #2201, opened 2024-02-19
use mocking for boot testing

Eslam-Nawara:
### Description

create unit tests for boot pkg using mocking

### Related Issues

List of related issues

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


pkg/app/boot.go
```diff
@@ -1,24 +1,64 @@
...
+	fs         fileSystem
+}
+
+type fileSystem interface {
+	Create(string) (io.ReadCloser, error)
+	MkdirAll(string, uint32) error
+	Stat(string) (any, error)
+}
+
+type defaultBootManager struct{}
```
muhamadazmy:
```suggestion
type defaultFileSystem struct{}
```

pkg/app/boot.go
```diff
@@ -1,24 +1,64 @@
...
+	}
+}
 
 // MarkBooted creates a file in a memory
 // this file then can be used to check if "something" has been restared
 // if its the first time it starts
 func MarkBooted(name string) error {
-	if err := os.MkdirAll(bootedPath, 0770); err != nil {
+	manager := defaultMarkManager()
+	return manager.markBooted(name)
```
muhamadazmy: No need to define a manager imho. All what u need to pass is, name, path, and something that implements a `fileSystem`

```suggestion
	return markBooted(name, bootedPath, &defaultFileSystem{})
```

in fact you don't even need to do `&defaultFileSystem{}` on each call this can be a global hidden var because it's mostly what you gonna use

```go 
type defaultFileSystem struct{}

var (
   defaultFS = &defaultFileSystem{}
)

// then in your call
func MarkBooted(name string) {
   return   markBooted(name, path, defaultFS)
}
```

pkg/app/boot.go
```diff
@@ -1,24 +1,64 @@
...
+
+func (dfs defaultBootManager) MkdirAll(path string, perm uint32) error {
+	return os.MkdirAll(path, fs.FileMode(perm))
+}
+
+func (dfs defaultBootManager) Stat(path string) (any, error) {
+	return os.Stat(path)
+}
+
+func defaultMarkManager() *bootManager {
```
muhamadazmy: this complicates this code unecessary imho. You don't need a "manager" and then have to create an instance of it every time.

Instead do as per other comments

muhamadazmy (changes requested):
The general approach does not match what we discusses.

it should be as follows:
- Public function with minimal arguments
- Hidden(private) functions with more arguments
- Public functions should (with no allocation per call) call the private ones with defaults (almost constant) variables.
- Tests then can make calls to private functions with mocks

pkg/cache/cache.go
```diff
@@ -11,19 +12,54 @@ const (
 	Megabyte = 1024 * 1024
 )
 
+// instance of default filesystem
+var defaultFS = &defaultFileSystem{}
```
muhamadazmy: I understand why you have this here again. But it doesn't look right to have the same code repeated in almost every package. May be a better approach is to define both the interface for Os and it's default implementation under `pkg`  directly.

Then it can be reused by multiple pkg directly

```yaml
- pkg:
   -  os.go
```

I think the interface should better have an different name that is more generic say `SystemOS` since most of the functions are just wrapper around the `os` standard package.

And `DefaultSystemOS` then can be the pointer to the default implementation

muhamadazmy: You can also provide a `os_test.go` file under pkg that has a public implementation of the mock so you can easily reuse it in all your test code

muhamadazmy: We can has another interface `SystemCall` also for the syscall and so on. What do you think?

pkg/container/logger/logger.go
```diff
@@ -29,28 +30,68 @@ type LogsRedis struct {
 	Stderr string `json:"stderr"`
 }
 
+type marshaler interface {
```
muhamadazmy: I can see you have good intentions to cover everything with tests but this can very easily go out of hand. 

If you think about it this test make the code unmanageable after a while and you have to choose what need to be tested and what not.

For example if you think about the value of this test it's very very small because there is not actual logic here to test. 

The Serialize function all what it does is it serialize using a standard json serializer and then write the content to a file using a write file from the standard library. 

So what does the test cover? the test code is already more than the logic because there is no logic in this function, no? and we definitely don't want to test json serializer nor the os.WriteFile functions.

Please rever this file and let's discuss what can be (and should) be tested and what's not.

Also please skip the entire `pkg/container` because it's not supported for workloads anymore it's only here because we use it internally to run zdb.

muhamadazmy (changes requested):

muhamadazmy (approved):

Issue #2200, opened 2024-02-19
unified API/Chain interface


muhamadazmy:
This is an improvement to zos design that should be done to:
- Make it easier to follow and debug code
  - Right now, zos establishes connections to tfchain when needed from some daemons to make requests.
  - ZOS daemons register their own rmb handlers so its always hard to figure out which service handles which rmb function because they all rely on the stand alone rmb-peer process.
- reduce the number of open connections to tfchain to one.
- Better handling for rmb requests (handling unknown requests, avoid relying on redis for rmb)

The idea is that we need to run ONE daemon that works as the gateway for zos connections to external services (chain and rmb) that then have the following:
- It itself exposes zbus functions that can be called from other services. (say a tfchain request) then it does the call on behalf of that daeomn. This mean that ALL chain requests must go through this API gateway daemon. It means also we can maintain a single connection to chain.
- It means also that only this daemon will need to access the node "identity" instead of granting all services access to it.
- The second most important is to also have a unified (rmb) API via this single services (using the rmb go client)  all rmb requests then can be:
  - Verified that the call is a valid call (function actually exists, and return proper error if calling an undefined functino)
  - Verify inputs, etc..
  - Then call the actual services over zbus stubs. Zbus is easier to follow than rmb handlers.
  - This means ALL the api is then available over a single services, then can be later easily replaced with an HTTP api or RPC (for zos4)

PR #2202, opened 2024-02-20
Add APIGateway to manage chain requests

AbdelrahmanElawady:
### Description

Add APIGateway to manage chain requests

### Changes

- daemons will call chain requests through APIGateway

### Related Issues

- #2200 

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


pkg/api_gateway/api_gateway.go
```diff
@@ -0,0 +1,109 @@
...
+
+func (g *apiGateway) GetNodes(farmID uint32) ([]uint32, error) {
+	return g.sub.GetNodes(farmID)
+}
+
+func (g *apiGateway) GetPowerTarget(nodeID uint32) (power substrate.NodePower, err error) {
+	return g.sub.GetPowerTarget(nodeID)
+}
+
+func (g *apiGateway) GetTwin(id uint32) (*substrate.Twin, error) {
```
muhamadazmy: I am not sure how returning a pointer will behave over zbus, but i think it will not work. You need to return an object instead of its pointer for this to work correctly

cmds/modules/api_gateway/main.go
```diff
@@ -0,0 +1,86 @@
...
+	idStub := stubs.NewIdentityManagerStub(redis)
+
+	sk := ed25519.PrivateKey(idStub.PrivateKey(cli.Context))
+	id, err := substrate.NewIdentityFromEd25519Key(sk)
+	log.Info().Str("address", id.Address()).Msg("node address")
+	if err != nil {
+		return err
+	}
+
+	subtrateURL := environment.MustGet().SubstrateURL
```
muhamadazmy: You can do `environment.GetSubstrate()` directly which returns a manager. then pass the manager as it to apigateway.

pkg/api_gateway/api_gateway.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	if errors.Is(err, substrate.ErrNotFound) {
+		SubstrateError.Err = err
+		SubstrateError.Code = pkg.CodeNotFound
+	} else if err != nil {
+		SubstrateError.Err = err
+		SubstrateError.Code = pkg.CodeGenericError
+	}
+	return *contract, SubstrateError
+}
```
muhamadazmy: I suggest a different implementation approach for functions like these to make it easier for u as follows
```suggestion
func (g *apiGateway) GetContract(id uint64) (result substrate.Contract, serr pkg.SubstrateError) {
	log.Trace().Str("method", "GetContract").Uint64("id", id).Msg("method called")
	
	contract, err := g.sub.GetContract(id)
	serr.Err = err
	
	if errors.Is(err, substrate.ErrNotFound) {
		serr.Code = pkg.CodeNotFound
		return
	} else if err != nil {
		serr.Code = pkg.CodeGenericError
		return
	}
	// we sure that contract is not null here for sure
	return *contract, serr
}
```

muhamadazmy: Also a good thing is if you can have an error builder function 
```go
serr := buildError(err)
```

So it fills up all the serr fields correctly (code, and whatever) for all known substate error this then can reduce the code above to 

```go
	contract, err := g.sub.GetContract(id)
	
         serr = buildErr(err)
         If err != nil { 
                 return 
         }

        return *contract, serr
```

pkg/gateway/gateway.go
```diff
@@ -536,24 +531,20 @@ func (g *gatewayModule) configPath(name string) string {
...
 	}
-	contract, err := sub.GetContract(contractID)
-	if errors.Is(err, substrate.ErrNotFound) {
+	contract, subErr := g.apiGateway.GetContract(context.Background(), contractID)
+	if subErr.IsCode(pkg.CodeNotFound) {
 		return fmt.Errorf("contract by name returned %d, but retreiving it results in 'not found' error", contractID)
 	}
-	if err != nil {
-		return err
+	if subErr.IsError() {
```
muhamadazmy: `else if` ?

pkg/network/mbus.go
```diff
@@ -7,25 +7,20 @@ import (
...
 
 // Authorized middleware allows only admins to make these calls
-func Authorized(mgr substrate.Manager, farmID uint32) (rmb.Middleware, error) {
-	sub, err := mgr.Substrate()
-	if err != nil {
-		return nil, err
-	}
-	defer sub.Close()
-	farm, err := sub.GetFarm(farmID)
+func Authorized(apiGateway *stubs.APIGatewayStub, farmID uint32) (rmb.Middleware, error) {
```
muhamadazmy, line 17: I am wondering if it might be better if we pass the zbus.Client instead of a specific stub since it's more generic and other sub modules can use it.

On the other hand this code should eventually be moved to the apiGatway itself soon in the future so we can leave it this way for now.

muhamadazmy (changes requested):
Thank you for the hard work. I really think this made it much clearer and abstracts and streamline access to the chain.

I only have few comments.

pkg/api_gateway/api_gateway.go
```diff
@@ -0,0 +1,202 @@
...
+
+func (g *apiGateway) GetNode(id uint32) (substrate.Node, error) {
+	log.Trace().Str("method", "GetNode").Uint32("id", id).Msg("method called")
+	node, err := g.sub.GetNode(id)
+	if node == nil {
+		node = &substrate.Node{}
+	}
+	return *node, err
+}
```
muhamadazmy: This code will work perfectly. I just think it's confusing to check the value for nil or not instead of the error. So again, while the result of the code is similar I just believe that the code below is less confusing to read and hard to break if it's needed to be modified later. What do you think?

```suggestion
func (g *apiGateway) GetNode(id uint32) (result substrate.Node, err error) {
	log.Trace().Str("method", "GetNode").Uint32("id", id).Msg("method called")
	node, err := g.sub.GetNode(id)
	if err != nil {
		return
	}

	return *node, err
}
```

pkg/api_gateway/api_gateway.go
```diff
@@ -0,0 +1,202 @@
...
+
+func buildSubstrateError(err error) (serr pkg.SubstrateError) {
+	if err == nil {
+		return
+	}
+
+	serr.Err = err
+	serr.Code = pkg.CodeGenericError
+
+	if errors.Is(err, substrate.ErrNotFound) {
```
muhamadazmy, line 198: Maybe it's a good idea to also add the few remaining error codes from the substrate package  for the future and completeness of the code

image

muhamadazmy (changes requested):
few tiny comments

muhamadazmy (approved):

Issue #2205, opened 2024-02-21
Add an RMB call for getting all deployments for a twin


AbdelrahmanElawady:
## Description

Add an RMB call for getting all deployments for a twin instead of calling `zos.deployment.get` for each one.

Issue #2204, opened 2024-02-21
Add RMB call for listing a network used private IPs


AbdelrahmanElawady:
## Description

Add an RMB call that receives a network name and then return the used private IPs on this network inside the node.

PR #2206, opened 2024-02-21
Add list and list_private_ips rmb calls

AbdelrahmanElawady:
### Description

Add RMB calls to get all deployments for a twin and get private ips reserved in a network

### Changes

- `zos.deployment.list` RMB call to get all deployments.
- `zos.network.list_private_ips` to get private IPs reserved for a given network.

### Related Issues

- #2204
- #2205

### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation

### DIFF
--- a/client/node.go
+++ b/client/node.go
@@ -146,9 +146,9 @@ func (n *NodeClient) DeploymentGet(ctx context.Context, contractID uint64) (dl g
 	return dl, nil
 }
 
-// DeploymentGetAll gets all deployments for a twin
-func (n *NodeClient) DeploymentGetAll(ctx context.Context) (dls []gridtypes.Deployment, err error) {
-	const cmd = "zos.deployment.get_all"
+// DeploymentList gets all deployments for a twin
+func (n *NodeClient) DeploymentList(ctx context.Context) (dls []gridtypes.Deployment, err error) {
+	const cmd = "zos.deployment.list"
 
 	err = n.bus.Call(ctx, n.nodeTwin, cmd, nil, &dls)
 	return
--- a/client/node.go
+++ b/client/node.go
@@ -146,6 +146,14 @@ func (n *NodeClient) DeploymentGet(ctx context.Context, contractID uint64) (dl g
 	return dl, nil
 }
 
+// DeploymentGetAll gets all deployments for a twin
+func (n *NodeClient) DeploymentGetAll(ctx context.Context) (dls []gridtypes.Deployment, err error) {
+	const cmd = "zos.deployment.get_all"
+
+	err = n.bus.Call(ctx, n.nodeTwin, cmd, nil, &dls)
+	return
+}
+
 // DeploymentGet gets a deployment via contract ID
 func (n *NodeClient) DeploymentChanges(ctx context.Context, contractID uint64) (changes []gridtypes.Workload, err error) {
 	const cmd = "zos.deployment.changes"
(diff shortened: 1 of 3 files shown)


client/node.go
```diff
@@ -146,6 +146,14 @@ func (n *NodeClient) DeploymentGet(ctx context.Context, contractID uint64) (dl g
 	return dl, nil
 }
 
+// DeploymentGetAll gets all deployments for a twin
+func (n *NodeClient) DeploymentGetAll(ctx context.Context) (dls []gridtypes.Deployment, err error) {
+	const cmd = "zos.deployment.get_all"
```
xmonader: I think we should call it `zos.deployment.list`

pkg/provision/mbus/reservation.go
```diff
@@ -80,6 +80,25 @@ func (d *Deployments) get(ctx context.Context, payload []byte) (interface{}, mw.
 	return deployment, nil
 }
 
+func (d *Deployments) getAll(ctx context.Context, payload []byte) (interface{}, mw.Response) {
```
xmonader: I think should be `list`

pkg/provision/mbus/api.go
```diff
@@ -85,6 +88,42 @@ func (n *Deployments) listPublicIps(ctx context.Context, _ []byte) (interface{},
 	return ips, nil
 }
 
+type nameArgs struct {
```
xmonader: shouldn't that be `listPrivateIPsArgs` ?

pkg/provision/mbus/api.go
```diff
@@ -85,6 +88,42 @@ func (n *Deployments) listPublicIps(ctx context.Context, _ []byte) (interface{},
...
+	var args nameArgs
+	if err := json.Unmarshal(payload, &args); err != nil {
+		return nil, err
+	}
+	deployments, err := d.getAll(ctx, payload)
+	if err != nil {
+		return nil, err.Err()
+	}
+	dls := deployments.([]gridtypes.Deployment)
+	ips := make([]string, 0)
```
xmonader, line 104: they are usable at zero value and len check works on them too

pkg/provision/mbus/api.go
AbdelrahmanElawady: it was done like this for consistency with the other [id](https://github.com/threefoldtech/zos/blob/main/pkg/provision/mbus/reservation.go#L16) input struct, but sure will change it

xmonader (dismissed):

pkg/provision/mbus/api.go
```diff
@@ -85,6 +88,42 @@ func (n *Deployments) listPublicIps(ctx context.Context, _ []byte) (interface{},
...
+func (d *Deployments) listPrivateIps(ctx context.Context, payload []byte) (interface{}, error) {
+	var args listPrivateIpsArgs
+	if err := json.Unmarshal(payload, &args); err != nil {
+		return nil, err
+	}
+	deployments, err := d.list(ctx, payload)
+	if err != nil {
+		return nil, err.Err()
+	}
+	dls := deployments.([]gridtypes.Deployment)
```
muhamadazmy: Why assertion? it's not a good idea to do type assertion because changes to the list return type can break this in runtime and the compiler will not know about it. Instead do the following:
```
fn listDeployments(ctx, twin_id) []Depoyments 
fn list() {
    return listDeployments(ctx, GetTwinId(ctx))
}
fn list_piv_ips() {
    deployments = listDeployments(ctx, GetTwinId(ctx))
    // work with deployments
}
```

muhamadazmy (changes requested):

pkg/provision/mbus/api.go
```diff
@@ -128,3 +166,11 @@ func (d *Deployments) changesHandler(ctx context.Context, payload []byte) (inter
 	}
 	return data, nil
 }
+
+func (d *Deployments) listHandler(ctx context.Context, payload []byte) (interface{}, error) {
+	data, err := d.list(ctx, payload)
+	if err != nil {
+		return nil, err
+	}
+	return data, nil
```
muhamadazmy:
```suggestion
	return d.list(ctx, payload)
```

This should work, no ?

pkg/provision/mbus/api.go
```diff
@@ -85,6 +88,41 @@ func (n *Deployments) listPublicIps(ctx context.Context, _ []byte) (interface{},
 	return ips, nil
 }
 
+type listPrivateIpsArgs struct {
+	NetworkName string `json:"network_name"`
```
muhamadazmy:
```suggestion
	NetworkName gridtypes.Name `json:"network_name"`
```

If you do this you won't need the cast in the code anymore

muhamadazmy (dismissed):

muhamadazmy (approved):

Issue #2213, opened 2024-03-01
don't send uptime on failure to resolve grid services domains


xmonader:
don't send uptime on failure to resolve grid services domains this should be a flag that the node is unhealthy, maybe we can try to do that 3 times before sending the uptime report, and if we fail we don't send it

PR #2215, opened 2024-03-03
stop sending uptime reports if the node can not reach grid services

Omarabdul3ziz:
### Description
stop sending uptime reports if the node can not reach grid services

### Changes
- add new `NotReachable` flag
- set the flag each time fails to resolve a service url, and unset on each new NetworkHealth run

### Related Issues
- https://github.com/threefoldtech/zos/issues/2213

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/pkg/power/uptime.go
+++ b/pkg/power/uptime.go
@@ -129,5 +129,9 @@ func isNodeHealthy() bool {
 		log.Error().Msg("node is running on limited cache")
 		healthy = false
 	}
+	if app.CheckFlag(app.NotReachable) {
+		log.Error().Msg("node can not reach grid services")
+		healthy = false
+	}
 	return healthy
 }
--- a/pkg/app/flag.go
+++ b/pkg/app/flag.go
@@ -15,6 +15,8 @@ const (
 	LimitedCache = "limited-cache"
 	// ReadonlyCache represents the flag for cache is read only
 	ReadonlyCache = "readonly-cache"
+	// NotReachable represents the flag when a grid service is not reachable
+	NotReachable = "not-reachable"
 )
 
 // SetFlag is used when the /var/cache cannot be mounted on a SSD or HDD,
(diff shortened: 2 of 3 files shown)


pkg/power/uptime.go
```diff
@@ -129,5 +129,9 @@ func isNodeHealthy() bool {
 		log.Error().Msg("node is running on limited cache")
 		healthy = false
 	}
+	if app.CheckFlag(app.NotReachable) {
```
xmonader, line 132: Why not invoke the Network health task here and if it returns an error you return false? then there will be no need to add more flags in app/flag.go or create more side effects in networkhealth.go

pkg/perf/networkhealth/networkhealth.go
```diff
@@ -45,6 +47,10 @@ func (t *NetworkHealthTask) Jitter() uint32 {
 }
 
 func (t *NetworkHealthTask) Run(ctx context.Context) (interface{}, error) {
+	if err := app.DeleteFlag(app.NotReachable); err != nil {
+		log.Error().Err(err).Msg("failed to unset not reachable flag")
+	}
+
```
muhamadazmy: I am wondering if it's a good idea to delete the flag before running the test. What do you think?

muhamadazmy (changes requested):
The general idea is okay. But i think the healthcheck task is a better place to have this check. This way it fits in the general idea of "health" of the node. It's also available via the perf stats api as well.

pkg/perf/networkhealth/networkhealth.go
Omarabdul3ziz: modified to delete the notreachable flag after the run only if there are no errors

pkg/power/uptime.go
Omarabdul3ziz: preferred to be aligned with the other health check, and make the uptime only check the flags. ok?

muhamadazmy (approved):
Can you run a test somehow to make sure this works fine. And that it is restored after the connection is restored

You can test this by modifying the zos /etc/hosts file to override one the service names to local ip or wrong address and watch it set the flag

Issue #2214, opened 2024-03-03
update the usage of the default http goclient to use proper timeouts instead of no timeout


xmonader:
## Describe the bug

Most of our code is using `http.Get` depending on the default go's http client behavior which is no timeout which can leads to 
- stalling execution
- resource leakage

Create a custom client whenever needed and set the proper timeout based on the operation 

## Screenshots

![image](https://github.com/threefoldtech/zos/assets/64129/08dd488c-f7d3-4758-8493-b28892ebf3e1)

PR #2216, opened 2024-03-03
add proper timeouts for the http get calls

Omarabdul3ziz:
### Description
by creating a custom HTTP client and assigning a default value as a timeout based on each operation

### Related Issues
- https://github.com/threefoldtech/zos/issues/2214

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/pkg/environment/config.go
+++ b/pkg/environment/config.go
@@ -27,10 +27,6 @@ type Config struct {
 	} `json:"mycelium"`
 }
 
-type httpClientOption struct {
-	timeout time.Duration
-}
-
 // Merge, updates current config with cfg merging and override config
 // based on some update rules.
 func (c *Config) Merge(cfg Config) {
@@ -46,16 +42,16 @@ func GetConfig() (base Config, err error) {
 		return
 	}
 
-	opts := httpClientOption{
-		timeout: defaultHttpTimeout,
+	httpClient := &http.Client{
+		Timeout: defaultHttpTimeout,
 	}
 
-	base, err = getConfig(env.RunningMode, baseExtendedURL, opts)
+	base, err = getConfig(env.RunningMode, baseExtendedURL, httpClient)
 	if err != nil {
 		return
 	}
 	if env.ExtendedConfigURL != "" {
-		custom, err := getConfig(env.RunningMode, env.ExtendedConfigURL, opts)
+		custom, err := getConfig(env.RunningMode, env.ExtendedConfigURL, httpClient)
 		if err != nil {
 			log.Error().Err(err).Msg("fetching the config from the env config-url failed")
 		}
@@ -77,17 +73,13 @@ func uniqueStr(slice []string) []string {
 	return list
 }
 
-func getConfig(run RunMode, url string, opts httpClientOption) (ext Config, err error) {
+func getConfig(run RunMode, url string, httpClient *http.Client) (ext Config, err error) {
 	if !strings.HasSuffix(url, "/") {
 		url += "/"
 	}
 	u := url + fmt.Sprintf("%s.json", run)
 
-	cl := &http.Client{
-		Timeout: opts.timeout,
-	}
-
-	response, err := cl.Get(u)
+	response, err := httpClient.Get(u)
 	if err != nil {
 		return ext, err
 	}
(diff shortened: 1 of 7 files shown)


xmonader (dismissed):
Looks good!

pkg/environment/config.go
```diff
@@ -36,12 +41,12 @@ func GetConfig() (base Config, err error) {
 	if err != nil {
 		return
 	}
-	base, err = getConfig(env.RunningMode, baseExtendedURL)
+	base, err = getConfig(env.RunningMode, baseExtendedURL, defaultHttpTimeout)
```
muhamadazmy: This is good solution of course. But I am wondering if in the future we needed to customize more of the client behaviour that the argument list will keep growing. May be there is a better way to do this 🤷🏼‍♂️

pkg/flist/flist.go
```diff
@@ -603,7 +605,12 @@ func (f *flistModule) Unmount(name string) error {
 func (f *flistModule) FlistHash(url string) (string, error) {
 	// first check if the md5 of the flist is available
 	md5URL := url + ".md5"
-	resp, err := http.Get(md5URL)
+
+	cl := &http.Client{
```
muhamadazmy: the http client in Go actually uses http2 which means it can reuse connections. But of course this won't work if we create a client everytime a client is needed.

I suggest that may be we instead share the client, maybe make it an attribute of the flistModule. this way we won't need to recreate everytime.

pkg/geoip/geoip.go
```diff
@@ -47,7 +52,11 @@ func getLocation(geoIPService string) (Location, error) {
 		City:      "Unknown",
 	}
 
-	resp, err := http.Get(geoIPService)
+	cl := &http.Client{
```
muhamadazmy: the geoip pkg is kinda different. Although i don't like global vars may be instead of declaring the timeout as const we can have a global (private) defaultClient in this pkg this way we also don't have to create an instance everytime. and will behave similar to `http.DefaultClient`

pkg/upgrade/hub/hub.go
```diff
@@ -86,7 +89,11 @@ func (h *HubClient) Info(repo, name string) (info FList, err error) {
 	}
 
 	u.Path = filepath.Join("api", "flist", repo, name, "light")
-	response, err := http.Get(u.String())
+	cl := &http.Client{
```
muhamadazmy: the HubClient can obviously also use a client attribute instead of re-creating the client. This way it can use the connection reuse features of http2 and saves on memory

muhamadazmy (changes requested):

pkg/environment/config.go
```diff
@@ -36,12 +45,17 @@ func GetConfig() (base Config, err error) {
 	if err != nil {
 		return
 	}
-	base, err = getConfig(env.RunningMode, baseExtendedURL)
+
+	opts := httpClientOption{
```
muhamadazmy: I was actually trying to push you to pass the Client itself  not config.

muhamadazmy: We don't need an option struct. Just create the client at the top level then pass it around. Imagine now if u added a new flag to the options. Then u will have to go everywhere and modify the client creation to actually use this flag. Right?

But if you create the Client onetime at the top level, then pass it around and later decided to change one of the client options, u can do it only in this one place

muhamadazmy (changes requested):
- Tests are falling
- Check my comment

pkg/upgrade/upgrade.go
```diff
@@ -95,8 +96,10 @@ func Zinit(socket string) func(u *Upgrader) error {
 
 // NewUpgrader creates a new upgrader instance
 func NewUpgrader(root string, opts ...UpgraderOption) (*Upgrader, error) {
+	hubClient := hub.NewHubClient(defaultHubTimeout)
 	u := &Upgrader{
 		root: root,
+		hub:  *hubClient,
```
muhamadazmy: when doing a `deref` (like *something) this copies the value. I think it might be better to change the `hub` attribute to `hub *hub.HubClient` instead.

muhamadazmy (approved):

PR #2217, opened 2024-03-04
move the network check to the health check task

Omarabdul3ziz:
### Description
move the network check to the health check task

### Changes
- remove the standalone perf task network-health and add it as a check in the healthcheck task.
- modify the schedule of the healthcheck to be every 15 instead of 20 mins
- modify the report of the network check to only report the failed services for a better alignment with the other healthcheck
- updated the diagnostics package to check for the healthcheck

### Related Issues

List of related issues

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/pkg/diagnostics/diagnostics.go
+++ b/pkg/diagnostics/diagnostics.go
@@ -14,7 +14,7 @@ import (
 
 const (
 	callTimeout    = 3 * time.Second
-	testNetworkKey = "perf.network-health"
+	testNetworkKey = "perf.healthcheck"
 )
 
 // Modules is all the registered modules on zbus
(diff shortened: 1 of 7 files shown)


pkg/diagnostics/diagnostics.go
```diff
@@ -115,7 +115,7 @@ func (m *DiagnosticsManager) getModuleStatus(ctx context.Context, module string)
 	}
 }
 
-func (m *DiagnosticsManager) isOnline(ctx context.Context) bool {
+func (m *DiagnosticsManager) isHealthy() bool {
```
muhamadazmy, line 118: The idea of diagnostics to return realtime health of system modules. I am wondering if we need to return health also since health checks results can also be retrieved over the perf api which is already exposed

muhamadazmy: Maybe it's better approach to have diagnostics unrelated to healthchecks. Unless there is something else i am missing

muhamadazmy (commented):
This is much better. But i have only one comment to discuss. Let me know what you think

pkg/diagnostics/diagnostics.go
Omarabdul3ziz: it was asked to add the network state for the diagnostics call for a future plan to use this call in the proxy or clients to decide if the node is reachable or not.
since we merged the checks (network/cache) i thought it is better to check for both and decide the total node health instead of just the network. 

so in the `proxy/node-indexer` the diagnostics call will be used to check for both fields `SystemOk` and `Healthy` to decide if the node is healthy or not

muhamadazmy (approved):

PR #2223, opened 2024-03-07
Move all rmb handlers to api-gateway

AbdelrahmanElawady:
### Description

Move all rmb handlers to api-gateway and use `peer.Router` instead of using `rmb-peer` binary.

### Changes

- All defined handlers and endpoints are now found under `pkg/api-gateway`.
- Unimplemented endpoints reply immediately that it's not implemented.

### Related Issues

- #2200 

### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation


cmds/modules/api_gateway/main.go
```diff
@@ -65,17 +67,33 @@ func action(cli *cli.Context) error {
...
 	utils.OnDone(ctx, func(_ error) {
 		log.Info().Msg("shutting down")
 	})
+	_, err = peer.NewPeer(
+		ctx,
+		hex.EncodeToString(pair.Seed()),
+		manager,
+		router.Serve,
+		peer.WithKeyType(peer.KeyTypeEd25519),
+		peer.WithRelay(environment.MustGet().RelayURL[0]),
```
muhamadazmy: We need to support having multiple relays. This is probably need to be modified in the sdk peer already.

pkg/api_gateway/network.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/threefoldtech/zos/pkg/gridtypes"
+)
+
+func (g *apiGateway) networkListWGPortsHandler(ctx context.Context, payload []byte) (interface{}, error) {
+	return g.networkerStub.WireguardPorts(ctx)
+}
+func (g *apiGateway) networkPublicConfigGetHandler(ctx context.Context, payload []byte) (interface{}, error) {
+	return g.networkerStub.GetPublicConfig(ctx)
+}
+func (g *apiGateway) networkInterfacesHandler(ctx context.Context, payload []byte) (interface{}, error) {
```
muhamadazmy: I am just wondering if this logic should be part of the api, or networkd. I mean the api here shows understanding of the internals of some network internals. 

Just a thought, no need to modify it immediately

cmds/modules/noded/main.go
```diff
@@ -297,12 +228,7 @@ func action(cli *cli.Context) error {
 
 	server.Register(zbus.ObjectID{Name: "host", Version: "0.0.1"}, host)
 	server.Register(zbus.ObjectID{Name: "system", Version: "0.0.1"}, system)
-
-	go func() {
-		if err := server.Run(ctx); err != nil && err != context.Canceled {
-			log.Fatal().Err(err).Msg("unexpected error")
-		}
-	}()
+	server.Register(zbus.ObjectID{Name: "host", Version: "0.0.1"}, perfMon)
```
muhamadazmy: isn't this a double register of the object `host` ?

muhamadazmy: check the return error from register. this should return an error because you trying to register an object with the same id

muhamadazmy (changes requested):
Amazing job Abderahman. I only have few comments below. We need to also make this is well tested before we deploy to devnet.

pkg/api_gateway/network.go
AbdelrahmanElawady: I think it's mostly preparing input for networkd and parsing its output so it isn't really a lot of networking logic, that's why I didn't move it.

cmds/modules/api_gateway/main.go
AbdelrahmanElawady: yep, it's already being done [here](https://github.com/threefoldtech/tfgrid-sdk-go/issues/917)

ashraffouda (dismissed):

muhamadazmy (approved):

Issue #2188, opened 2024-01-15
internet package should try also other urls than bootstrap.grid.tf


xmonader:
because it might be down and we won't be sure if we got actual internet and can reach bootstrap.grid.tf or if we have it and the bootstrap website is down

muhamadazmy:
we actually then need to test the `hub.grid.tf` because if that one is down we won't be able to continue the boot anyway

muhamadazmy:
@Eslam-Nawara this is a very easy task. we have an `internet` binary (under cmds/internet) that tries to test connectivity to some public services. We need to make sure we also include the hub.grid.tf

ashraffouda:
verified

PR #2224, opened 2024-03-07
test internet connection against hub.grid.tf

Eslam-Nawara:
### Description

test internet connection against hub.grid.tf

### Changes

add new connectivity test against hub.grid.tf in internet pkg

### Related Issues

- #2188 

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/cmds/internet/main.go
+++ b/cmds/internet/main.go
@@ -85,7 +85,6 @@ func check() error {
 	f := func() error {
 		retries += 1
 
-		log.Info().Msg("testing internet connection. trying out bootstrap.grid.tf:80")
 		// print some helpful debugging information to help with debugging issues
 		// if internet connection wasn't established
 		if err := debugZos(); err != nil {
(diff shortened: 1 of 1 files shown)


cmds/internet/main.go
```diff
@@ -94,7 +93,19 @@ func check() error {
 
 		// we only care about possibility of establishing a connection
 		// so just establishing a connection then close it is good enough
-		con, err := net.Dial("tcp", "bootstrap.grid.tf:http")
+
+		log.Info().Msg("testing internet connection. trying out hub.grid.tf:80")
+		con, err := net.Dial("tcp", "hub.grid.tf:http")
```
xmonader, line 98: Should we retry that maybe 3 times? @muhamadazmy @Eslam-Nawara

muhamadazmy: the entire function is retried forever anyway, until it succeeds

muhamadazmy (approved):

cmds/internet/main.go
```diff
@@ -94,7 +93,19 @@ func check() error {
...
+		if err != nil {
+			return errors.Wrap(err, "failed to reach hub.grid.tf")
+		}
+
+		if err := con.Close(); err != nil {
+			return err
+		}
+
+		log.Info().Msg("testing internet connection. trying out bootstrap.grid.tf:80")
+		con, err = net.Dial("tcp", "bootstrap.grid.tf:http")
```
muhamadazmy, line 108: Why not instead of copying the code you do that in loop instead 

```go
 for _, address := range []string{"bootstrap.grid.tf:http", "hub.grid.tf:http"} {
    //try to dial or return on error
    // close connection
}
```

this way u don't have to copy same code

muhamadazmy (changes requested):
The branch name causes the build to fail. I also can't manage to checkout the branch locally in my repo.

Issue #2234, opened 2024-03-14
cloud-init cidata seed are not deleted when Vm is deleted


muhamadazmy:
## Describe the bug

cloud-init cidata seed are not deleted when Vm is deleted

## To Reproduce
- create a vm 
- delete the vm
- cidata for this Vm under /var/cache/modules/vmd/cloud-init/<seed-disk>  is not deleted

Eslam-Nawara:
I checked the issue and tried to reproduce but found that the cidata were removed with no issues
- I tried deploying 3 vms
- deleted two vms and checked the data path
- deleted the last vm and checked the path and every thing is looking fine to me 
``` console
~ # ls -al /var/cache/modules/vmd/cloud-init
total 120
drwxr-xr-x    1 root     root           114 Mar 19 10:10 .
drwxr-xr-x    1 root     root            28 Nov 28 14:11 ..
-rw-r--r--    1 root     root       2097152 Mar 19 10:04 4653-103178-vmkax35
-rw-r--r--    1 root     root       2097152 Mar 19 10:09 4653-103184-vmkax36
-rw-r--r--    1 root     root       2097152 Mar 19 10:10 4653-103186-vmkax37
~ # ls -al /var/cache/modules/vmd/cloud-init
total 40
drwxr-xr-x    1 root     root            38 Mar 19 10:13 .
drwxr-xr-x    1 root     root            28 Nov 28 14:11 ..
-rw-r--r--    1 root     root       2097152 Mar 19 10:04 4653-103178-vmkax35
~ # ls -al /var/cache/modules/vmd/cloud-init
total 0
drwxr-xr-x    1 root     root             0 Mar 19 10:15 .
drwxr-xr-x    1 root     root            28 Nov 28 14:11 ..
~ #
``` 

I've noticed that the deletion errors are ignored, maybe we should log the errors if there is any

PR #2241, opened 2024-03-18
remove cloud-init cidata on deleting vms

Eslam-Nawara:
### Description

remove cloud-init cidata seed when Vm is deleted

### Related Issues

- #2234 
### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/pkg/primitives/vm/vm.go
+++ b/pkg/primitives/vm/vm.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net"
 	"os"
+	"path/filepath"
 
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog/log"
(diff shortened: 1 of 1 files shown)


xmonader (approved):

pkg/primitives/vm/vm.go
```diff
@@ -325,6 +326,12 @@ func (p *Manager) Deprovision(ctx context.Context, wl *gridtypes.WorkloadWithID)
 		log.Error().Err(err).Str("name", volName).Msg("failed to delete rootfs volume")
 	}
 
+	cidataPath := filepath.Join("var", "cache", "modules", "vmd", "cloud-init", wl.ID.String())
```
muhamadazmy, line 329: While this will work, but it's completely out of place. The concept of separation of concerns dictates that a single module/object should not know about the details implementations of another module. 

It would be logical that the module that actually creates the cidata file is the one that also deletes it. no?

Otherwise the code logic will be scattered all over the place and will be completely unmanageable.

muhamadazmy: You can see already that you had to define a path that is completely static. what if the modules that creates the cidata now decided to change this path, is this still gonna work?

muhamadazmy (changes requested):
Please check my comments on the PR.
I also checked the code of `vmd` and found that we already have code to clean up cidata but for some reason it doesn't work for some reason

https://github.com/threefoldtech/zos/blob/main/pkg/vm/manager.go#L587


Please spend some time following code paths and logic before jumping to implementation

PR #2243, opened 2024-03-20
run cidata cleanup periodically

Eslam-Nawara:
### Description

if a problem happened deleting vm and cidata didn't get deleted it will be left forever

### Changes

- added a cleanup for cidata every 10 minuites 

### Related Issues

- #2234 

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/pkg/vm/monitor.go
+++ b/pkg/vm/monitor.go
@@ -20,6 +20,7 @@ const (
 	failuresBeforeDestroy = 4
 	monitorEvery          = 10 * time.Second
 	logrotateEvery        = 10 * time.Minute
+	cleanupEvery          = 10 * time.Minute
 )
 
 var (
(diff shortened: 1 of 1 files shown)


pkg/vm/monitor.go
```diff
@@ -118,6 +125,32 @@ func (m *Module) monitor(ctx context.Context) error {
 	return nil
 }
 
+func (m *Module) cleanupCidata() error {
```
muhamadazmy, line 128: This look really good now. I only have one comment:

I think it's important to make sure that this does not race with the creation of VM. Imagine while a machine is still starting (it's cidata has been created but the vm is still not running) the timer ticked and u started the clean up. 

This can cause a big problem, no? Luckily the solution is easy since we already have a lock that is acquired on creation, it's also acquired on `monitor` which also does some clean up.

Just include those 2 lines at the beginning of the function
```go
	m.lock.Lock()
	defer m.lock.Unlock()
```

muhamadazmy (changes requested):

muhamadazmy (approved):

PR #2247, opened 2024-03-27
delete the whole keys to reset the errors map

Omarabdul3ziz:
### Description
when a diagnostics call is made it checks on the length of the map to report unhealthy if the errors map is not empty.
setting keys to nil leaves the key in the map so it will fail the check

### Changes
delete the whole keys to reset the errors map on each healthcheck run

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/pkg/perf/healthcheck/healthcheck.go
+++ b/pkg/perf/healthcheck/healthcheck.go
@@ -62,7 +62,7 @@ func (h *healthcheckTask) Run(ctx context.Context) (interface{}, error) {
 	log.Debug().Msg("starting health check task")
 	for k := range h.errors {
 		// reset errors on each run
-		h.errors[k] = nil
+		delete(h.errors, k)
 	}
 
 	for label, check := range h.checks {
--- a/pkg/perf/healthcheck/healthcheck.go
+++ b/pkg/perf/healthcheck/healthcheck.go
@@ -65,8 +65,8 @@ func (h *healthcheckTask) Run(ctx context.Context) (interface{}, error) {
 		if len(errors) == 0 {
 			continue
 		}
-		stringErrs := errorsToStrings(errors)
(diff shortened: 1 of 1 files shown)


AbdelrahmanElawady (approved):

pkg/perf/healthcheck/healthcheck.go
```diff
@@ -62,7 +62,7 @@ func (h *healthcheckTask) Run(ctx context.Context) (interface{}, error) {
 	log.Debug().Msg("starting health check task")
 	for k := range h.errors {
 		// reset errors on each run
-		h.errors[k] = nil
```
AbdelrahmanElawady, line 65: This will cause the errors on ZUI to be permanent since we loop over errors map and send the label with its errors. If the label was removed it won't be set in case of no errors so it won't be sent to ZUI for resetting.

AbdelrahmanElawady (changes requested):

AbdelrahmanElawady (dismissed):

pkg/perf/healthcheck/healthcheck.go
```diff
@@ -60,30 +58,27 @@ func (h *healthcheckTask) Description() string {
...
+	errs := make(map[string][]string)
 
 	for label, check := range h.checks {
 		errors := check(ctx)
 		if len(errors) == 0 {
 			continue
 		}
 		stringErrs := errorsToStrings(errors)
-		h.errors[label] = append(h.errors[label], stringErrs...)
+		errs[label] = append(errs[label], stringErrs...)
```
muhamadazmy: I think this can just be 
```suggestion
		errs[label] = stringErrs
```
since errs[label] will always be nil.

muhamadazmy (commented):
just one comment

muhamadazmy (approved):

Issue #2250, opened 2024-03-29
managed ssh keys


muhamadazmy:
Right now the zos imagine contains few authorized keys for some developers to allow remote debugging*. IMHO this shouldn't be hardcoded in the base image but instead, managed upon request as follows:

- On boot, identityd should take control over authorized keys. It can start it completely removing it from the root. This will make sure the base images keys are ineffective. 
- Based on the set of usernames that should be defined in `zos-config` repo (need to be added), the defined users keys are then retrieved fro github and updated into the authorized_keys files. ONLY this set of users are.
  - Normally this list then should be empty for mainnet.
  - Should this keys be fetched periodically ? or is it a one time thing!
- For all networks except mainnet, identityd should also respect an optional kernel cmdline argument (say `ssh-key`) that accepts a username, this user key is fetched from github and added to the list of pre-configured users from zos-config before updating the authorized keys files.
- Special case for mainnet freefarm since we always need access to this one for development and debugging purposes, the keys should use the list from testnet always.

> *remote debugging: this is only possible to networks that are actually reachable by developers so in real life this is probably only possible for freefarm and local development nodes

Eslam-Nawara:
# Investigation and Solution:
- currently, I'm investigating the issue and have questions,
   - if we assumed that this is run only once, is there something indicating that `identityd` is running for the first time?
   - the other case if we run it periodically, are we sure that the user won't add another ssh key manually to the `authorized_keys` file, then the update will overwrite it?
- investigating `google/go-github` to use it to get a file content from GitHub.

PR #2251, opened 2024-04-01
manage ssh keys

Eslam-Nawara:
### Description

- manage ssh keys

### Changes

- remove authorized keys hardcoded in the base image
- get authorized users from `zos-config` and get their keys
- update `.ssh/authorized-keys` with the new keys

### Related Issues

- #2250 

### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation

### DIFF
--- a/cmds/identityd/main.go
+++ b/cmds/identityd/main.go
@@ -2,8 +2,11 @@ package main
 
 import (
 	"context"
+	"encoding/json"
+	"flag"
 	"fmt"
 	"os"
+	"path/filepath"
 	"time"
 
 	"github.com/pkg/errors"
@@ -15,7 +18,7 @@ import (
 	"github.com/threefoldtech/zos/pkg/environment"
 	"github.com/threefoldtech/zos/pkg/identity"
 
-	"flag"
+	"github.com/google/go-github/v60/github"
 
 	"github.com/rs/zerolog/log"
 	"github.com/threefoldtech/zbus"
@@ -146,7 +149,6 @@ func main() {
 		log.Error().Err(err).Msg("error during update")
 		os.Exit(1)
 	}
-
 }
 
 func getIdentityMgr(root string, debug bool) (pkg.IdentityManager, error) {
@@ -171,4 +173,50 @@ func getIdentityMgr(root string, debug bool) (pkg.IdentityManager, error) {
 		Msg("farmer identified")
 
 	return manager, nil
+}
+
+func authKeysHandler(ctx context.Context, env string, farm uint64) error {
+	authorizedKeysPath := filepath.Join("/", "root", ".ssh", "authorized_keys")
+
+	err := os.Remove(authorizedKeysPath)
+	if err != nil {
+		return err
+	}
+
+	if env == "main" {
+		if farm != 1 {
+			return nil
+		}
+		env = "testing"
+	}
+
+	// get the content of the file in `zos-config` according to the network
+	github := github.NewClient(nil)
+	file, _, _, err := github.Repositories.GetContents(ctx, "threefoldtech", "zos-config", "authorized-users.json", nil)
+	if err != nil {
+		return err
+	}
+
+	fileContentJSON, err := file.GetContent()
+	if err != nil {
+		return err
+	}
+
+	var authorizedUsers map[string][]string
+	err = json.Unmarshal([]byte(fileContentJSON), &authorizedUsers)
+	if err != nil {
+		return err
+	}
+
+	var allKeys string
+
+	for _, user := range authorizedUsers[env] {
+		keys, _, err := github.Users.ListKeys(ctx, user, nil)
+		if err != nil {
+			return err
+		}
+		allKeys = fmt.Sprintf("%s\n%s", allKeys, keys[0].GetKey())
+		fmt.Println(keys)
+	}
+	return os.WriteFile(authorizedKeysPath, []byte(allKeys), 0644)
 }
--- a/cmds/identityd/main.go
+++ b/cmds/identityd/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"context"
-	"encoding/json"
 	"flag"
 	"fmt"
 	"os"
@@ -10,7 +9,9 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/sethvargo/go-retry"
 	"github.com/threefoldtech/zos/pkg/app"
+	"github.com/threefoldtech/zos/pkg/kernel"
 	"github.com/threefoldtech/zos/pkg/stubs"
 	"github.com/threefoldtech/zos/pkg/upgrade"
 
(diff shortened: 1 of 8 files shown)


cmds/identityd/main.go
```diff
@@ -2,20 +2,24 @@ package main
...
+	"github.com/threefoldtech/zos/pkg/kernel"
 	"github.com/threefoldtech/zos/pkg/stubs"
 	"github.com/threefoldtech/zos/pkg/upgrade"
 
 	"github.com/threefoldtech/zos/pkg"
 	"github.com/threefoldtech/zos/pkg/environment"
 	"github.com/threefoldtech/zos/pkg/identity"
 
-	"flag"
+	"github.com/google/go-github/v60/github"
```
muhamadazmy: We don't need to use an external library, please remove this. All user keys are available at 

`https://github.com/<username>.keys` 

You can try this yourself. So there is absolutely no need to fetch this unnecessary dependency

qemu/README.md
```diff
@@ -112,11 +112,12 @@ sudo dnsmasq --strict-order \
...
 
 ```bash
-sudo ./vm.sh -n node-01 -c "farmer_id=47 printk.devmsg=on runmode=dev"
+sudo ./vm.sh -n node-01 -c "farmer_id=47 printk.devmsg=on runmode=dev user=<github username>"
 ```
 
 where `runmode` is one of `dev` , `test`  or `prod`, 
-and `farmer_id` is the id of the farm you registered with `tffarmer`
+      `farmer_id` is the id of the farm you registered with `tffarmer`
+      `user` is github username provided if a user need to pass ssh-key to the node
```
muhamadazmy: I think the name `user` is very generic and can be confusing. Plus, i see that in the qemu/Makefile you actually pass it as `authorized_user`

cmds/identityd/main.go
```diff
@@ -172,3 +180,56 @@ func getIdentityMgr(root string, debug bool) (pkg.IdentityManager, error) {
...
+
+	if found {
+		authorizedUsers = append(authorizedUsers, extraUser)
+	}
+
+	github := github.NewClient(nil)
+	var allKeys string
+
+	for _, user := range authorizedUsers {
+		if err := retry.Do(ctx, retry.WithMaxRetries(3, retry.NewConstant(1*time.Microsecond)), func(ctx context.Context) error {
```
muhamadazmy: it's nice that you retry this, it's a good idea. But we already have a package that we use for retries called `backoff` it's already available in zos. So better use backoff instead of fetching another dependency.

cmds/identityd/main.go
```diff
@@ -172,3 +180,56 @@ func getIdentityMgr(root string, debug bool) (pkg.IdentityManager, error) {
...
+		runningMode = environment.RunningTest
+	}
+
+	config, err := environment.GetConfigForMode(runningMode)
+	if err != nil {
+		return err
+	}
+
+	authorizedUsers := config.Users.Authorized
+	extraUser, found := kernel.GetParams().GetOne("authorized_user")
```
muhamadazmy: `authorized_user` while this is a correct name for this variable but I think it's not clear what this user is authorized to. Right? this can a little bit confusing. I think a more accurate name like `ssh-user` would be better choice, so it's clear that this user will have ssh access.

cmds/identityd/main.go
```diff
@@ -172,3 +180,56 @@ func getIdentityMgr(root string, debug bool) (pkg.IdentityManager, error) {
...
+	if err != nil {
+		return err
+	}
+
+	env := environment.MustGet()
+	runningMode := env.RunningMode
+
+	if runningMode == environment.RunningMain {
+		if env.FarmID != 1 {
+			return nil
```
muhamadazmy: I understand that you wanna avoid having user configured for mainnet if it's not freefarm, but this also means if you return here that you are not going to process users configured in zos-config.

As explained in the issue the flow should be like this:
- You always apply config associated with the network (runningmode) in case of mainnet, this list in the config should be empty (but can have users in the future for any reason!) 
- So you don't return when runmode is main, you just keep processing, what u do is that u don't respect the `ssh-user` kernel cmdline argument for mainnet. So users can't grant themselves access to the node.
- If on mainnet, freefarm, we take the list of users for testnet.

cmds/identityd/main.go
```diff
@@ -139,14 +143,18 @@ func main() {
 		log.Info().Msg("received a termination signal")
 	})
 
+	err = manageSSHKeys(ctx)
+	if err != nil {
+		log.Fatal().Err(err).Send()
```
muhamadazmy: Instead of `Send()` please just use `Msg("failed to configure ssh users")` so it's clear what this error is related to.

Also, I don't believe failure to configure ssh user should be a `fatal` error. Instead log the error and continue. I mean it doesn's sound right that we fail the system because we can't configure ssh users. 

Since ssh users are mainly for debugging it is okay if they didn't deploy correctly, and that should be seen in the logs.

cmds/identityd/main.go
muhamadazmy: So a simple `http.Get()` to that url. then directly write then (after checking status codes) directly copying the Body of the response to the output file.

cmds/identityd/main.go
```diff
@@ -172,3 +180,56 @@ func getIdentityMgr(root string, debug bool) (pkg.IdentityManager, error) {
...
+	if found {
+		authorizedUsers = append(authorizedUsers, extraUser)
+	}
+
+	github := github.NewClient(nil)
+	var allKeys string
+
+	for _, user := range authorizedUsers {
+		if err := retry.Do(ctx, retry.WithMaxRetries(3, retry.NewConstant(1*time.Microsecond)), func(ctx context.Context) error {
+			keys, _, err := github.Users.ListKeys(ctx, user, nil)
```
muhamadazmy: Please use normal http.Get as i explained above. Then you can do `io.Copy` of the body of the response directly to the open file of authorized_keys. 

It's also important to handle the status code of the request correctly. Some error codes (say 404) means u shouldn't retry and can assume it's a permanent error.

cmds/identityd/main.go
```diff
@@ -172,3 +180,56 @@ func getIdentityMgr(root string, debug bool) (pkg.IdentityManager, error) {
...
+			if err != nil {
+				return err
+			}
+
+			for _, key := range keys {
+				allKeys = fmt.Sprintf("%s\n%s", allKeys, key.GetKey())
```
muhamadazmy: Using `http.Get` you can directly do an `io.Copy` but comment on this line here. This is not the best way to do this imho. appending strings like this is not the best performant. because of allocation and reallocation and moving of memory, etc.. What is better is 

- The best option is to do a `file.Write` directly for each key
- You could even have all keys appended to a list of keys then do strings.Join in one go, while this still have to allocate a lot of memory but at least it can be done onetime not with each sprintf

muhamadazmy (changes requested):

muhamadazmy (approved):

Issue #2236, opened 2024-03-14
automatic vmlinuz decompression for micro VMs


muhamadazmy:
Right now, flists that provide their own kernel has to have it already uncompressed [as per docs here](https://github.com/threefoldtech/zos/blob/main/docs/manual/zmachine/zmachine.md#vm) 

The idea is that while users usually forget about this, it also problematic if the kernel was updated then the VM restarted. This can brick the VM.

Instead zos should detect the kernel if it's compressed, and then does the uncompression automatically if needed (check the method in the docs above)

rawdaGastan:
WC:
- removed `lzo` pkg because it requires an extra lib to be installed
- 2 errors for `zstd` and `bzip` will be ignored
- other algorithms worked well
- uncompressed kernel will override the compressed one

WIP: 
- testing different kernels through deployments

rawdaGastan:
This part is removed now

PR #2252, opened 2024-04-01
uncompress kernel

rawdaGastan:
### Related Issues

- https://github.com/threefoldtech/zos/issues/2236


pkg/primitives/vm/decompress_kernel.go
```diff
@@ -0,0 +1,157 @@
...
+	"os"
+	"os/exec"
+
+	"github.com/cyberdelia/lzo"
+	"github.com/klauspost/compress/zstd"
+	"github.com/pierrec/lz4/v4"
+	"github.com/ulikunitz/xz"
+)
+
+func isValidELFKernel(KernelImagePath string) error {
```
xmonader, line 22: I think we can use debug/elf https://pkg.go.dev/debug/elf

pkg/primitives/vm/decompress_kernel.go
```diff
@@ -0,0 +1,219 @@
...
+func unZstd(kernelStream io.Reader) (io.Reader, error) {
+	r, err := zstd.NewReader(kernelStream)
+	if err != nil {
+		return nil, err
+	}
+
+	return r, nil
+}
+
+func tryDecompressKernel(KernelImagePath string) error {
```
muhamadazmy: why argument name is starting with an UpperCase ?

pkg/primitives/vm/decompress_kernel.go
```diff
@@ -0,0 +1,219 @@
...
+			name:           "zstd",
+		},
+	}
+
+	var errs error
+
+	for _, algo := range algos {
+		err = decompressData(kernelData, tmpFile, algo)
+		if err == nil {
+			f, err := os.Create(KernelImagePath)
```
muhamadazmy: you need to close this file

pkg/primitives/vm/decompress_kernel.go
```diff
@@ -0,0 +1,219 @@
...
+		return fmt.Errorf("kernel image is required")
+	}
+
+	// if kernel is already an elf (uncompressed)
+	if err := isValidELFKernel(KernelImagePath); err == nil {
+		return nil
+	}
+
+	// Prepare temp files:
+	tmpFile, err := os.CreateTemp("/tmp", "vmlinux-")
```
muhamadazmy: according to the docs. you should set `dir` (first argument) to empty string `""` to use the default tmp directory. It's better than setting it up explicitly imho

pkg/primitives/vm/decompress_kernel.go
```diff
@@ -0,0 +1,219 @@
...
+	if err != nil {
+		return err
+	}
+	defer tmpFile.Close()
+	defer os.Remove(tmpFile.Name())
```
muhamadazmy: The close has to happen first then deleting right ?
so this has to be in the opposite order. 

Or better 

```suggestion
        defer func() {
            tmpFile.Close()
            or.Remove(tmpFile.Name())
        }
```

pkg/primitives/vm/utils.go
```diff
@@ -148,6 +148,12 @@ func (p *Manager) prepContainer(
 
 	// can be overridden from the flist itself if exists
 	if len(imageInfo.KernelPath) != 0 {
+		// try to decompress kernel
+		err := tryDecompressKernel(imageInfo.KernelPath)
+		if err != nil {
```
muhamadazmy: Just some cosmetics but i prefer always to do:

```go 
if err := functionThatMightFail(); err != nil {
   // handle error
}
```
Go error handling is already annoying, so this makes it a little bit compact

pkg/primitives/vm/decompress_kernel.go
```diff
@@ -0,0 +1,219 @@
...
+		{
+			decompressFunc: unlzma,
+			headerBytes:    []byte("\135\000\000\000"),
+			name:           "unlzma",
+		},
+		// {
+		// 	decompressFunc: lZop,
+		// 	headerBytes:    []byte("\211\114\132"),
+		// 	name:           "lzop",
+		// },
```
muhamadazmy: This function should still be there, with an implementation that only return error. But we can then return a clear error that we don't support lzop

muhamadazmy (changes requested):

muhamadazmy (approved):

Issue #2219, opened 2024-03-05
More robust NTP server name resolution


scottyeager:
While investigating a [case](https://github.com/threefoldtech/zos/issues/2194) of farmer's nodes not working properly due to failure to reach NTP servers, I landed on an interesting [thread](https://bbs.archlinux.org/viewtopic.php?id=292688) discussing the root cause of the issue.

The summary is this:

* Zos resolves `[0-3].pool.ntp.org` to find the IP of an NTP server to use
* ntp.org is using geolocation on the source IP of these requests and mapping them to regional server pools
* The farmer with the issue is located in the Philippines, and the pool of servers provided to the farmer's nodes is polluted with faulty servers (this is a country specific pool that can be queried directly at `ph.pool.ntp.org`)
* Until the ntp.org pool is cleaned up, this farmer and potentially any others in the Philippines will be liable to have time sync issues with Zos

I think we can make some simple changes to improve the robustness of Zos in reaching a functional NTP server. Maybe a three tiered approach:

1. What we do now, query `[0-3].pool.ntp.org`
2. Assuming we have already completed geoip, query the appropriate [continent pool](https://www.ntppool.org/zone/@) based on the node's location. For example `[0-3].asia.pool.ntp.org`
3. Go round robin on the remaining continent pools, if the local one has failed

That said, I am just now reading this on the ntp.org [vendors page](https://www.ntppool.org/vendors.html):

>To allow you to use the pool as the default time service in your application, we will set you up with special hostnames, for example 0.vendor.pool.ntp.org, 1.vendor.pool.ntp.org, 2.vendor.pool.ntp.org and 3.vendor.pool.ntp.org.
>
> You **must absolutely not use the default pool.ntp.org zone names** as the default configuration in your application or appliance.

Pretty sure that's meant to include us, and it would explain why all the distros have their own subdomains (`arch.pool.ntp.org`, `ubuntu.pool.ntp.org`, etc). But I think it leaves us with the same issue, as these subdomains seem to resolve in the same way as the base domain.

I'm happy to pursue getting our own subdomain from ntp.org. I can also ask while I'm at it what their recommended approach regarding this issue would be.

muhamadazmy:
Hi @scottyeager 
  That's an amazing research. In fact we relay on the ntp pool to provide the proper (and local) address for the ntp server to contact. The idea is that reduces the latency and hence improves time syncing accuracy. 

[...]

PR #2260, opened 2024-04-14
add kernel flag to override ntp conf

Omarabdul3ziz:
### Description
ntp flag is a comma seperated list of ntp servers, which will override the content in /etc/ntp.conf

### Changes
ntp flag is a comma seperated list of ntp servers, which will override the content in /etc/ntp.conf

### Related Issues
- https://github.com/threefoldtech/zos/issues/2219

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/cmds/internet/main.go
+++ b/cmds/internet/main.go
@@ -219,15 +219,17 @@ func configureZOS() error {
 			}
 		}
 
-		if env.NTPServer != nil && *env.NTPServer != "" {
+		if env.NTPServer != nil && strings.TrimSpace(*env.NTPServer) != "" {
 			// override /etc/ntp.conf
 			servers := strings.Split(*env.NTPServer, ",")
 			var buffer strings.Builder
 			for _, server := range servers {
-				buffer.WriteString(fmt.Sprintf("server %s\n", server))
+				server = strings.TrimSpace(server)
+				if isValidNtp(server) {
+					buffer.WriteString(fmt.Sprintf("server %s\n", server))
+				}
 			}
-			err := os.WriteFile(ntpFilePath, []byte(buffer.String()), 0644)
-			if err != nil {
+			if err := os.WriteFile(ntpFilePath, []byte(buffer.String()), 0644); err != nil {
 				return errors.Wrap(err, "failed to write ntp server conf")
 			}
 		}
@@ -258,4 +260,8 @@ func configureZOS() error {
 	}
 
 	return backoff.RetryNotify(f, backoff.NewExponentialBackOff(), errHandler)
+}
+
+func isValidNtp(server string) bool {
+	return strings.HasSuffix(server, ".pool.ntp.org")
 }
--- a/cmds/internet/main.go
+++ b/cmds/internet/main.go
@@ -2,8 +2,10 @@ package main
 
 import (
 	"flag"
+	"fmt"
 	"net"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/cenkalti/backoff"
@@ -23,6 +25,8 @@ import (
 
 	"github.com/threefoldtech/zos/pkg/version"
 )
+
+const ntpFilePath = "/etc/ntp.conf"
 
 func main() {
 	app.Initialize()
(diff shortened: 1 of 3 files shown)


cmds/internet/main.go
```diff
@@ -215,6 +219,19 @@ func configureZOS() error {
...
+			servers := strings.Split(*env.NTPServer, ",")
+			var buffer strings.Builder
+			for _, server := range servers {
+				buffer.WriteString(fmt.Sprintf("server %s\n", server))
+			}
+			err := os.WriteFile(ntpFilePath, []byte(buffer.String()), 0644)
+			if err != nil {
+				return errors.Wrap(err, "failed to write ntp server conf")
+			}
+		}
```
xmonader: Always check with strings.TrimSpace(var) != ""

xmonader: these can be joined  `if err:= ... ; err != nil`

cmds/internet/main.go
```diff
@@ -215,6 +219,19 @@ func configureZOS() error {
 			}
 		}
 
+		if env.NTPServer != nil && *env.NTPServer != "" {
+			// override /etc/ntp.conf
+			servers := strings.Split(*env.NTPServer, ",")
+			var buffer strings.Builder
+			for _, server := range servers {
```
xmonader, line 226: No validation on the server variable before generating a broken ntp config file?

pkg/environment/environment.go
```diff
@@ -250,6 +253,12 @@ func getEnvironmentFromParams(params kernel.Params) (Environment, error) {
 		}
 	}
 
+	if ntps, ok := params.Get("ntp"); ok {
```
xmonader: Should we use GetOne?

qemu/README.md
```diff
@@ -115,9 +115,12 @@ sudo dnsmasq --strict-order \
...
 
-where `runmode` is one of `dev` , `test`  or `prod`, 
-      `farmer_id` is the id of the farm you registered with `tffarmer`
-      `ssh-user` is github username provided if a user need to pass ssh-key to the node
+### Flags and kernel args:
+
+- `runmode` is one of `dev` , `test`  or `prod`.
+- `farmer_id` is the id of the farm you registered with `tffarmer`
+- `ssh-user` is github username provided if a user need to pass ssh-key to the node
+- `ntp` a comma separated list with ntp servers to override `/etc/ntp.conf`
```
xmonader, line 123: Add an example

cmds/internet/main.go
```diff
@@ -215,6 +219,19 @@ func configureZOS() error {
...
+				buffer.WriteString(fmt.Sprintf("server %s\n", server))
```
xmonader: The server should have space trimmed, but it won't matter given the whitespace tolerance in the ntp file

xmonader (changes requested):

cmds/internet/main.go
Omarabdul3ziz: hmm other than not empty i think we can also validate if the server is suffixed with `.pool.ntp.org`?

muhamadazmy (changes requested):
Thanks @Omarabdul3ziz for your work. But (may be i should have said it earlier) I think the changes related to NTP should be part of 0-initramfs.

- First of all, ntp is not managed by zos repo, but it's part of 0-initramfs, it's fully installed and configured by this repo.
- The change definitely does not belong to `internet`. If you had doubts where you put the code you should have checked with me. Why internet should care about what ntp config is, right ?
- I noticed you don't even restart the service after updating it's config. Why do you think config changes will take effect. Note that, ntp is already running at this stage. If you check the ntp.yaml you see that it does not has any dependencies, hence it will get started (probably just give few errors until the node has internet access before it resumes)

Taking all of this into account what I think what should be done is the following:
- Changes has to be part of 0-initramfs 
- The changes are pretty simple, just modify the ntp.yaml file to do the following:
  - this has to be done in bash (more specifically `sh`)  
  - check if `ntp=` is passed as a cmdline argument `/proc/cmdline`
  - if set, create the file, with proper config (no validation since it's on the user now)
  - `exec ntpd -n` as last line of the script 

this will make sure that ntp is definitely configured before starting

Issue #2228, opened 2024-03-12
[update] if zos cache is read only because of bad disk, update with in memory cache


muhamadazmy:
The idea is that we still need to force the nodes to run latest code which might have a fix for the issues it's facing. this way we make sure all nodes are running latest code

PR #2261, opened 2024-04-15
Try upgrading in memory if there is an error with flist download

AbdelrahmanElawady:
### Description

Try upgrading in memory if there is an error with flist download

### Related Issues

- #2228

### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation

### DIFF
--- a/pkg/upgrade/upgrade.go
+++ b/pkg/upgrade/upgrade.go
@@ -294,25 +294,17 @@ func (u *Upgrader) updateTo(link hub.TagLink) error {
 	return nil
 }
 
-func (u *Upgrader) flistCache(inMemory ...bool) string {
-	root := u.root
-	if len(inMemory) > 0 && inMemory[0] {
-		root = filepath.Join("/tmp", root)
-	}
-	return filepath.Join(root, "cache", "flist")
+func (u *Upgrader) flistCache() string {
+	return filepath.Join(u.root, "cache", "flist")
 }
 
-func (u *Upgrader) fileCache(inMemory ...bool) string {
-	root := u.root
-	if len(inMemory) > 0 && inMemory[0] {
-		root = filepath.Join("/tmp", root)
-	}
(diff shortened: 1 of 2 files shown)


pkg/upgrade/upgrade.go
```diff
@@ -326,17 +334,33 @@ func (u *Upgrader) getFlist(repo, name string) (meta.Walker, error) {
 // install from a single flist.
 func (u *Upgrader) install(repo, name string) error {
 	log.Info().Str("repo", repo).Str("name", name).Msg("start installing package")
+	inMemory := false
 
 	store, err := u.getFlist(repo, name)
 	if err != nil {
-		return errors.Wrapf(err, "failed to process flist: %s/%s", repo, name)
+		// try in memory
```
muhamadazmy, line 369: I am wondering if handling ALL errors and do an in memory update is the right thing to do. I mean we already know if we have a cache problem (flags from storage are set if there is no cache and or cache is read only)

I think a better approach was to basically you check what is the status of system announced by storage, and then decide if u gonna do in-memory or normal update

muhamadazmy: I mean what if this is just a network issue, and nothing wrong with storage?

This also mean that the decision to use a in-memory update, or not has to be decided at the highest level (in the Run implementation to be exact) when making the call to update.

muhamadazmy: This also means, instead of passing around an optional `inMemory` flag, you can instead pass a `Cache` interface object (or any other name) that returns different passes where the files are stored (this can be point to persisted cache, or /tmp)

Why this is better than using a inMemory flag, simply because I don't have to do `if` every where since i can simply do `cache.Flist()` or `cache.File()` which returns the proper (created) paths.

muhamadazmy (changes requested):

muhamadazmy (approved):

Issue #2255, opened 2024-04-08
ZUI still shows errors even if healthcheck is okay


muhamadazmy:
## Describe the bug

The problem is, the checks are scheduled to run every 15 minutes. On boot when the network is not fully up some network checks fail and thus, shows error on ZUI.

Those error can get immediately resolved, but the UI will keep showing them until the next check. 

## To Reproduce

Start ZOS

## Suggested solution

What if, if network checks failed, that we keep running the checks with an exponential delay. Once resolved, we go back to the 15min schedule.

The perf framework does not support that of course so this has to be handled by the check loop itself.

PR #2262, opened 2024-04-16
add backoff retry on service check

Omarabdul3ziz:
### Description
add backoff retry on service check to pass time if the network is not setup yet

### Related Issues
- https://github.com/threefoldtech/zos/issues/2255

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/pkg/perf/healthcheck/healthcheck.go
+++ b/pkg/perf/healthcheck/healthcheck.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/cenkalti/backoff"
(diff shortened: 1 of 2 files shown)


muhamadazmy (changes requested):
This is not the right approach to fix this. Your solution will NOT show any error to the user until the issue is resolved.

The correct logic is that if we have errors, we need to show them to the user (via the PushErrors) then retry again

So while a backoff strategy is a good idea, this solution does not satisfy the requirements. You need to take another look

Hint: the retry has to be done on a higher layer

pkg/perf/healthcheck/healthcheck.go
```diff
@@ -60,21 +62,35 @@ func (h *healthcheckTask) Run(ctx context.Context) (interface{}, error) {
...
+	cl := perf.GetZbusClient(ctx)
+	zui := stubs.NewZUIStub(cl)
+
 	for label, check := range h.checks {
-		errors := check(ctx)
-		if len(errors) == 0 {
-			continue
+		op := func() error {
+			errors := check(ctx)
+			errs[label] = errorsToStrings(errors)
```
muhamadazmy: why the map? why do you need to use `errs` here this way?

pkg/perf/healthcheck/healthcheck.go
```diff
@@ -60,21 +62,35 @@ func (h *healthcheckTask) Run(ctx context.Context) (interface{}, error) {
...
+		bo := backoff.NewExponentialBackOff()
+		bo.InitialInterval = 30 * time.Second
+		bo.MaxInterval = 30 * time.Second
+		bo.MaxElapsedTime = 10 * time.Minute
 
-	for label := range h.checks {
-		err := zui.PushErrors(ctx, label, errs[label])
-		if err != nil {
+		if err := backoff.RetryNotify(op, bo, notify); err != nil {
 			return nil, err
```
muhamadazmy: I don't believe you should return here, since there are other checks you need to do.

pkg/perf/healthcheck/healthcheck.go
```diff
@@ -60,21 +62,35 @@ func (h *healthcheckTask) Run(ctx context.Context) (interface{}, error) {
...
+		notify := func(err error, t time.Duration) {
+			log.Error().Err(err).Str("check", label).Dur("retry-in", t).Msg("failed health check. retrying")
+		}
 
-	cl := perf.GetZbusClient(ctx)
-	zui := stubs.NewZUIStub(cl)
+		bo := backoff.NewExponentialBackOff()
+		bo.InitialInterval = 30 * time.Second
+		bo.MaxInterval = 30 * time.Second
+		bo.MaxElapsedTime = 10 * time.Minute
```
muhamadazmy: hmm, i am wondering if this is a good number. Note that this will block other checks from running. I am wondering if we should start all checks in parallel so they don't block each other (since they are not related) then each check can has it's own 10 min retry loop

muhamadazmy (changes requested):

pkg/perf/healthcheck/healthcheck.go
Omarabdul3ziz: errs are returned at the end as the result of the health check task, this is what is stored on redis and responded to the rmb call asking for the health check

muhamadazmy (approved):

Issue #2242, opened 2024-03-19
Node info need to automatically update if zos interface ip has changed


muhamadazmy:
## Describe the bug
This code exited before, to handle changes to ips assigned to zos interface and update the node information on chain. Seems this code was removed in an old refactor

ashraffouda:
verified

ashraffouda:
verified

ashraffouda:
verified

PR #2264, opened 2024-04-17
Handle changes to zos interface ip

Eslam-Nawara:
### Description

update networkd to update registration info on IP change.

### Changes

- update ZOSAddresses to send notification only on address update.
- update zui to load initial value of address before waiting for any updates.

### Related Issues

- #2242 

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/pkg/network/networker.go
+++ b/pkg/network/networker.go
@@ -257,6 +257,7 @@ func (n *networker) createMacVlan(iface string, master string, hw net.HardwareAd
 
 	if _, ok := err.(netlink.LinkNotFoundError); ok {
 		macVlan, err = macvlan.Create(iface, master, netNs)
+
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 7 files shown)


pkg/network/networker.go
```diff
@@ -1277,14 +1275,24 @@ func (n *networker) ZOSAddresses(ctx context.Context) <-chan pkg.NetlinkAddresse
 		return result
 	}
 
+	done := make(chan struct{})
+	updateChan := make(chan netlink.AddrUpdate)
+	err := netlink.AddrSubscribe(updateChan, done)
```
muhamadazmy: I think you can directly use `ctx.Done()`  here directly instead of creating a done channel that only closes when ctx is cancelled

muhamadazmy (changes requested):
Great work Eslam :) 

Have only small comment

muhamadazmy (approved):

Issue #2257, opened 2024-04-12
support `volume` primitive which is a subvolume with quota like a disk for workloads


muhamadazmy:
In the future, we gonna rely less on `zmount` (disks) and instead rely more on volumes. 

Under the hood, volumes are `btrfs` subvolumes, but this will probably change in the future to `bcachefs`. Over time, vdisks (currently called zmount) support will be dropped.

The reasons:
- Much better usage of caching provided by `bcachefs`
- Much better in utilising and sharing underlying storage.
- Normally better performance instead of sparse files that are stored on host file system
- Snapshots, which allow uploading backups of the volume without interrupting guest system

muhamadazmy:
@AbdelrahmanElawady you can start working on that one as follows:
- First start by defining the new workload type, and it's parameters.
- Workload result. is probably just the volume name/id 
- Extend the VM to support attaching "volumes" same way as we do with disks 
- Internally, VM will have to deploy after volumes (again, same with disks)
- During provision, vmd will take care of starting volume virtiofsd.

PR #2270, opened 2024-04-23
Add volume workload

AbdelrahmanElawady:
### Description

Add `volume` workload to allow the user to create volumes and attach them to zmachines, it should offer more performance, usability and easier upgrades to `bcachefs`

### Changes

- new `volume` workload
- it supports growing and shrinking the quota while the VM is running

### Related Issues

- #2257 

### Checklist

- [x] Tests included
- [x] Build pass
- [x] Documentation

### DIFF
--- a/pkg/primitives/volume/volume.go
+++ b/pkg/primitives/volume/volume.go
@@ -3,9 +3,7 @@ package volume
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
-	"os"
 
 	"github.com/threefoldtech/zbus"
 	"github.com/threefoldtech/zos/pkg/gridtypes"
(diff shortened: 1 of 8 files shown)


pkg/primitives/volume/volume.go
```diff
@@ -0,0 +1,99 @@
...
+		return nil, fmt.Errorf("failed to lookup volume with name %q: %w", volumeName, err)
+	} else if exists {
+		return VolumeResult{ID: volumeName}, provision.ErrNoActionNeeded
+	}
+
+	_, err = storage.VolumeCreate(ctx, volumeName, volume.Size)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create new volume with name %q: %w", volumeName, err)
+	}
+	return VolumeResult{ID: volumeName}, nil
```
muhamadazmy: This is of course correct, but I think it's safer to use the Name returned by the VolumeCreate. This ensure things will keep working if we changed the implementation of VolumeCreate for example to return a different name/id of the volume.

pkg/primitives/volume/volume.go
```diff
@@ -0,0 +1,99 @@
...
+	storage := stubs.NewStorageModuleStub(m.client)
+
+	var volume Volume
+	if err := json.Unmarshal(wl.Data, &volume); err != nil {
+		return nil, fmt.Errorf("failed to parse workload data as volume: %w", err)
+	}
+	volumeName := wl.ID.String()
+
+	vol, err := storage.VolumeLookup(ctx, volumeName)
+	if errors.Is(err, os.ErrNotExist) {
```
muhamadazmy: I am sure this will not work. And you know the reason. Remember that the error concrete type is lost of zbus. One of many reasons zbus needs fixing.

muhamadazmy: It's better to do exists first, then do lookup to get the volume, if lookup failed then it's an error that must be returned.

muhamadazmy (changes requested):
Thanks @AbdelrahmanElawady for amazing work. That's a very clean implementation. I only had 2 small comments.

muhamadazmy (approved):

Issue #2272, opened 2024-04-23
Clock skew in nodes


rawdaGastan:
## Describe the bug

a clock skew on node `4465` happened for around 30 minutes and ntpd gives up if the skew is too big

## Screenshots

![image](https://github.com/threefoldtech/zos/assets/47260239/0f4bb4d4-4fef-4115-b9c7-cbf58c51de82)

## solution

We should have some code to monitor time skew and if it's too big we just restart ntpd. Restarting ntpd forces it to resync even if the skew is huge

PR #2275, opened 2024-04-23
monitor ntp of the node

rawdaGastan:
### Issues

- https://github.com/threefoldtech/zos/issues/2272


cmds/modules/noded/main.go
```diff
@@ -113,6 +117,51 @@ func action(cli *cli.Context) error {
 		return nil
 	}
 
+	z := zinit.Default()
+	// monitor ntp
+	go func() {
```
muhamadazmy: - Why this has to be in `noded`
- Why this has to be in `main` like this. That's a LOT of logic in main function. If we keep doing this for each small monitoring we need to do then main will just become a soup of un-releated pieces of code

muhamadazmy: Please instead implement it as a `healthcheck`  https://github.com/threefoldtech/zos/blob/main/pkg/perf/healthcheck/healthcheck.go#L25

- This will allow u to run periodically
- You have time to take an action
- The state if the clock skew will be also collectable by users

muhamadazmy (changes requested):

pkg/perf/healthcheck/ntp.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func getCurrentLocalTime() (time.Time, error) {
+	timeRes, err := http.Get("https://worldtimeapi.org/api/timezone/UTC")
+	if err != nil {
+		return time.Time{}, errors.Wrapf(err, "failed to get date")
+	}
+
+	var result struct {
+		DateTime string `json:"datetime"`
```
muhamadazmy: I really think you can make this of type `time.Time` and it will automatically parse this from json for you. 

I create this example for you https://go.dev/play/p/BRuBs0VauD3

muhamadazmy: Also ZOS uses UTC not local time hence you should use the `utc_datetime`

muhamadazmy (changes requested):

pkg/perf/healthcheck/ntp.go
rawdaGastan: It gets the UTC time, so `utc_datetime` is the same as `datetime`

muhamadazmy (approved):

Issue #2263, opened 2024-04-17
[zui] on boot, zui takes really long time before it shows anything.


muhamadazmy:
## Description 
On boot, zui starts and you already can see screen has rendered black. But then it takes a lot of time before it shows anything. This is (i assume) duo to it trying to make zbus calls to other services to collect information but those services are not up yet.

What we need to do is:
- Make sure we should the ZUI interface. Show missing information as (loading) without blocking the entire interface.
- Show loading progress somehow ? maybe list of services and their status

Eslam-Nawara:
Work Completed:
- updated `zui` start immediately and not block until services are up.
- updated the start screen to show the status of some services until the services are active, then show the Network conf and system resources as normal.

ashraffouda:
verified

PR #2279, opened 2024-04-25
fix zui blocking untill services are up

Eslam-Nawara:
### Description

Fix `zui` to show the state of services while starting them instead of black screen.

### Related Issues
- #2263 

### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation


cmds/modules/zui/service.go
```diff
@@ -0,0 +1,152 @@
...
+	if _, err := register.NodeID(ctx); err != nil {
+		if isInProgressError(err) {
+			return InProgressStatus
+		} else {
+			return red(FailedStatus)
+		}
+	} else {
+		return green(activeStatus)
+	}
+}
```
ashraffouda, line 132: wouldn't be a bit clean if we removed those else statements? to be like this 
```
func getRegistrarStatus(ctx context.Context, client zbus.Client) string {
	register := stubs.NewRegistrarStub(client)
	if _, err := register.NodeID(ctx); err != nil {
		if isInProgressError(err) {
			return InProgressStatus
		}
		return red(FailedStatus)
	}
	return green(activeStatus)

}
```

ashraffouda (changes requested):

cmds/modules/zui/service.go
```diff
@@ -0,0 +1,149 @@
...
+		{"Noded", InactiveStatus},
+		{"Powerd", InactiveStatus},
+	}
+
+	render.Signal()
+
+	done := make(chan bool)
+
+	go func() {
+		networkStatus := make(chan string)
```
ashraffouda: isn't it better if we have only one chan that all services write to it and we check the returned string against which service is up or so?

ashraffouda (changes requested):

cmds/modules/zui/service.go
muhamadazmy: I agree

muhamadazmy (commented):
Really impressive work @Eslam-Nawara The behaviour of ZUI is much better now and i like how it looks as well.

I agree with Ashraf on his comment, please check it out so we can merge. What he basically suggest is we use a single channel that has status and service name not just a bool.

cmds/modules/zui/prov.go
```diff
@@ -45,11 +46,11 @@ func provRender(client zbus.Client, render *signalFlag, prov *widgets.Table) err
 
 	prov.Rows = [][]string{
 		{"", "Total", "Reserved"},
-		{"CRU", "", ""},
-		{"Memory", "", ""},
-		{"SSD", "", ""},
-		{"HDD", "", ""},
-		{"IPv4", "", ""},
+		{"CRU", "Loading...", "Loading..."},
```
xmonader: extract the `Loading ...` literal into a string and reuse it for consistency

xmonader (approved):

ashraffouda (approved):

Issue #2233, opened 2024-03-14
development tools to help building and debugging custom images


muhamadazmy:
The idea is that we need to have so development/debugging tools to help with building custom VM images (microvms) 
- I already created this script which allows you to boot a VM https://gist.github.com/muhamadazmy/a10bfb0cc77084c9b09dea5e49ec528e
- We also need to inject cidata (cloud-init seed) that include config to configure a simple network, and also add user key to the vm to make sure cloud-init is well configured

A-Harby:
- Missing dep installation: mtools and cargo.
- Installation happened in the /root.
-  Got this when trying to run with flist `https://hub.grid.tf/omarabdulaziz.3bot/ubuntu-jammy.flist`
```
Booting ...
+ boot
+ sudo cloud-hypervisor --cpus boot=1,max=1 --memory size=1024M,shared=on --kernel /tmp/overlay/kernel --initramfs /tmp/overlay/initramfs-linux.img --fs tag=vroot,socket=/tmp/virtiofs.sock --disk path=/tmp/cidata.img --cmdline 'rw console=ttyS0 reboot=k panic=1 root=vroot rootfstype=virtiofs rootdelay=30' --serial tty --console off
:: running hook [bootstrap]
mounting pts
ln: /dev/fd/fd: No such file or directory
trying to mount root
:: mounting 'vroot' on real root
:: trial '1'
:: running late hook [bootstrap]
settign up environment
setting up system
seed directory: /seed
root directory: /new_root
failed to apply network: failed to load network config file: failed to open file /seed/network-config: open /seed/network-config: no such file or directory
generating host ssh keys
PING 8.8.8.8 (8.8.8.8): 56 data bytes
ping: sendto: Network is unreachable
fill /etc/hosts
ERROR: Root device mounted successfully, but /sbin/init does not exist.
Bailing out, you are on your own. Good luck.

sh: can't access tty; job control turned off
[rootfs ]#
```

Omarabdul3ziz:
thanks ya harby, i will look at it

Omarabdul3ziz:
@A-Harby i couldn't know exactly what happened to make the booting stuck in the initramfs and failed to chroot to the new_root to continue the booting process, but mostly it is a wrong deps installation.
in the mentioned pr i separated the deps installtion on a file and add some missing deps.

on a new env setted up with the new `install_deps` script i managed to run a VM with the mentioned flist. could you try again by cleaning your env and setup with the script?

[...]

PR #2285, opened 2024-04-30
add debug-image script

Omarabdul3ziz:
### Description
add debug-image script

### Related Issues
- https://github.com/threefoldtech/zos/issues/2233
- https://github.com/threefoldtech/zos/issues/2290
- https://github.com/threefoldtech/zos/issues/2289

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/scripts/create_image.sh
+++ b/scripts/create_image.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+mkdir $1
+sudo debootstrap jammy $1  http://archive.ubuntu.com/ubuntu
+
+sudo arch-chroot $1 /bin/bash <<EOF
+
+export PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+rm /etc/resolv.conf
+echo 'nameserver 1.1.1.1' > /etc/resolv.conf
+
+apt-get update -y
+apt-get install -y cloud-init openssh-server curl
+cloud-init clean
+
+apt-get install -y linux-modules-extra-5.15.0-25-generic
+echo 'fs-virtiofs' >> /etc/initramfs-tools/modules
+update-initramfs -c -k all
+
+EOF
--- a/scripts/debug_image.sh
+++ b/scripts/debug_image.sh
@@ -1,67 +1,77 @@
 #!/bin/bash
-socket="/tmp/virtiofs.sock"
 
-rootfs=""
-kernel="$rootfs/boot/vmlinuz"
-initram="$rootfs/boot/initrd.img"
-cmdline="rw console=ttyS0 reboot=k panic=1 root=vroot rootfstype=virtiofs rootdelay=30"
-overlayfs="/tmp/merged"
-flist=""
+set -euo pipefail
 
+# constants
+readonly SOCKET="/tmp/virtiofs.sock"
+readonly OVERLAYFS="/tmp/merged"
+
+# defaults
+cidata="/tmp/cidata.img"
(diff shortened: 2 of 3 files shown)


xmonader (commented):
your PR is missing docs

muhamadazmy (changes requested):
The script looks fine in general but we need to make it a little bit more robus:
- Check other binaries if they are missing for example, i didn't have the mkdosfs.
- You need to auto decompress the kernel. you can simply use this script https://github.com/torvalds/linux/blob/master/scripts/extract-vmlinux
  - you can just copy the code from that one i guess
  - extracted kernel can be put under /tmp/ and used normally in the cloud-hypervisor command.

I was also thinking that we should not modify the `roofs` that was created by the user. Booting a vm will probably write files in that rootfs (for example cloud-init config). I was wondering if we can use overlay mount to create a separate rw layer on top of the rootfs directory. You gonna need this anyway if you using an flist mount as your rootfs.

The idea is you can create a directory under `/tmp` and then create an overlay mount where the rootfs is the lowerdir, then use the directory under temp for the upperdir. ant have a new mount that is used as your new rootfs. This means that changes done from within cloud-hypervisor (once you login) you can do changes that will not affect your actual rootfs.

scripts/create_image.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+mkdir $1
+sudo debootstrap jammy $1  http://archive.ubuntu.com/ubuntu
+
+sudo arch-chroot $1 /bin/bash <<EOF
+
+export PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+rm /etc/resolv.conf
```
xmonader, line 9: no need for the rm, in the next line you used `>` instead of `>>` so that should already overwrite the file

scripts/create_image.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
```
xmonader, line 2: always add `set -ex`

scripts/create_image.sh
```diff
@@ -0,0 +1,20 @@
...
+mkdir $1
```
xmonader, line 3: sanity check for the number of arguments and show a usage message if it's not the case

scripts/create_image.sh
```diff
@@ -0,0 +1,20 @@
...
+
+export PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+rm /etc/resolv.conf
+echo 'nameserver 1.1.1.1' > /etc/resolv.conf
+
+apt-get update -y
+apt-get install -y cloud-init openssh-server curl
+cloud-init clean
+
+apt-get install -y linux-modules-extra-5.15.0-25-generic
```
xmonader, line 16: ain't that quite outdated kernel?

scripts/debug_image.sh
```diff
@@ -0,0 +1,191 @@
+#!/bin/bash
+
```
xmonader, line 2: `set -ex`

scripts/create_image.sh
muhamadazmy: I think we have the tools now to try a newer version and immediately try it out see if it works

muhamadazmy (dismissed):
This looks really nice!

xmonader (approved):
احسنت يابني

Issue #2295, opened 2024-05-07
expand config and environment docs and tests


xmonader:
https://github.com/threefoldtech/zos/blob/1fb8d7923cdd1401f4d1972ac84c354136dd90c3/pkg/environment

PR #2326, opened 2024-05-12
Add environment pkg docs

Eslam-Nawara:
### Description

Document environment package.

### Related Issues

- #2295 

### Checklist

- [ ] Tests included
- [ ] Build pass
- [x] Documentation

### DIFF
--- a/pkg/environment/README.md
+++ b/pkg/environment/README.md
@@ -1,0 +1,35 @@
+# Environment module
+
+This module is used to get information about the running environment of a node.
+
+It is also used to get the organization default configurations defined [here](https://github.com/threefoldtech/zos-config)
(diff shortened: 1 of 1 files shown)


pkg/environment/README.md
```diff
@@ -0,0 +1,35 @@
+# Environment module
+
+This module is used to get information about the running environment of a node.
```
xmonader, line 3: is that os.Getenv or that's the kernel passed arguments or what specifically?

Eslam-Nawara: It defines different constants based on the current running environment, but it uses both to get them.

muhamadazmy (approved):

Issue #2338, opened 2024-05-22
Add mycelium support to zos


ashraffouda:
we need to run myceilum instance for zos to be able to download flists overy myceilum network

Eslam-Nawara:
### WIP: 
- Integrated mycelium in zos host.
- Support serving flists over mycelium.
- Add mycelium support for zdbs with backward compatibility.
- Tested mycelium on the host, and connecting zdbs over mycelium.

> Need to test the update against node with public namespace.

A-Harby:
Verified, devnet.

I was able to deploy ZDB using ts client and connect to mycelium ip.
![image](https://github.com/threefoldtech/zos/assets/109473873/6a3cb6ef-7bf7-4471-a99b-5bf5672be389)
![image](https://github.com/threefoldtech/zos/assets/109473873/3741c42d-a8ac-4551-b0c2-1a3a4ef92eb5)

PR #2342, opened 2024-05-23
Integrate mycelium in zos

Eslam-Nawara:
### Description

Integrate mycelium in zos and allow flists to be served over mycelium

### Changes

Add mycelium to networkd

### Related Issues

- #2338 

### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation


xmonader:
don't forget to add an ADR please

pkg/flist/flist.go
```diff
@@ -190,13 +193,13 @@ func New(root string, storage *stubs.StorageModuleStub) pkg.Flister {
...
 // TODO: how to know that this ro mount is no longer used, hence can be unmounted and cleaned up?
-func (f *flistModule) mountRO(url, storage string) (string, error) {
+func (f *flistModule) mountRO(url, storage, namespace string) (string, error) {
 	// this should return always the flist mountpoint. which is used
 	// as a base for all RW mounts.
 	sublog := log.With().Str("url", url).Str("storage", storage).Logger()
 	sublog.Info().Msg("request to mount flist")
 
-	hash, flistPath, err := f.downloadFlist(url)
+	hash, flistPath, err := f.downloadFlist(url, namespace)
```
muhamadazmy: You probably need to use this same namespace at line `262` below

cmds/modules/networkd/main.go
```diff
@@ -126,6 +128,17 @@ func action(cli *cli.Context) error {
 		return errors.Wrap(err, "failed to start yggdrasil")
 	}
 
+	log.Debug().Msg("starting mycelium")
```
xmonader, line 126: I think this and ygg code can be extracted into their own functions and same for the DMZ

cmds/modules/networkd/main.go
```diff
@@ -148,9 +161,34 @@ func action(cli *cli.Context) error {
 		if err := dmzYgg.SetYggIP(ip, gw.IP); err != nil {
 			return errors.Wrap(err, "failed to set yggdrasil ip for dmz")
 		}
+
+		// set up mycelium in ndmz
+		dmzMy, err := mycelium.NewMyNamespace(dmz.Namespace())
```
xmonader: can be extracted into another function

docs/architecture/decisions/0010-mycelium.md
```diff
@@ -0,0 +1,20 @@
...
+
+Accepted
+
+## Context
+
+Support mycelium network for zdbs and zos host to serve flists
+
+## Decision
+
+Integrate mycelium in zos and allow zos host to have mycelium IPs, serve flists over mycelium, and support mycelium on zdbs
```
xmonader: is it serve or mount flists over mycelium?

pkg/network/types/constant.go
```diff
@@ -11,8 +11,10 @@ const (
...
 	// by the bootstrap of networkd
 	DefaultBridge = "zos"
-	//PublicBridge name
+	// PublicBridge name
 	PublicBridge = "br-pub"
-	//YggBridge ygg bridge
+	// YggBridge ygg bridge
 	YggBridge = "br-ygg"
+	// MyBridge mycelium bridge
+	MyBridge = "br-my"
```
xmonader: MyCeliumBridge

pkg/network/networker.go
```diff
@@ -192,6 +196,48 @@ func (n *networker) detachYgg(id string, netNs ns.NetNS) error {
...
+	// new hardware address for mycelium interface
+	hw := ifaceutil.HardwareAddrFromInputBytes([]byte("my:" + id))
+
+	myc, err := n.myc.InspectMycelium()
+	if err != nil {
+		return net.IPNet{}, err
+	}
+
+	ip, err := myc.SubnetFor(hw)
+	log.Info().Msgf("myc subnet is %s", ip)
```
xmonader: myc -> mycelium

pkg/network/networker.go
```diff
@@ -114,13 +117,14 @@ func NewNetworker(identity *stubs.IdentityManagerStub, ndmz ndmz.DMZ, ygg *yggdr
 		portSet:        set.NewInt(),
 
 		ygg:  ygg,
+		myc:  myc,
```
xmonader: mycelium:

pkg/network/mycelium/mycelium.go
```diff
@@ -0,0 +1,235 @@
...
+
+	f, err := os.Create(path)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	key := x25519.EdPrivateKeyToX25519([]byte(privateKey))
+
+	return json.NewEncoder(f).Encode(key)
```
muhamadazmy: Why are u using `json` here. the priv key of mycelium is not json encoded. It's a binary file. i don't believe this file is loaded by mycelium and is either ignored is overridden.

If this ever worked, then i assume it's because mycelium just read key length bytes from the file and used that as seed for the key!

This is definitely an issue, because it means the key u generated here is not secure. because it's not the full "Random" key but only part of the key json encoded bytes.

Do you get what I mean ?

pkg/network/mycelium/utils.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"context"
+	"crypto/ed25519"
+
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog/log"
+	"github.com/threefoldtech/zos/pkg/zinit"
+)
+
+func EnsureMycelium(ctx context.Context, privateKey ed25519.PrivateKey, ns MyceliumNamespace) (*MyServer, error) {
```
xmonader: move this into mycelium.go and remove utils file

pkg/network/mycelium/config.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/decred/base58"
+	"github.com/pkg/errors"
+)
+
+type NodeConfig struct {
+	KeyFile    string
+	TunName    string
+	Peers      []string
+	PrivateKey ed25519.PrivateKey
+	PublicKey  string
```
muhamadazmy: Why is this needed?

pkg/network/mycelium/config.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/decred/base58"
+	"github.com/pkg/errors"
+)
+
+type NodeConfig struct {
+	KeyFile    string
+	TunName    string
+	Peers      []string
+	PrivateKey ed25519.PrivateKey
```
muhamadazmy: Why do you keep the private key as ed25519 while we need x25519. I think it's better to convert the key before storing it on the NodeConfig from the start instead of changing it on start

pkg/network/mycelium/config.go
```diff
@@ -0,0 +1,56 @@
...
+	PublicKey  string
+	NodeInfo   map[string]interface{}
```
muhamadazmy: What is this and what is it used for?

We don't need to copy everything over from ygg. Ygg has a completely different configuration and even requires a config file with way more config than mycelium.

Please drop all attributes that are not needed.

pkg/network/mycelium/mycelium.go
```diff
@@ -0,0 +1,235 @@
...
+	"github.com/threefoldtech/zos/pkg/network/namespace"
+	"github.com/threefoldtech/zos/pkg/zinit"
+)
+
+const (
+	tunName      = "my0"
+	myBin        = "mycelium"
+	zinitService = "mycelium"
+	confPath     = "/tmp/mycelium_priv_key.bin"
+	MyListenTCP  = 9651
```
muhamadazmy, line 28: you will have to add this to the list of reserved ports of the system

pkg/network/mycelium/mycelium.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+// NodeID returns mycelium node ID of s
+func (s *MyServer) NodeID() (ed25519.PublicKey, error) {
+	if s.cfg.PublicKey == "" {
+		panic("EncryptionPublicKey empty")
+	}
+
+	return hex.DecodeString(s.cfg.PublicKey)
+}
```
muhamadazmy: Not needed

pkg/network/mycelium/mycelium.go
```diff
@@ -0,0 +1,235 @@
...
+	if err != nil {
+		return inspection, err
+	}
+
+	bin, err := exec.LookPath(myBin)
+	if err != nil {
+		return inspection, err
+	}
+
+	cmd := fmt.Sprintf(`exec ip netns exec %s %s inspect --json --key-file %s `, s.ns, bin, confPath)
```
muhamadazmy: Any idea why your command is a `sh` then u do exec and not the command directly!
I mean why not do 
```
os.Command("ip", "netns", "exec", s.ns,  bin, "inspect", "--json", "--key-file", confPath)
```
?!

pkg/network/mycelium/mycelium.go
```diff
@@ -0,0 +1,235 @@
...
+	cfg *NodeConfig
+	ns  string
+}
+
+type MyceliumInspection struct {
+	PublicKey string `json:"publicKey"`
+	Address   net.IP `json:"address"`
+}
+
+type Subnet [8]byte
```
muhamadazmy: this is not used

pkg/network/mycelium/mycelium.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+// Tun return the name of the TUN interface created by mycelium
+func (s *MyServer) Tun() string {
+	return s.cfg.TunName
+}
+
+// SubnetFor return an IP address out of the node allocated subnet by hasing b and using it
+// to generate the last 64 bits of the IPV6 address
+func (m *MyceliumInspection) SubnetFor(b []byte) (net.IPNet, error) {
```
muhamadazmy: I really think this should be called `IPFor` not `SubnetFor` even if it's called like that in ygg it's still wrong.

muhamadazmy (changes requested):

pkg/network/mycelium/mycelium.go
Eslam-Nawara: it's added to reserved ports [here](https://github.com/threefoldtech/zos/pull/2342/files/29253a863afba63ab9b040e4e464e26a970bf775#diff-4e15b6f2760eb52c8292eadf92ce4a6ba3afcf08d34b528d56081b9bc1623554R127)

muhamadazmy (approved):

Issue #2346, opened 2024-05-29
openRPC based APIs


xmonader:
the plan would be exposing zos APIs over mycelium IP, Let's start by creating openRPC documentation, underlying handlers won't need to be changed 

KDS is suggesting documenting using this https://pypi.org/project/tabella/

ashraffouda:
to have openrpc for zos api we need some modification to myceilum for user verification so it would be better if the user starts myceilum with his mnemonics so it has the same identity on the chain and from that we can check the sender ip and using myceilum we can get the key hence we can look up his twin with that key from the chain so we verify that it is the same user. so what needed from myceilum 
1- user start myceilum with his chain mnemonics so we can derive the same key
2- from ip we can get the address (pk)

ashraffouda:
so far I checked those pkgs 
1- https://github.com/zchee/go-openrpc
this repo doesn't have a working code. all commits seems like initial commits and not maintaind and is like 5 years old

2- https://github.com/gregdhill/go-openrpc
this one is working I did a hello world example but also not maintained and is 2 years old also using deprecated pkgs like packr but this is not a big deal we can either use it as it is or maybe later we update it those deprecated stuff is not used alot in the code
this one starts from api.json file and it will generate the server stubs, for client generation in go or in any other language we need to use different libraries 
3- https://github.com/etclabscore/go-openrpc-reflect
 I didn't do hands on on it but the difference between it and the second one is this one starts from the code and tries to reflect the server code and generate the json spec file but this one expecting to follow some code styles also not sure if this will be working with all our types in zos or not

p.s all the above pkgs doesn't provide client code generation, so whatever we will choose from the above we will relay on an external openrpc client generation tool

Omarabdul3ziz:
Update:

to simplify things here is what we need: 
- a server code: using the jsonrpc protocol (should be written in go)
- an openrpc doc: a document describing the jsonrpc API 
- a client code: uses the jsonrpc protocol to call the API (should be in different langs: go, ts, v, ...)

as an implementation, i am thinking of:
- Server could be implemented by the `net/rpc` package it should have all that we need. i already created a working example with it.
- Docs file should be auto-generated from the server, according to the openrpc this could done by registering an `rpc.discover` method to the server, after some searching the only tool that can do this is https://github.com/etclabscore/go-openrpc-reflect i tried it and it is compatible with the servers from `net/rpc` i managed to create a simple doc with it. but unfortunately, it has some problems and limitations:
[...]

PR #2371, opened 2024-07-02
openrpc based api

Omarabdul3ziz:
### Pr contains
- openrpc spec file that represents all zos endpoints and their types.
- a simple tool that generates go code based on this specs
- a service that runs jsonrpc protocol serving all zos endpoints using `net/rpc` package

### Internal modifications:
- seperate the perf task getters to have a method for each task with a
  defined result type instead of interface{}
- modify healthcheck task to have its result as slice of object
  containes the check name instead of map from name to error
- modify public ip task to have its result as slice of reports with the
  ip included instead of map from the ip to report
- update performance monitor stub
- modify the diagnositcs response to be slice of modules status instead
  of map

### Related issues
- https://github.com/threefoldtech/zos/issues/2346


Omarabdul3ziz:
close for https://github.com/threefoldtech/zosbase/pull/26

PR #2395, opened 2024-08-13
storage: concurrent disk image write.

### LINKED ISSUE #2391: faster OS image download
## Is your feature request related to a problem? Please describe

The full VM OS image download is slow from Australia, it took around 20 mins

[Feature Request: Faster or Local OS load for (big) images](https://git.ourworld.tf/tfgrid/circle_engineering/issues/33)

https://github.com/threefoldtech/test_feedback/issues/419

## Describe the solution you'd like

There are two kind of solutions:
1. from hub side
   It is already described on https://git.ourworld.tf/tfgrid/circle_engineering/issues/33: create replica of flist hub that are closer to the users

2.  speedup the download from `zos` side
    `rfs` download the block on `read` operation. Because the download triggered by `io.Copy` (https://github.com/threefoldtech/zos/blob/0ea61706e1a501d4e774a9195c139e2995bdd1cb/pkg/storage/disk.go#L148-L152), the flow become something like this:
        - OS try to read some block
        - `rfs` call  remote `hub` to download the block
        - repeat until finish
       The process happens sequentially, hence very slow

    Some alternatives:
         a. **Do the download on `open`, it is what happens in `0-fs`**
             IMO It is quite a big change from architecture point of view, i currently don't know the impact of this change
         b. **paralelize the `io.Copy`**
             It is a simple solution, but if the slowness happen on other usecase, we have to apply the same technique, which might be error prone.
             
While making hub replica would help to speedup the issue, i think solution 2.b also worth to do because it would further speedup the process and also quite easy to do.

iwanbk:
Write it concurrently to speed it up from the previous sequential write.

### Description

Change image DiskWrite from sequential to concurrent, to make it faster.

### Changes

- change to concurrent write

### Related Issues

Fixes :
- #2391 
- #2405

### Checklist

- [x] Tests included -> manual test
- [x] Build pass
- [x] Documentation

### DIFF
--- a/pkg/storage/disk.go
+++ b/pkg/storage/disk.go
@@ -14,6 +14,7 @@ import (
 	log "github.com/rs/zerolog/log"
 	"github.com/threefoldtech/zos/pkg"
 	"github.com/threefoldtech/zos/pkg/gridtypes"
+	"golang.org/x/sync/errgroup"
 )
 
 const (
@@ -145,10 +146,35 @@ func (s *Module) DiskWrite(name string, image string) error {
 		return fmt.Errorf("image size is bigger than disk")
 	}
 
-	_, err = io.Copy(file, source)
-	if err != nil {
+	// writing the image concurrently to speedup the previous sequential write.
+	// the sequential write is slow because the data source is from the remote server.
+	var (
+		// use errgroup because there is no point in continuing if one of the goroutines failed
+		group               = new(errgroup.Group)
+		concurrentNum int   = 5
+		imgSize       int64 = imgStat.Size()
+		chunkSize           = imgSize / int64(concurrentNum)
+	)
+
+	log.Info().Int("concurrentNum", concurrentNum).Msg("writing image concurrently")
+	for i := 0; i < concurrentNum; i++ {
+		index := i
+		group.Go(func() error {
+			start := chunkSize * int64(index)
+			len := chunkSize
+			if index == concurrentNum-1 { //last chunk
+				len = imgSize - start
+			}
+			wr := io.NewOffsetWriter(file, start)
+			rd := io.NewSectionReader(source, start, len)
+			_, err = io.Copy(wr, rd)
+			return err
+		})
+	}
+	if err := group.Wait(); err != nil {
 		return errors.Wrap(err, "failed to write disk image")
 	}
+	log.Info().Msg("writing image finished")
 
 	return nil
 }
--- a/pkg/storage/disk.go
+++ b/pkg/storage/disk.go
@@ -150,19 +150,19 @@ func (s *Module) DiskWrite(name string, image string) error {
 	// the sequential write is slow because the data source is from the remote server.
 	var (
 		// use errgroup because there is no point in continuing if one of the goroutines failed
-		group               = new(errgroup.Group)
-		concurrentNum int   = 5
-		imgSize       int64 = imgStat.Size()
-		chunkSize           = imgSize / int64(concurrentNum)
+		group            = new(errgroup.Group)
+		numWorkers int   = 5
+		imgSize    int64 = imgStat.Size()
+		chunkSize        = imgSize / int64(numWorkers)
 	)
 
-	log.Info().Int("concurrentNum", concurrentNum).Msg("writing image concurrently")
-	for i := 0; i < concurrentNum; i++ {
+	log.Info().Int("numWorkers", numWorkers).Msg("writing image concurrently")
+	for i := 0; i < numWorkers; i++ {
 		index := i
 		group.Go(func() error {
 			start := chunkSize * int64(index)
 			len := chunkSize
-			if index == concurrentNum-1 { //last chunk
+			if index == numWorkers-1 { //last chunk
 				len = imgSize - start
 			}
 			wr := io.NewOffsetWriter(file, start)
--- a/pkg/storage/disk.go
+++ b/pkg/storage/disk.go
@@ -14,6 +14,7 @@ import (
 	log "github.com/rs/zerolog/log"
 	"github.com/threefoldtech/zos/pkg"
 	"github.com/threefoldtech/zos/pkg/gridtypes"
+	"github.com/threefoldtech/zos/pkg/gridtypes/zos"
 	"golang.org/x/sync/errgroup"
 )
 
@@ -146,6 +147,23 @@ func (s *Module) DiskWrite(name string, image string) error {
 		return fmt.Errorf("image size is bigger than disk")
 	}
 
+	// do sequential copy on HDD only node
+	// otherwise do concurrent copy
+	if n, _ := s.Total(zos.SSDDevice); n == uint64(0) {
+		_, err = io.Copy(file, source)
+
+	} else {
+		err = s.concurrentCopy(file, source, imgStat)
+	}
+
+	if err != nil {
+		return errors.Wrap(err, "failed to write disk image")
+	}
+
+	return nil
+}
+
+func (s *Module) concurrentCopy(dst io.WriterAt, src io.ReaderAt, imgStat os.FileInfo) error {
 	// writing the image concurrently to speedup the previous sequential write.
 	// the sequential write is slow because the data source is from the remote server.
 	var (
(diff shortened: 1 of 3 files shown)


muhamadazmy (commented):
In general, this looks good. But I would love to see a benchmarking that compares the performance of the 2 approaches.

iwanbk:
@muhamadazmy 
The benchmarking already in the linked issue.
Including here because here is indeed the better place

I've tried to paralelize the io.Copy with this environment:
- zos node: In Indonesia, run on my old 2016 PC on qemu. (indonesia is close to Australia)
- os image: nixos

the results:

- original io.Copy: 1 hour (much worse than the reported 20 mins :))
- 5 goroutines: 20 mins
- 10 goroutines: 20 mins
- 15 goroutines: 24 mins

muhamadazmy:
That's cool.

One thing i had to add. Flists with raw images are obsolete and should not be used anymore. Zos still supports it for backwards compatibility only but definitely new workloads with this kind of images should not be allowed.

muhamadazmy:
Forgot to say that make sure to clean up the cache between the benchmarks runs. Since rfs caches the downloaded content in zos-cache. Means that second run will always go faster than first run since it doesn't have to download the image again

iwanbk:
> Forgot to say that make sure to clean up the cache between the benchmarks runs. Since rfs caches the downloaded content in zos-cache. Means that second run will always go faster than first run since it doesn't have to download the image again

sure, it only took ~2 mins with the cache.

iwanbk:
actually i not only deleted the cache, but delete all the qemu cow disks because i don't know how to delete it.
I guess it is under `/var/cache/modules/flistd`?
```bash
var/cache/modules/flistd # ls
cache       flist       log         mountpoint  pid         ro
```

pkg/storage/disk.go
```diff
@@ -145,10 +146,35 @@ func (s *Module) DiskWrite(name string, image string) error {
...
 	}
 
-	_, err = io.Copy(file, source)
-	if err != nil {
+	// writing the image concurrently to speedup the previous sequential write.
+	// the sequential write is slow because the data source is from the remote server.
+	var (
+		// use errgroup because there is no point in continuing if one of the goroutines failed
+		group               = new(errgroup.Group)
+		concurrentNum int   = 5
```
xmonader: change to numWorkers

pkg/storage/disk.go
```diff
@@ -145,10 +146,35 @@ func (s *Module) DiskWrite(name string, image string) error {
 		return fmt.Errorf("image size is bigger than disk")
 	}
 
-	_, err = io.Copy(file, source)
-	if err != nil {
+	// writing the image concurrently to speedup the previous sequential write.
+	// the sequential write is slow because the data source is from the remote server.
+	var (
+		// use errgroup because there is no point in continuing if one of the goroutines failed
+		group               = new(errgroup.Group)
```
xmonader: should be usable on the zero value or with a context if required

pkg/storage/disk.go
xmonader: how was the 5 choosen?

pkg/storage/disk.go
```diff
@@ -145,10 +146,35 @@ func (s *Module) DiskWrite(name string, image string) error {
...
+		index := i
+		group.Go(func() error {
+			start := chunkSize * int64(index)
+			len := chunkSize
+			if index == concurrentNum-1 { //last chunk
+				len = imgSize - start
+			}
+			wr := io.NewOffsetWriter(file, start)
+			rd := io.NewSectionReader(source, start, len)
+			_, err = io.Copy(wr, rd)
```
xmonader: I believe using io.CopyBuffer could avoid temporal buffers https://pkg.go.dev/io#CopyBuffer

xmonader: also, no need to call file.Sync?

pkg/storage/disk.go
iwanbk: > should be usable on the zero value

could you elaborate more on this?

> or with a context if required

not possible for now:
- the func doesn't have context
- `io.Copy` itself doesn't have context.

I think make it context aware will need separate ticket

pkg/storage/disk.go
iwanbk: > how was the 5 choosen?

see https://github.com/threefoldtech/zos/pull/2395#issuecomment-2285301322

xmonader:
All in all looks good, also I wonder about if that code path should be enabled in case of HDD only nodes?

pkg/storage/disk.go
iwanbk: > I believe using io.CopyBuffer could avoid temporal buffers https://pkg.go.dev/io#CopyBuffer

the difference between `io.Copy` and `CopyBuffer` is : CopyBuffer lets us choose our own buffer.
`io.Copy` will also use buffer.
The slowness here is because of downloading the file from the `hub`, so i don't think that the buffer will make big differences here.

iwanbk: > also, no need to call file.Sync?

i don't think so, we're copying to `rfs` anyway.
CMIIW

pkg/storage/disk.go
iwanbk: > I think make it context aware will need separate ticket

`DiskWrite` handled by `storaged`, called by `provisiond` through `zbus`.
So, plain Go context will not really work here.
But of course we can start with `context` inside `storaged`.

As a side note, i noticed that `DiskWrite` keep going even tho `provisiond` already cancel the deployment/contract.
So, we already have real cancellation issue over zbus.

iwanbk:
> All in all looks good, also I wonder about if that code path should be enabled in case of HDD only nodes?

can you elaborate more on this?
Is it because HDD only node will be slow?
And how is `provisiond` behavior regarding this?

pkg/storage/disk.go
xmonader: can you please mention the cancellation problem as a known issue on the top of PR to be easily found later on

pkg/storage/disk.go
xmonader: io.CopyBuffer uses a buffer your provide instead of io.Copy keeping allocating them during every call (it's still an io to save imo)

xmonader:
> > All in all looks good, also I wonder about if that code path should be enabled in case of HDD only nodes?
> 
> can you elaborate more on this? Is it because HDD only node will be slow? And how is `provisiond` behavior regarding this?

zos right now supports also HDD nodes, I'm believe sequential nature of HDD would have the performance impacted (worse) with the concurrency

xmonader:
on another note I'd also add  the concept of retries to the code if possible

pkg/storage/disk.go
iwanbk: > change to numWorkers

done

pkg/storage/disk.go
iwanbk: > can you please mention the cancellation problem as a known issue on the top of PR to be easily found later on

Maybe create separate issue because it is not really specific to this PR

pkg/storage/disk.go
iwanbk: > instead of io.Copy keeping allocating them during every call 

No, the buffer used by `io.Copy` also only allocated once.
See https://github.com/golang/go/blob/master/src/io/io.go#L417-L427


if what you mean is creating global buffer, outside the `DiskWrite` func then it would be lot of works for very small benefit.

iwanbk:
> zos right now supports also HDD nodes, I'm believe sequential nature of HDD would have the performance impacted (worse) with the concurrency

~~This is true if we write to to regular file, but this is not the case here:~~
1. ~~the destination is `rfs`, and `rfs` doesn't the image in a single regular file.~~
2. also need to be aware that the slowness is on the downloading side (from hub) , not on the write side to the disk image

As a side note, current code is favoring SSD disks

iwanbk:
> on another note I'd also add the concept of retries to the code if possible

why not handle it inside `rfs`?

iwanbk:
> zos right now supports also HDD nodes, I'm believe sequential nature of HDD would have the performance impacted (worse) with the concurrency

I assume it won't happen because the slowness from remote `rfs` will cover it.
But because it is hard to test, i think disable it on HDD-only is safer.

fixed it on caf83ac67e274e6ea04e640701eb7506fd8c500d @xmonader

PR #2397, opened 2024-08-14
docs for network light

ashraffouda:
Signed-off-by: Ashraf Fouda <ashraf.m.fouda@gmail.com>### Description

Describe the changes introduced by this PR and what does it affect

### Changes

List of changes this PR includes

### Related Issues

List of related issues

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


bootstrap/bootstrap/src/config.rs
```diff
@@ -105,6 +114,12 @@ impl Config {
                     .takes_value(false)
                     .help("run in debug mode, will use the bootstrap:development.flist"),
             )
+            .arg(
+                Arg::with_name("light")
+                    .short("l")
+                    .takes_value(false)
+                    .help("run in light mode, will use the bootstrap:development.flist"),
+            )
```
muhamadazmy: The light flag same as the runmode should be parsed from the kernel cmdline arguments not from the bootstrap command arguments.

cmds/modules/netlightd/nft.go
```diff
@@ -16,6 +16,7 @@ func ensureHostFw(ctx context.Context) error {
 nft 'add table inet filter'
 nft 'add table arp filter'
 nft 'add table bridge filter'
+nft 'add table nat'
```
muhamadazmy: I really think this entire file can completely replaced now with a `.nft` file basically rewrite this as a ruleset (similar to what we have under nft/rules.nft) 

It's way cleaner than executing individual commands

cmds/modules/provisiond/reporter.go
```diff
@@ -166,87 +166,16 @@ func (r *Reporter) getVmMetrics(ctx context.Context, slot rrd.Slot) error {
...
-			return errors.Wrapf(err, "failed to store metrics for '%s'", wl)
-		}
-	}
-
-	return nil
-}
-
-// getVmMetrics will collect network consumption every 5 min and store
-// it in the rrd database.
-func (r *Reporter) getGwMetrics(ctx context.Context, slot rrd.Slot) error {
```
muhamadazmy, line 194: This will probably be re-added

etc/zinit/networkd.yaml
```diff
@@ -1,4 +1,5 @@
-exec: networkd --broker unix:///var/run/redis.sock --root /var/cache/modules/networkd
-test: zbusdebug --module network
+exec: netlightd --broker unix:///var/run/redis.sock --root /var/cache/modules/networkd
+
+# test: zbusdebug --module network
```
muhamadazmy: We still have to have something here to let other modules know that this networking is ready. I am sure netlightd also has a "zbus" service which means we can use the same test command but with different module name (if the module name was changed even)

pkg/environment/environment.go
```diff
@@ -321,6 +321,10 @@ func getEnvironmentFromParams(params kernel.Params) (Environment, error) {
 	if e := os.Getenv("ZOS_BIN_REPO"); e != "" {
 		env.BinRepo = e
 	}
+	env.Light = false
```
muhamadazmy: This of course is not needed, since the zero value for bool in Go is false.

pkg/gridtypes/zos/types.go
```diff
@@ -12,10 +12,14 @@ const (
 	ZMountType gridtypes.WorkloadType = "zmount"
 	// NetworkType type
 	NetworkType gridtypes.WorkloadType = "network"
-	// ZDBType type
-	ZDBType gridtypes.WorkloadType = "zdb"
+	// NetworkLightType type
+	NetworkLightType gridtypes.WorkloadType = "network-light"
+	// ZDBLightType type
+	ZDBLightType gridtypes.WorkloadType = "zdb-light"
```
muhamadazmy: Why do we need zdb-light? is it because it doesn't have ygg IP? 

I think we should only use the "-light" suffix if the data type for input or output changes (like vm and network) but I don't believe this applies to zdb

pkg/netlight/resource/resource.go
```diff
@@ -0,0 +1,381 @@
...
+
+	if namespace.Exists(nsName) {
+		return &Resource{name}, nil
+	}
+
+	return nil, fmt.Errorf("resource not found: %s", name)
+}
+
+// myceliumIpSeed is 6 bytes
+func (r *Resource) AttachPrivate(id string, vmIp net.IP) (device pkg.TapDevice, err error) {
```
muhamadazmy, line 217: I can't figure out if this an idempotent function or not. But it's usually important that such functions to be idempotent so recreation doesn't happen if the service was restarted for example

pkg/network_light.go
```diff
@@ -0,0 +1,43 @@
...
+	Namespace(id string) string
+	Ready() error
+	ZOSAddresses(ctx context.Context) <-chan NetlinkAddresses
+}
+
+type TapDevice struct {
+	Name    string
+	Mac     net.HardwareAddr
+	IP      *net.IPNet
+	Gateway *net.IPNet
```
muhamadazmy: I think this should be `Routes` instead so we can also have custom routes not only default GW. It's needed with mycelium for example and in other cases where custom routing is required

muhamadazmy (commented):
I didn't cover all the PR because of its size. But i think I have left enough comments for the first iteration.

I think using the `Light` prefix might not be the best approach. Maybe use a version instead? so it's future proof

cmds/modules/provisiond/reporter.go
ashraffouda: we decided to postpone gw and qsfs for later versions of zos light

bootstrap/bootstrap/src/bootstrap.rs
```diff
@@ -26,15 +26,19 @@ fn bootstrap_zos(cfg: &config::Config) -> Result<()> {
     let flist = match &cfg.runmode {
         RunMode::Prod => match &cfg.version {
             Version::V3 => "zos:production-3:latest.flist",
+            _ => bail!("unsupported version in old style"),
```
muhamadazmy, line 29: it's safe to completely remove the old style from code since ALL environments have been updated to new style. We needed to keep the old style for a while because bootstrap is shared between all environments (since it's part of base image) because some envs used new style and others didn't.

Anyway, you can assume that old style does no longer exist

cmds/modules/netlightd/main.go
```diff
@@ -0,0 +1,170 @@
...
+
+	rules, err := nftRules.Open("nft/rules.nft")
+	if err != nil {
+		return fmt.Errorf("failed to load rules.nft file")
+	}
+
+	if err := nft.Apply(rules, ""); err != nil {
+		return fmt.Errorf("failed to apply host nft rules: %w", err)
+	}
+
```
muhamadazmy, line 109: close the `rules` file here

cmds/modules/netlightd/nft/rules.nft
```diff
@@ -0,0 +1,29 @@
+add table inet filter;
```
muhamadazmy, line 1: What I meant with this is to make it similar to other `rules.nft` (in the same PR here style. It doesn't have commands but instead the complete structure of the rule set

It contains simply the output of running `nft list ruleset` 

If you don't get what  I mean let me explain over DM

pkg/netlight/resource/resource.go
```diff
@@ -0,0 +1,399 @@
...
+
+	rules, err := nftRules.Open("nft/rules.nft")
+	if err != nil {
+		return nil, fmt.Errorf("failed to load nft.rules file")
+	}
+
+	if err := nft.Apply(rules, nsName); err != nil {
+		return nil, fmt.Errorf("failed to apply nft rules for namespace '%s': %w", name, err)
+	}
+
```
muhamadazmy, line 146: close rules file here

pkg/network_light.go
```diff
@@ -0,0 +1,43 @@
...
+	Namespace(id string) string
+	Ready() error
+	ZOSAddresses(ctx context.Context) <-chan NetlinkAddresses
+}
+
+type TapDevice struct {
+	Name   string
+	Mac    net.HardwareAddr
+	IP     *net.IPNet
+	Routes []Route
```
muhamadazmy, line 30: Where is Route defined

pkg/upgrade/upgrade.go
```diff
@@ -200,17 +200,21 @@ func (u *Upgrader) nextUpdate() time.Duration {
 func (u *Upgrader) remote() (remote hub.TagLink, err error) {
 	mode := u.boot.RunMode()
 	// find all taglinks that matches the same run mode (ex: development)
+	envName := mode.String()
+	if u.boot.Version().Major == 4 {
+		envName = fmt.Sprintf("%s-light", envName)
+	}
```
muhamadazmy, line 206: Since this is a fork and the entire fork knows that this is version 4. Hence you can always assume the envName to be "<env>-light" without even the need to check for the major version

pkg/vm/ch.go
```diff
@@ -24,43 +23,6 @@ const (
 	cloudConsoleBin = "cloud-console"
 )
 
-// startCloudConsole Starts the cloud console for the vm on it's private network ip
-func (m *Machine) startCloudConsole(ctx context.Context, namespace string, networkAddr net.IPNet, machineIP net.IPNet, ptyPath string, logs string) (string, error) {
```
muhamadazmy, line 28: I think the cloud console can stay since the user network will always have a mycelium IP. the console can still work over the mycelium IP in case the VM is not reachable for any reason. 
- So the ip of the vm is the VM private ip (private range)
- The console listen port is calculated the same way
- The console IP itself is the mycelium IP assigned to this user mycelium instance.

pkg/vm/ch.go
```diff
@@ -131,12 +93,16 @@ func (m *Machine) Run(ctx context.Context, socket, logs string) (pkg.MachineInfo
 
 		for _, nic := range m.Interfaces {
 			var typ InterfaceType
-			typ, _, err = nic.getType()
+			typ, idx, err := nic.getType()
```
muhamadazmy, line 96: When you do Attach on the resource, you should create a tap device (not macvtap) which makes it easier to attach it to CH command line. The macvtap case was only used for public IPs specifically.

pkg/netlight/resource/resource.go
```diff
@@ -0,0 +1,399 @@
...
+	privateNetBr := fmt.Sprintf("r%s", r.name)
+	hw := ifaceutil.HardwareAddrFromInputBytes([]byte(tapName))
+
+	ip := &net.IPNet{
+		IP:   vmIp,
+		Mask: gw.Mask,
+	}
+	_, getLinkErr := netlink.LinkByName(tapName)
+	if getLinkErr != nil {
+		mtap, err := macvtap.CreateMACvTap(tapName, privateNetBr, hw)
```
muhamadazmy, line 262: We create TAP devices for VMs not Macvtap. this makes it un0necessary complex to pass this device to the VM when it can just be a tap device

pkg/netlight/resource/resource.go
```diff
@@ -0,0 +1,399 @@
...
+		return
+	}
+
+	deviceName := ifaceutil.DeviceNameFromInputBytes([]byte(id))
+	tapName := fmt.Sprintf("m-%s", deviceName)
+
+	myBr := fmt.Sprintf("m%s", r.name)
+	hw := ifaceutil.HardwareAddrFromInputBytes([]byte(tapName))
+
+	_, err = macvtap.CreateMACvTap(tapName, myBr, hw)
```
muhamadazmy, line 319: Same for mycelium this can perfectly be a tap device (not macvtap) hence it becomes simpler to pass to the `cloud-hypervisor` command line

muhamadazmy (changes requested):
Thanks @ashraffouda for the great work. I still have few comments in case you are interested

pkg/netlight/resource/resource.go
delandtj: macvtaps or macvlans are a disaster if they're more than one with a bridge interface as parent.
Somewhere since that dummy necessity version, macvlans have problems ... what they specifically are, I'm not sure yet, but I suspect seen that macvlans are also bridges in a lighter version, they don't cooperate well with regular bridges

delandtj: I felt like since the https://bugzilla.kernel.org/show_bug.cgi?id=214631 bug, macvlans went worse and worse over time.
probably because of their infrequent use.
let's see how much work it is, but I guess

pkg/primitives/vm-light/vm.go
```diff
@@ -199,42 +187,14 @@ func (p *Manager) virtualMachineProvisionImpl(ctx context.Context, wl *gridtypes
 		if err != nil {
```
iwanbk, line 187: according to https://github.com/threefoldtech/zos/pull/2397/files#diff-21cf55cfe1ba697c2e7be4ae50cbe646bca4331061bba94ebc35fb5f7d05ca98R162-R165

we only have/use one private network, why still looping all net interfaces instead of using the only used one?

pkg/primitives/vm-light/vm.go
```diff
@@ -199,42 +187,14 @@ func (p *Manager) virtualMachineProvisionImpl(ctx context.Context, wl *gridtypes
...
-			return result, err
-		}
-		ifs = append(ifs, wl.ID.Unique("ygg"))
-
-		log.Debug().Msgf("Planetary: %+v", inf)
 		networkInfo.Ifaces = append(networkInfo.Ifaces, inf)
-		result.PlanetaryIP = inf.IPs[0].IP.String()
 	}
 
 	if config.Network.Mycelium != nil {
```
iwanbk, line 193: is it allowed to not have mycelium?

iwanbk: it is not allowed according to the specs, CMIIW

pkg/netlight/network.go
```diff
@@ -0,0 +1,378 @@
...
+	}
+)
+
+type networker struct {
+	ipamLease string
+}
+
+var _ pkg.NetworkerLight = (*networker)(nil)
+
+func NewNetworker() (pkg.NetworkerLight, error) {
```
iwanbk, line 50: is it common to add `er` to the `struct` in `zos`?
because `er` suffix usually used for `interface`, not `struct`

pkg/netlight/resource/resource.go
```diff
@@ -0,0 +1,399 @@
...
+			return err
+		}
+		return nil
+	}); err != nil {
+		return
+	}
+	if len(addrs) != 1 {
+		return device, fmt.Errorf("expect addresses on private interface to be 1 got %d", len(addrs))
+	}
+	gw := addrs[0].IPNet
```
iwanbk, line 245: for readability, i think we can move this block into separate func, maybe something like:
`getAndValidateAddr()`

And maybe including the code to get the namespace
```go
nsName := fmt.Sprintf("n%s", r.name)
	netNs, err := namespace.GetByName(nsName)
```

iwanbk: i think the code below it can also be included
```go
if !gw.Contains(vmIp) {
		return device, fmt.Errorf("ip not in range")
	}
```

pkg/netlight/resource/resource.go
```diff
@@ -0,0 +1,399 @@
...
+
+	privateNetBr := fmt.Sprintf("r%s", r.name)
+	hw := ifaceutil.HardwareAddrFromInputBytes([]byte(tapName))
+
+	ip := &net.IPNet{
+		IP:   vmIp,
+		Mask: gw.Mask,
+	}
+	_, getLinkErr := netlink.LinkByName(tapName)
+	if getLinkErr != nil {
```
iwanbk, line 261: so, is it OK if the `tap` already exists?

pkg/netlight/resource/mycelium.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+)
+
+type MyceliumInspection struct {
+	PublicKey string `json:"publicKey"`
+	Address   net.IP `json:"address"`
+}
+
+func inspectMycelium(seed []byte) (inspection MyceliumInspection, err error) {
+	// we check if the file exists before we do inspect because mycelium will create a random seed
```
iwanbk, line 44: > we check if the file exists

can't find the check

pkg/netlight/resource/mycelium.go
```diff
@@ -0,0 +1,252 @@
...
+		return inspection, err
+	}
+
+	defer os.RemoveAll(tmp.Name())
+
+	if _, err := tmp.Write(seed); err != nil {
+		return inspection, fmt.Errorf("failed to write seed: %w", err)
+	}
+
+	tmp.Close()
```
iwanbk, line 57: move to the above of the `if` block?

xmonader:
ايش منتظرين طال عمرك

ashraffouda:
this pr is merged in zos4 long time ago will close it

PR #2408, opened 2024-08-26
don't check the passed version if we are in 1st stage

ashraffouda:
### Description
don't consider version in first stage of bootstrap

### Changes

skip passed version if bootstrap is running in 1st stage

### Related Issues

List of related issues

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/bootstrap/bootstrap/src/config.rs
+++ b/bootstrap/bootstrap/src/config.rs
@@ -119,12 +119,15 @@ impl Config {
         if stage == 0 {
             bail!("invalid stage value 0, stages starting from 1");
         }
-
+        let mut ver = Version::V3;
+        if stage == 2 {
+            ver = version()?
+        }
         Ok(Config {
             stage,
             debug: matches.occurrences_of("debug") > 0,
             runmode: runmode()?,
-            version: version()?,
+            version: ver,
         })
     }
 }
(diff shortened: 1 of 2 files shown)


rawdaGastan (approved):

iwanbk:
@ashraffouda maybe add the context/reason of this change?

ashraffouda:
> @ashraffouda maybe add the context/reason of this change?

bootstrap running in two stages, during 1st stage bootstrap only upgrades itself so if we added a new version like v4 in bootstrap and the running image has old version of bootstrap it will fail because bootstrap doesn't know the new version at the 1st stage, so it first needs to run and update itself to get the new changes which contains the newly added version

bootstrap/bootstrap/src/config.rs
```diff
@@ -119,12 +119,15 @@ impl Config {
         if stage == 0 {
             bail!("invalid stage value 0, stages starting from 1");
         }
-
+        let mut ver = Version::V3;
```
iwanbk, line 122: So, looks like it will become the version latest-1?

what happens when we upgrade  `zos`  with `latest - 2` or `latest` version?
In this context, it means v2 or v4

ashraffouda: at 1st step we set it to v3 this is the current supported version so it will not fail at 1st step

Issue #2403, opened 2024-08-21
Completely forego the usage of macvlans, as they wreak havoc in ARP/IPv6 ND when more than one macvlan is created form a bridge interface


delandtj:
Creating a macvlan from a bridge interface seems to conflict in arp/nd for the bridge, as as well the bridge(duh) as macvlans have bridging code and somewhere along the path things don't fit any more.
Creating macvlans from physical interfaces doesn't seem to create problems, but from bridge interfaces (when there is more than one and both interfaces need to communicate) it does weird things, where neighbors get lost, asn as such routes get lost, ans a such, connections get lost.

some comments needed here : 
  - if a macvlan or macvtap is created on a bridge for a workload that is still alive, we will have to see if we want to leave it like that or recreate it.  Or just tell the user to recreate his workload.
  - zdbs are using the old container code, so I guess we can do that 'quite' easily?
  - all other workloads are taps ?

scottyeager:
There might be some info about the sources of conflict [here](https://cnly.github.io/2018/11/09/conflicts-and-limitations-of-bridge-and-macvlan-devices.html) (just a note for reference).

So far I understand this affects the Mycelium interface for Zdbs. Are VMs also affected? If I read correctly here it seems so:

https://github.com/threefoldtech/zos/blob/9cd81f3ec8049a224b3c819e3b329a28835ae92c/docs/internals/network/topology/readme.md?plain=1#L84

> * we will have to see if we want to leave it like that or recreate it.  Or just tell the user to recreate his workload.

I think it's fine to leave existing VMs as is and let the users recreate if they want. If the user is experience a problem due to this, they probably have found another communication route and are happy enough with their VM as is.

For Zdb, ideally the endpoints could retain all the same IP addresses and nothing changes for the users. If that's not possible, oh well. There's not much usage yet anyway.

LeeSmet:
VMs always use regular TAP interfaces (NOT macvtaps), and therefore these should be fine. The problem exists primarily for zdbs because (for mycelium), a packet traverses 2 macvlans in bridge mode which are themselves enslaved by a bridge. It appears to be specifically the second macvlan which causes issues when acting as router, since "regular" public ipv6 has the same setup (macvlan from br-pub to zdb ns and macvlan from br-pub to public ns), however here the packet "exits" on the bridge since the bridge enslaved a physical interface which is the path for the router (so for regular IPv6 the public namespace is not involved).

ashraffouda:
[...]

PR #2414, opened 2024-09-03
replace macvlan with veth pairs

Omarabdul3ziz:
### Description
replace macvlan with veth pairs

### Related Issues
- https://github.com/threefoldtech/zos/issues/2403

List of related issues

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/pkg/network/ifaceutil/interface.go
+++ b/pkg/network/ifaceutil/interface.go
@@ -134,13 +134,20 @@ func RandomName(prefix string) (string, error) {
 }
 
 // MakeVethPair creates a veth pair
-func MakeVethPair(name, master string, mtu int) (netlink.Link, error) {
+func MakeVethPair(name, master string, mtu int, peerPrefix string) (netlink.Link, error) {
 	masterLink, err := netlink.LinkByName(master)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("master link: %s not found: %v", master, err)
 	}
-
-	peer := fmt.Sprintf("%s-p", name)
+	peer := ""
+	if peerPrefix == "" {
+		peer = fmt.Sprintf("p-%s", name)
+	} else {
+		peer = fmt.Sprintf("%s-%s", peerPrefix, name)
+	}
+	if len(peer) > 15 {
+		peer = peer[0:15]
+	}
 	veth := &netlink.Veth{
 		LinkAttrs: netlink.LinkAttrs{
 			Name:  name,
--- a/pkg/network/ifaceutil/interface.go
+++ b/pkg/network/ifaceutil/interface.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/containernetworking/plugins/pkg/ip"
 	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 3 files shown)


pkg/network/networker.go
```diff
@@ -156,28 +157,54 @@ func (n *networker) attachYgg(id string, netNs ns.NetNS) (net.IPNet, error) {
...
+			}
+
+			if err := options.SetIPv6Forwarding(true); err != nil {
+				return fmt.Errorf("failed to enable ipv6 forwarding in namespace %q: %w", name, err)
+			}
+
+			return netlink.RouteAdd(&route)
+		}); err != nil {
+			return net.IPNet{}, fmt.Errorf("failed to setup link addresses %s: %w", ZDBYggIface, err)
+		}
```
iwanbk: Looks like this code block does the same thing as https://github.com/threefoldtech/zos/pull/2414/files#diff-4e15b6f2760eb52c8292eadf92ce4a6ba3afcf08d34b528d56081b9bc1623554R252-R285

extract it to a func?

pkg/network/ifaceutil/interface.go
```diff
@@ -179,6 +217,40 @@ func MakeVethPair(name, master string, mtu int) (netlink.Link, error) {
 	return veth2, nil
 }
 
+// moveLinkToNamespace moves the veth link to the specified namespace and renames it
+func moveLinkToNamespace(veth netlink.Link, tmpName, finalName string, netNs ns.NetNS, peerLink netlink.Link) (netlink.Link, error) {
```
ashraffouda: i suggest we make tmpName > name the underlying method shouldn't know about tmpName I think
and finalName >  renameTo which would be a pointer because it maybe nil if we need to just move it to the name space

pkg/network/networker.go
```diff
@@ -210,62 +203,120 @@ func (n *networker) attachMycelium(id string, netNs ns.NetNS) (net.IPNet, error)
...
 }
 
-// ensurePrepare ensurets that a unique namespace is created (based on id) with "prefix"
+func setLinkAddr(name string, ip *net.IPNet) error {
+	link, err := netlink.LinkByName(name)
+	if err != nil {
+		return fmt.Errorf("failed to set link address: %w", err)
+	}
+
+	// if err := options.Set(name, options.IPv6Disable(false)); err != nil {
```
ashraffouda: remove unneeded commented coder

pkg/network/networker.go
```diff
@@ -210,62 +203,120 @@ func (n *networker) attachMycelium(id string, netNs ns.NetNS) (net.IPNet, error)
...
+	}
+	err = netlink.AddrAdd(link, &addr)
+	if err != nil && !os.IsExist(err) {
+		return fmt.Errorf("failed to add ip address to link: %w", err)
+	}
+
+	return netlink.LinkSetUp(link)
+}
+
+func createInterface(ifName, bridge string, netNs ns.NetNS, ip *net.IPNet, route *netlink.Route) error {
```
ashraffouda: rename it to CreateVethInterface

ashraffouda (changes requested):

pkg/network/ifaceutil/interface.go
```diff
@@ -133,25 +134,60 @@ func RandomName(prefix string) (string, error) {
...
+// - netNs: network namespace to move the veth interface to. (could be nil)
+//
+// Returns the created netlink.Link or an error
+func MakeVethPair(name, master string, mtu int, netNs ns.NetNS) (netlink.Link, error) {
 	masterLink, err := netlink.LinkByName(master)
+	if err != nil {
+		return nil, fmt.Errorf("master link: %s not found: %v", master, err)
+	}
+
+	nsName := filepath.Base(netNs.Path())
```
muhamadazmy: What if netNs is `nil` I see you already calling the function with `nil` as a last argument already in some places

pkg/network/ifaceutil/interface.go
```diff
@@ -133,25 +134,60 @@ func RandomName(prefix string) (string, error) {
...
+	// collide with the name on the host and error out
+	// tmpName will be renamed after moving it to the namespace
+	tmpName, err := ip.RandomVethName()
 	if err != nil {
 		return nil, err
 	}
+	ifName := name
+	if netNs != nil {
+		ifName = tmpName
+	}
```
muhamadazmy: Would be easier to read if it's re-written like
```go
ifName := name
if netNs != nil {
   ifName, err = ip.RandomVethName()
   if err != nil {
       return nil, err
   }
}
```

pkg/network/ifaceutil/interface.go
```diff
@@ -161,24 +197,67 @@ func MakeVethPair(name, master string, mtu int) (netlink.Link, error) {
...
+	if err := netlink.LinkSetMaster(peerLink, masterLink); err != nil {
+		return nil, fmt.Errorf("failed to set master for peer link %q: %w", peer, err)
 	}
 
 	// Re-fetch the link to get its creation-time parameters, e.g. index and mac
-	veth2, err := netlink.LinkByName(name)
+	// either from the host, or from the namespace
+	var veth2 netlink.Link
+
+	veth2, err = netlink.LinkByName(name)
```
muhamadazmy: Does a link with name `name` even exist here? Note that so far the 2 links that exists are `ifName` and `peer`, not `name`!

Also `veth2` is not a good varible name. i don't know what you refer too.

pkg/network/ifaceutil/interface.go
```diff
@@ -161,24 +197,67 @@ func MakeVethPair(name, master string, mtu int) (netlink.Link, error) {
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// make sure the lowerhalf is up, this automatically sets the upperhalf UP
+	if err := netlink.LinkSetUp(peerLink); err != nil {
+		return nil, fmt.Errorf("failed to set veth peer up: %w", err)
+	}
+
```
muhamadazmy, line 208: I don't believe this function should return anything because the `veth` can be inside a namespace. So returning a link that is inside a namespace does not make any sense. Normally to work with links that lives inside a namespace you have to do 
```go
ns.Run(func() {
  // load the nic here
  // work with the nic
}
```

pkg/network/networker.go
```diff
@@ -156,28 +157,21 @@ func (n *networker) attachYgg(id string, netNs ns.NetNS) (net.IPNet, error) {
...
 
-	routes := []*netlink.Route{
-		{
-			Dst: &net.IPNet{
-				IP:   net.ParseIP("200::"),
-				Mask: net.CIDRMask(7, 128),
-			},
-			Gw: gw.IP,
-			// LinkIndex:... this is set by macvlan.Install
+	route := netlink.Route{
```
muhamadazmy: I think even if this `now` has only a single route. It's important to understand that this won't always be the case. It's totally legal to configure multiple routes and taking away this ability is not recommended.

muhamadazmy (commented):
thanks @Omarabdul3ziz for the good work :) I left few comments below that might be helpful

ashraffouda (dismissed):

iwanbk (dismissed):

muhamadazmy (approved):

Issue #2413, opened 2024-09-02
Controlled rollout of ZOS on mainnet


xmonader:
We need to implement a Controlled rollout for ZOS upgrades, specially on mainnet to facilitate controlled experiments on different nodes or farms. The primary goal is to allow for testing the next version of ZOS on selected nodes or farms without impacting the entire network. This will be crucial for evaluating new features or optimizations before rolling them out to the broader network.

- We should be able to define the mainnet farms list to be used in the A/B testing
- We should be able to define the mainnet nodes list to be used in the A/B testing
- `safe_to_upgrade_network` defaulted to `false`:  This flag will be used to indicate whether it is safe to proceed with network-wide upgrades with the latest zos version specified on chain or not

TODO...

iwanbk:
Doesn't mean to be picky on the wording part.

> We need to implement A/B Testing functionality for ZOS upgrades to facilitate controlled experiments on different nodes or farms. The primary goal is to allow for testing the next version of ZOS on selected nodes or farms without impacting the entire network

But i think what you really mean is `staging`, not A/B testing @xmonader ?
The main difference is for `staging`, we really want that the feature/version to be finally deployed.
While in A/B testing, we want to choose one between two or more alternatives.

> This will be crucial for evaluating new features or optimizations before rolling them out to the broader network.

Other than staging, we can also employ feature flag/toggle technique:
- activate the feature for several controlled users/nodes
- deactivate when things go wrong.

xmonader:
We have already qanet, testnet as staging environments for the release in  pipeline, what is needed is controlled rollout on a small, defined, subset of nodes on mainnet.

iwanbk:
Oh okay, feature flag/toggle then

xmonader:
Flag is already toggled/set as part of zos upgrade. When we want to upgrade nodes on mainnet network, we create a proposal - on tfchain - that has a zos version to upgrade the whole network to, and as soon as the node picks up that proposal approval it starts its upgrade. What is needed is breaking that into two steps:

- changing the version of zos on chain
- controlled rollout, by allowing some of the nodes to reflect as soon as they become aware of a new version and manual approval to propagate across the network - after testing on that's first batch of nodes-.

iwanbk:
> controlled rollout

OK, so it is clear that we want is controlled rollout.

rawdaGastan:
Farm IDs can be included in the A/B testing but we can't include node IDs.
We can't get the ID of the node before the registration.

[...]

PR #2417, opened 2024-09-11
add rollout configs for zos update

rawdaGastan:
### Related Issues

- https://github.com/threefoldtech/zos/issues/2413


pkg/upgrade/rollout.go
```diff
@@ -0,0 +1,60 @@
...
+	rolloutConfig, err := parseRolloutConfig(response.Body)
+	if err != nil {
+		return RolloutConfig{}, err
+	}
+
+	if _, ok := rolloutConfig[env]; !ok {
+		return RolloutConfig{}, fmt.Errorf("env '%s' does not exist in the configs", env)
+	}
+
+	return rolloutConfig[env], nil
```
iwanbk: we can also do this
```go
val, ok := rolloutConfig[env]
if !ok { 
    return RolloutConfig{}, fmt.Errorf("env '%s' does not exist in the configs", env)
}
return val, nil 
```
cleaner IMO

pkg/upgrade/rollout.go
```diff
@@ -0,0 +1,60 @@
...
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/threefoldtech/zos/pkg/environment"
+)
+
+const defaultRolloutConfigURL = "https://raw.githubusercontent.com/threefoldtech/zos/development_rollout_update/pkg/upgrade/config.json"
+
+// RolloutConfig is the configuration for A/B testing before zos update
```
iwanbk: >  A/B testing

`whitelist node/farm?` looks more suitable

pkg/upgrade/rollout.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/threefoldtech/zos/pkg/environment"
+)
+
+const defaultRolloutConfigURL = "https://raw.githubusercontent.com/threefoldtech/zos/development_rollout_update/pkg/upgrade/config.json"
```
rawdaGastan: update the url

pkg/upgrade/rollout.go
xmonader: please don't use `white/black` lists in code. only use allowed/disallowed lists. (test_farms is also good)

pkg/upgrade/rollout.go
```diff
@@ -0,0 +1,60 @@
...
+
+// RolloutConfig is the configuration for A/B testing before zos update
+type RolloutConfig struct {
+	TestFarms     []uint32 `json:"test_farms"`
+	SafeToUpgrade bool     `json:"safe_to_upgrade"`
```
xmonader: I know this is reflecting the current specs, but should consider specifying the version instead of having just a bool allowing the "latest" whatever that latest is?

pkg/upgrade/rollout.go
xmonader: are we going to maintain it here instead of zos-config repo?https://github.com/threefoldtech/zos-config

pkg/upgrade/rollout.go
```diff
@@ -0,0 +1,60 @@
...
+		if network != environment.RunningDev.String() && network != environment.RunningQA.String() && network != environment.RunningTest.String() && network != environment.RunningMain.String() {
+			return nil, fmt.Errorf("invalid network passed: %s", network)
+		}
+	}
+
+	return conf, nil
+}
+
+func readRolloutConfig(env string) (RolloutConfig, error) {
+	response, err := http.Get(defaultRolloutConfigURL)
```
xmonader: Shouldn't we check against status code ?

pkg/upgrade/rollout.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to read the rollout config file: %+w", err)
+	}
+
+	if err = json.Unmarshal(configFile, &conf); err != nil {
+		return nil, err
+	}
+
+	for network := range conf {
+		if network != environment.RunningDev.String() && network != environment.RunningQA.String() && network != environment.RunningTest.String() && network != environment.RunningMain.String() {
```
xmonader: I believe it could get simplified e.g 
```
var allowedNetworks = map[string]struct{}{
    environment.RunningDev.String():  {},
    environment.RunningQA.String():   {},
    environment.RunningTest.String(): {},
    environment.RunningMain.String(): {},
}

for network := range conf {
    if _, exists := allowedNetworks[network]; !exists {
        return nil, fmt.Errorf("invalid network passed: %s", network)
    }
}
```

pkg/upgrade/rollout.go
```diff
@@ -0,0 +1,60 @@
...
+	TestFarms     []uint32 `json:"test_farms"`
+	SafeToUpgrade bool     `json:"safe_to_upgrade"`
+}
+
+func parseRolloutConfig(file io.Reader) (map[string]RolloutConfig, error) {
+	var conf map[string]RolloutConfig
+
+	configFile, err := io.ReadAll(file)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read the rollout config file: %+w", err)
```
xmonader: Why do we need to add plus flag to %w verb?

xmonader (approved):

Issue #2421, opened 2024-09-17
add features endpoint for zos3 and zos4


ashraffouda:
we need an endpoint that gets the supported features for zos node hence we have two different version now 3 and 4 and each one has its own workload types so we need this for dashboard for example to determine which types of workloads are valid for the node we are deploying on
something like the types here https://github.com/threefoldtech/zos/blob/main/pkg/gridtypes/zos/types.go#L10
the endpoint should return the supported types for workload names

Omarabdul3ziz:
Verified 

call a node with `zos.system.node_features_get` rmb endpoint

zos3 nodes 
![image](https://github.com/user-attachments/assets/b88258bb-d723-4e7a-9f40-1e50da17d088)

zos4 nodes 
![image](https://github.com/user-attachments/assets/eb5668a8-a1ef-49e0-9b25-d503461b2fcd)

respond with the expected

PR #2422, opened 2024-09-17
add new endpoint to fetch node features

Eslam-Nawara:
### Description

add new endpoint to fetch node features

### Changes

add new endpoint to fetch node features

### Related Issues

- #2421 

### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation

### DIFF
--- a/client/node.go
+++ b/client/node.go
@@ -320,8 +320,8 @@ func (n *NodeClient) NetworkGetPublicConfig(ctx context.Context) (cfg pkg.Public
 	return
 }
 
-func (n *NodeClient) NetworkGetNodeFeatures(ctx context.Context) (feat pkg.NodeFeatures, err error) {
-	const cmd = "zos.network.node_features_get"
+func (n *NodeClient) SystemGetNodeFeatures(ctx context.Context) (feat []pkg.NodeFeature, err error) {
+	const cmd = "zos.system.node_features_get"
 
 	err = n.bus.Call(ctx, n.nodeTwin, cmd, nil, &feat)
 	return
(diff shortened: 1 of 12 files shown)


client/node.go
```diff
@@ -322,6 +320,13 @@ func (n *NodeClient) NetworkGetPublicConfig(ctx context.Context) (cfg pkg.Public
 	return
 }
 
+func (n *NodeClient) NetworkGetNodeFeatures(ctx context.Context) (feat pkg.NodeFeatures, err error) {
+	const cmd = "zos.network.node_features_get"
```
xmonader: Why on the zos.network? I assumed it will be on `zos.system.node_features`

Eslam-Nawara: since this is related to netowrkd-light I assumed it would be on network, but can move it to be system.node_features

xmonader: My reasoning is as a user I should ask the node what version do you have? (which is `zos.system.version`) or what features are enabled on the node so `zos.system.node_features_get`

pkg/network.go
```diff
@@ -257,6 +260,32 @@ type PublicConfig struct {
 	Domain string `json:"domain"`
 }
 
+type NodeFeatures struct {
```
xmonader: NodeFeatures should have been a slice of a Feature []Feature and that feature is a typed string that can be any of `zdb, zmachine, .. etc)

pkg/monitor.go
```diff
@@ -78,13 +78,17 @@ type PoolStats struct {
...
+type NodeFeature string
+
 // SystemMonitor interface (provided by noded)
 type SystemMonitor interface {
 	NodeID() uint32
 	Memory(ctx context.Context) <-chan VirtualMemoryStat
 	CPU(ctx context.Context) <-chan TimesStat
 	Disks(ctx context.Context) <-chan DisksIOCountersStat
 	Nics(ctx context.Context) <-chan NicsIOCounterStat
+	GetNodeFeatures() ([]NodeFeature, error)
```
xmonader: About the suggested interface, will this ever fail? worst case would be an empty slice :)

xmonader (approved):

muhamadazmy (approved):

Issue #2401, opened 2024-08-20
Periodically check/update the node location


0oM4R:
## Is your feature request related to a problem? 
changes on MaxMind Database lead to inconstant location data,  https://github.com/threefoldtech/tfgrid-sdk-ts/issues/3021#issuecomment-2282836629

 
## Describe the solution you'd like

On an event on the node, we can call the geoip service and check if it doesn't match the stored value, then the location should be updated.

**Requires:** https://github.com/threefoldtech/tfchain/issues/995

Omarabdul3ziz:
blocked: https://github.com/threefoldtech/tfchain/issues/995#issuecomment-2358567305

ashraffouda:
works on zos3 needs to be added to zos4 
need to finish this on zos4 to be closed https://github.com/threefoldtech/zos4/pull/11

ashraffouda:
on zos4 now

PR #2427, opened 2024-09-22
init node location updater

Omarabdul3ziz:
- add updater package
- expose GetState from registerer
- fix longitude spelling
- https://github.com/threefoldtech/zos/issues/2401

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


muhamadazmy (changes requested):
While i am not sure why zos needs to update location every 24 hours i don't believe this is the right way.

The registration service should be the only service to issue an update node (or create node) start to update the node outside this context can lead to unmanaged and hard to debug code.

Registration service already has a mechanism to update the node when some changes are detected (like ip change) and i think this should be included there

Omarabdul3ziz:
the issue is updates happen to some ips on the MaxMind database it may be a change in the country name from English chars to native chars like `Turkey` > `Türkiye` or the country name changed.

the suggestion was to update the node location periodically. since the changes in the database aren't frequent i thought once a day is a good period to check.

i moved the code for update location to be part of the register package

pkg/registrar/registrar.go
```diff
@@ -146,6 +152,10 @@ func (r *Registrar) register(ctx context.Context, cl zbus.Client, env environmen
 			if err := r.reActivate(ctx, cl, env); err != nil {
 				log.Error().Err(err).Msg("failed to reactivate account")
 			}
+		case <-time.After(updateLocationInterval):
+			if err := r.updateLocation(ctx, cl); err != nil {
```
muhamadazmy: Code in general looks way better. But why just not call `register()` here. It already collects the location, then compare if the node information has changed before calling update.

muhamadazmy (changes requested):

muhamadazmy (approved):

PR #2431, opened 2024-10-02
use chain to get version and safe to upgrade flag

rawdaGastan:
### Related Issues

- https://github.com/threefoldtech/zos/issues/2413

### DIFF
--- a/pkg/upgrade/upgrade.go
+++ b/pkg/upgrade/upgrade.go
@@ -2,6 +2,7 @@ package upgrade
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"math/rand"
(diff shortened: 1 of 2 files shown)


pkg/upgrade/upgrade.go
```diff
@@ -49,6 +51,38 @@ const (
...
+	if err != nil {
+		return ChainVersion{}, nil, err
+	}
+
+	v, err := con.GetZosVersion()
+	if err != nil {
+		return ChainVersion{}, nil, errors.Wrap(err, "failed to get zos version from chain")
+	}
+
+	var chainVersion ChainVersion
```
ashraffouda, line 77: we need to handle old version style for a while just in case it is not updated on some envs

ashraffouda (approved):

Issue #2441, opened 2024-10-22
reduce mycelium logs


Omarabdul3ziz:
mycelium makes a lot of INFO logs which makes it harder to trace its errors or other modules errors. could we change the log level to a higher version if it is acceptable 
![image](https://github.com/user-attachments/assets/a9da1146-0906-43d7-b33e-b757047512cb)

Eslam-Nawara:
added silent flag when running mycelium and now only errors are logged from mycelium
```console
zinit log mycelium
```
![image](https://github.com/user-attachments/assets/32ead1fb-b8bf-4b53-85cd-367ad4ca92c8)

```console
zinit log
```
![image](https://github.com/user-attachments/assets/51c6062a-6dbf-4861-a51e-bcc1a5166058)

Omarabdul3ziz:
Verified 

now it shows only the error msgs
![image](https://github.com/user-attachments/assets/c844ebea-869b-4a45-995f-cea4fb6cfafc)

PR #2442, opened 2024-10-22
reduce mycelium logs by adding silent flag

Eslam-Nawara:
### Description

reduce mycelium logs by adding silent flag

### Changes

add silent flag to mycelium

### Related Issues

- #2441 

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/pkg/network/mycelium/mycelium.go
+++ b/pkg/network/mycelium/mycelium.go
@@ -88,6 +88,7 @@ func (s *MyceliumServer) Ensure(z *zinit.Client, ns string) error {
 		bin,
 		"--key-file", confPath,
 		"--tun-name", s.cfg.TunName,
+		"--silent",
 		"--peers",
 	}
 


Omarabdul3ziz (approved):

iwanbk:
how `silent` is this flag @LeeSmet ?
I'm afraid that it would make us difficult to debug, if some issue happens

Because the noise mostly came from the routing, maybe we could set `routing` log level to error?

Eslam-Nawara:
@iwanbk the `silent` flags logs only error level. so any error would appear in the logs.

```console
--silent
    Disable all logs except error logs
```

 this flag is already used with mycelium instance used in all network resources.
check https://github.com/threefoldtech/zos/blob/ef4e2807597ba6b530fb455d647ea9bce5101071/pkg/network/nr/net_resource.go#L308

LeeSmet:
It basically supresses everything which is not an error.

Changing logs to error just to have them logged is not a good idea imo, since "error" log level also conveys that something went wrong

iwanbk (approved):

Issue #2444, opened 2024-10-24
Only allow verified user to deploy


AhmedHanafy725:
require: https://github.com/threefoldtech/tfgrid-kyc-verifier/issues/5

The account's status can be queried from the KYC service using the account's address.

Eslam-Nawara:
## WIP 
After the kyc server is updated I tested the status endpoint and now it's working as expected

I was able to boot a zos node on mainnet that is able to restrict deployments to users with verified accounts and I was able to only deploy on this node with a verified account, deployment failed if the account is not verified

**Note:** this only works with mainnet as there's only one kyc server instance is up only for mainnet, for all other stacks the check is skipped for now

- Deployment with verified account
![image](https://github.com/user-attachments/assets/6254a971-c031-432c-b29b-fd9364533e19)

- Deployment with unverified account 
![image](https://github.com/user-attachments/assets/2b583eab-0e34-4db1-bb69-fad46d7ea4a3)

rawdaGastan:
## Verification for zos3 and zos4

#### Failure scenarios

- Tried to deploy with unverified account and got `unverified` error

![Screenshot from 2024-10-30 13-24-09](https://github.com/user-attachments/assets/b77e1b90-39ce-4ae0-b3f8-97018d9b3426)

- Tried to deploy with unverified account without verification trials and got `no verification` error

![image](https://github.com/user-attachments/assets/4ef054ad-846e-4e04-b27b-246a4a06d9fc)

#### Success scenario

- Tried with a verified account and I could deploy without any errors

Tried a local running nodes `255` on zos4 and `197` on zos3

PR #2445, opened 2024-10-24
only allow deployments of verified users

Eslam-Nawara:
### Description

only allow deployments of verified users

### Related Issues

- #2444 

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/pkg/gridtypes/deployment.go
+++ b/pkg/gridtypes/deployment.go
@@ -5,11 +5,8 @@ import (
 	"crypto/ed25519"
 	"crypto/md5"
 	"encoding/hex"
-	"encoding/json"
 	"fmt"
 	"io"
-	"net/http"
-	"time"
 
 	sr25519 "github.com/ChainSafe/go-schnorrkel"
 	"github.com/gtank/merlin"
@@ -17,8 +14,10 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
-// ErrWorkloadNotFound error
-var ErrWorkloadNotFound = fmt.Errorf("workload not found")
+var (
+	// ErrWorkloadNotFound error
+	ErrWorkloadNotFound = fmt.Errorf("workload not found")
+)
 
 const (
 	SignatureTypeEd25519 = "ed25519"
@@ -33,10 +32,8 @@ type Verifier interface {
 	Verify(msg []byte, sig []byte) bool
 }
 
-type (
-	Ed25519VerifyingKey []byte
-	Sr25519VerifyingKey []byte
-)
+type Ed25519VerifyingKey []byte
+type Sr25519VerifyingKey []byte
 
 func (k Ed25519VerifyingKey) Verify(msg []byte, sig []byte) bool {
 	return ed25519.Verify([]byte(k), msg, sig)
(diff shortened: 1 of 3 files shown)


Eslam-Nawara:
This pr would depend on the deployment of kyc server implemented [here](https://github.com/threefoldtech/tfgrid-kyc-verifier/pull/4) having this [requirements](https://github.com/threefoldtech/tfgrid-kyc-verifier/issues/5#issuecomment-2434853707), **it should not be merged before the deployment of the server with the required endpoint**

Merging the pr would cause that no unverified user would be able to use any zos3 to deploy any solution, **so it shouldn't be merged before the dashboard has a way to verify the twins**

Also we need to verify the twin used in github workflows so the test can run deployments

pkg/gridtypes/deployment.go
```diff
@@ -665,3 +672,46 @@ func (o JobOperation) String() string {
 		return "unknown"
 	}
 }
+
+// getTwinVerificationState make sure the account used is verified we have the user public key in bytes(pkBytes)
+func getTwinVerificationState(twinID uint32) (status string) {
+	verificationServiceURL := "https://kyc1.gent01.dev.grid.tf/api/v1/status"
```
xmonader: this url should be retrieved from stack information

ashraffouda (dismissed):

pkg/provision/engine.go
```diff
@@ -1162,3 +1189,57 @@ func (e *NativeEngine) GetWorkloadStatus(id string) (gridtypes.ResultState, bool
...
+		Result struct {
+			Status string `json:"status"`
+		} `json:"result"`
+		Error string `json:"error"`
+	}
+
+	err = json.Unmarshal(body, &result)
+	if err != nil {
+		return
+	}
```
muhamadazmy: You can create a json Decoder with the body and hence you will not have to read it in memory first

pkg/provision/engine.go
```diff
@@ -1162,3 +1189,57 @@ func (e *NativeEngine) GetWorkloadStatus(id string) (gridtypes.ResultState, bool
 
 	return wl.Result.State, true, nil
 }
+
+// getTwinVerificationStatus make sure the account used is verified we have the user public key in bytes(pkBytes)
+func getTwinVerificationStatus(twinID uint32) (status string) {
+	status = "FAILED"
```
muhamadazmy: Please use a const instead. If possible values can only be success/failure then it's better to use a boolean maybe

Define a new type alias so it's not confused with actual strings

pkg/provision/engine.go
```diff
@@ -1012,6 +1024,18 @@ func (n *NativeEngine) CreateOrUpdate(twin uint32, deployment gridtypes.Deployme
 		return fmt.Errorf("twin id mismatch (deployment: %d, message: %d)", deployment.TwinID, twin)
 	}
 
+	// make sure the account used is verified
+	check := func() error {
+		if getTwinVerificationStatus(twin) != "VERIFIED" {
```
muhamadazmy: Define constants or even better create a new type alias

pkg/provision/engine.go
```diff
@@ -1162,3 +1189,57 @@ func (e *NativeEngine) GetWorkloadStatus(id string) (gridtypes.ResultState, bool
...
+
+	err = json.Unmarshal(body, &result)
+	if err != nil {
+		return
+	}
+
+	if response.StatusCode != http.StatusOK {
+		log.Error().Msgf("failed to verify user status: %s", result.Error)
+		return
+	}
```
muhamadazmy: This check needs to be done before you actually deserialize the body, no? If status is not okay a big chance the body is not json

muhamadazmy (dismissed):

rawdaGastan (approved):

Issue #2446, opened 2024-10-24
prepare script to migrate existing macvlans on nodes


Omarabdul3ziz:
prepare script to migrate existing macvlans on nodes.

since canceling zdb workload doesn't remove the instance/netns, a script is required to migrate all the wiring to use veth pairs instead of the existing macvlan

related issue: https://github.com/threefoldtech/zos/issues/2403

Omarabdul3ziz:
### update
script is working now with myc but still fixing an issue in ygg routing rules

Eslam-Nawara:
Verified
connected to node 11 on dev net and followed the verification steps here and now zdbs network is setup with veth pairs instead of macvlan
![image](https://github.com/user-attachments/assets/54737469-b9b4-4ef3-a91a-3e0e1e44b18d)
for node 256 (zos4 node)
![image](https://github.com/user-attachments/assets/a8b0ecfa-b18d-41eb-89b7-75dc2696a7bd)

PR #2449, opened 2024-10-28
zdb network migration

Omarabdul3ziz:
- add a migration script to rewire zdb network setup with veth pairs instead of macvlan
- run the script after the old setup is done
- in case no zdb setup found or it is initialized with veth the migration script should do no modifications
- https://github.com/threefoldtech/zos/issues/2446

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/pkg/network/migrate_macvlan.go
+++ b/pkg/network/migrate_macvlan.go
@@ -1,0 +1,173 @@
+package network
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/containernetworking/plugins/pkg/ns"
+	"github.com/threefoldtech/zos/pkg/network/ifaceutil"
+	"github.com/threefoldtech/zos/pkg/network/namespace"
+	"github.com/threefoldtech/zos/pkg/network/options"
+	"github.com/vishvananda/netlink"
+)
+
+const (
+	zdbNsPrefix = "zdb-ns-"
+	macvlanType = "macvlan"
+)
+
+type Veth struct {
+	name      string // iface name inside the namespace
+	bridgeIdx int    // bridge index on the host
+	addresses []netlink.Addr
+	routes    []netlink.Route
+}
+
+func (n *networker) MigrateZdbMacvlanToVeth() error {
+	netNss, err := namespace.List(zdbNsPrefix)
+	if err != nil {
+		return fmt.Errorf("failed to list namespaces with prefix %q: %w", zdbNsPrefix, err)
+	}
+	if len(netNss) != 1 {
+		return fmt.Errorf("should found only one namespace with prefix %q, found %v", zdbNsPrefix, len(netNss))
+	}
+
+	netNs, err := namespace.GetByName(netNss[0])
+	if err != nil {
+		return fmt.Errorf("failed to get namespace with name %q: %w", netNss[0], err)
+	}
+	defer netNs.Close()
+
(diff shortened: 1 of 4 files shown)


pkg/network/migrate_macvlan.go
```diff
@@ -0,0 +1,173 @@
...
+	routes    []netlink.Route
+}
+
+func (n *networker) MigrateZdbMacvlanToVeth() error {
+	netNss, err := namespace.List(zdbNsPrefix)
+	if err != nil {
+		return fmt.Errorf("failed to list namespaces with prefix %q: %w", zdbNsPrefix, err)
+	}
+	if len(netNss) != 1 {
+		return fmt.Errorf("should found only one namespace with prefix %q, found %v", zdbNsPrefix, len(netNss))
```
xmonader: should find

pkg/network/migrate_macvlan.go
```diff
@@ -0,0 +1,173 @@
...
+			bridgeIdx: l.Attrs().ParentIndex,
+			addresses: addrs,
+			routes:    routes,
+		})
+	}
+
+	return veths, nil
+}
+
+func deleteOldLink(name string) error {
```
xmonader, line 121: I think we should log here

pkg/network/migrate_macvlan.go
```diff
@@ -0,0 +1,173 @@
...
+
+	var veths []Veth
+	if err := netNs.Do(func(_ ns.NetNS) error {
+		veths, err = getOldInterfaces()
+		return err
+	}); err != nil {
+		return fmt.Errorf("failed to get old links from namespace %q: %w", filepath.Base(netNs.Path()), err)
+	}
+
+	for _, veth := range veths {
```
xmonader, line 55: this can happen concurrently (but I assume with the usage it maybe ok)

pkg/network/migrate_macvlan.go
```diff
@@ -0,0 +1,173 @@
...
+			return fmt.Errorf("failed to get master by index %v: %w", veth.bridgeIdx, err)
+		}
+
+		if err := netNs.Do(func(_ ns.NetNS) error {
+			return deleteOldLink(veth.name)
+		}); err != nil {
+			return fmt.Errorf("failed to remove old link in namespace: %w", err)
+		}
+
+		if err := ifaceutil.MakeVethPair(veth.name, master.Attrs().Name, 1500, netNs); err != nil {
```
xmonader: extract the MTU value into a const

xmonader (approved):

xmonader:
Lgtm

Omarabdul3ziz:
wait to catch/deploy a zdb instance to test the migration on dev net

LeeSmet (commented):
I assume it should be fine due to presence of other links, but perhaps there should also be  a check to make sure the used bridge is up at the end of the conversion?

Omarabdul3ziz:
to verify the script. i chose a node that has the old network setup (node 14 on dev)
image

i deployed a zdb instance which is wired with macvlan
image

after the node got the new version `648200db` with the script. the setup has changed to be veth
image

and the instance is still reachable 
image

Omarabdul3ziz:
@LeeSmet noticed the comment after the merge.

I think the script shouldn't affect the bridge. If the old setup is working, it means the used bridges (br-pub/br-my/br-ygg) are already up.

Issue #2429, opened 2024-09-24
networkd: extend dhcp timeout on failure


maxux:
The DHCP request always tries for 20 seconds and then timeouts.
https://github.com/threefoldtech/zos/blob/main/pkg/network/dhcp/dhcp.go#L50-L51

On networks that have switches with spanning-tree enabled or anything related, these ports can take several seconds before starting to forward and 20 seconds after the port is brought up can be too short.

Just extending this timeout would increase boot time for everybody, which is not the best idea either. I think a good compromise would be testing with a small timeout first then increase it if it fails.

xmonader:
Can this wait till the next release? or is it a blocker?

maxux:
AFAIK this can wait, farmer affected fixed their switch.

Eslam-Nawara:
azmy is advising against adding another retry mechanism on top of udhcpc command check https://github.com/threefoldtech/zos/pull/2450#pullrequestreview-2407711981

PR #2450, opened 2024-10-28
extend dhcp timeout on failure

Eslam-Nawara:
### Description

extend dhcp timeout on failure

### Related Issues

- #2429 

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/pkg/network/dhcp/dhcp.go
+++ b/pkg/network/dhcp/dhcp.go
@@ -43,12 +43,22 @@ func (p *ProbeOutput) IPNet() (*net.IPNet, error) {
 
 func Probe(ctx context.Context, inf string) (output ProbeOutput, err error) {
 	// use udhcpc to probe the interface.
-	// this depends on that the interface is UP
+	// this check that the interface is UP before running udhcpc
+
+	interfaceByName, err := net.InterfaceByName(inf)
+	if err != nil {
+		return output, err
+	}
+
+	if (interfaceByName.Flags & net.FlagUp) == 0 {
+		return output, fmt.Errorf("failed to probe interface %s interface is not up", inf)
+	}
 
 	var stdout bytes.Buffer
 	var stderr bytes.Buffer
 
 	t := 10
+	var maxRetries uint64 = 5
 	check := func() error {
 		args := []string{
 			"-i", inf, // the interface to prope
@@ -68,7 +78,10 @@ func Probe(ctx context.Context, inf string) (output ProbeOutput, err error) {
 		return cmd.Run()
 	}
 
-	if err := backoff.Retry(check, backoff.NewExponentialBackOff()); err != nil {
+	if err := backoff.Retry(
+		check,
+		backoff.WithMaxRetries(backoff.NewExponentialBackOff(), maxRetries),
+	); err != nil {
 		return output, errors.Wrapf(err, "failed to probe interface '%s': %s", inf, stderr.String())
 	}
 
--- a/pkg/network/dhcp/dhcp.go
+++ b/pkg/network/dhcp/dhcp.go
@@ -42,21 +42,21 @@ func (p *ProbeOutput) IPNet() (*net.IPNet, error) {
 }
 
 func Probe(ctx context.Context, inf string) (output ProbeOutput, err error) {
-	// use udhcpc to prope the interface.
+	// use udhcpc to probe the interface.
 	// this depends on that the interface is UP
 
 	var stdout bytes.Buffer
 	var stderr bytes.Buffer
 
-	t := 1
+	t := 10
 	check := func() error {
 		args := []string{
 			"-i", inf, // the interface to prope
-			"-q",      // exist once lease is optained
+			"-q",      // exit once lease is optained
 			"-f",      // foreground
-			"-T", "1", // every second
-			"-t", fmt.Sprint(t), // send 't' dhcp queries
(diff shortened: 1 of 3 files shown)


pkg/network/dhcp/dhcp.go
```diff
@@ -43,22 +44,31 @@ func (p *ProbeOutput) IPNet() (*net.IPNet, error) {
...
 	var stderr bytes.Buffer
 
-	cmd.Stdout = &stdout
-	cmd.Stderr = &stderr
-	if err := cmd.Run(); err != nil {
+	t := 1
+	check := func() error {
+		args := []string{
+			"-i", inf, // the interface to prope
+			"-q",      // exist once lease is optained
```
xmonader: Exit after obtaining lease

pkg/network/dhcp/dhcp.go
```diff
@@ -43,22 +44,31 @@ func (p *ProbeOutput) IPNet() (*net.IPNet, error) {
...
+			"-f",      // foreground
+			"-T", "1", // every second
```
xmonader: This pauses between the packets timeout

pkg/network/dhcp/dhcp.go
```diff
@@ -43,22 +44,31 @@ func (p *ProbeOutput) IPNet() (*net.IPNet, error) {
...
+			"-t", fmt.Sprint(t), // send 't' dhcp queries
+			"-s", "/usr/share/udhcp/probe.script", // use the prope script
```
xmonader: probe

pkg/network/dhcp/dhcp.go
```diff
@@ -43,22 +44,31 @@ func (p *ProbeOutput) IPNet() (*net.IPNet, error) {
...
-	cmd.Stderr = &stderr
-	if err := cmd.Run(); err != nil {
+	t := 1
+	check := func() error {
+		args := []string{
+			"-i", inf, // the interface to prope
+			"-q",      // exist once lease is optained
+			"-f",      // foreground
+			"-T", "1", // every second
+			"-t", fmt.Sprint(t), // send 't' dhcp queries
```
xmonader: dhcp queries -> dhcp discover packets

pkg/network/dhcp/dhcp.go
```diff
@@ -43,22 +44,31 @@ func (p *ProbeOutput) IPNet() (*net.IPNet, error) {
...
+
+		cmd := exec.CommandContext(ctx, "udhcpc", args...)
+		cmd.Stdout = &stdout
+		cmd.Stderr = &stderr
+
+		t += 5
+		return cmd.Run()
+	}
+
+	if err := backoff.Retry(check, backoff.NewExponentialBackOff()); err != nil {
```
xmonader: should we set max retries here?

pkg/network/dhcp/dhcp.go
```diff
@@ -43,22 +44,31 @@ func (p *ProbeOutput) IPNet() (*net.IPNet, error) {
...
 		return output, errors.Wrapf(err, "failed to prope interface '%s': %s", inf, stderr.String())
```
xmonader: failed to probe interface

pkg/network/dhcp/dhcp.go
```diff
@@ -43,22 +44,31 @@ func (p *ProbeOutput) IPNet() (*net.IPNet, error) {
 func Probe(ctx context.Context, inf string) (output ProbeOutput, err error) {
 	// use udhcpc to prope the interface.
```
xmonader: probe

pkg/network/dhcp/dhcp.go
```diff
@@ -43,22 +44,31 @@ func (p *ProbeOutput) IPNet() (*net.IPNet, error) {
...
 	// this depends on that the interface is UP
```
xmonader: should we have a guard here so we don't bother if the interface is down?

pkg/network/dhcp/dhcp.go
iwanbk: consider to add `misspell` to the golangci-lint config https://golangci-lint.run/usage/linters/#misspell

pkg/network/dhcp/dhcp.go
```diff
@@ -41,25 +42,47 @@ func (p *ProbeOutput) IPNet() (*net.IPNet, error) {
...
+
+		t += 5
+		return cmd.Run()
+	}
+
+	if err := backoff.Retry(
+		check,
+		backoff.WithMaxRetries(backoff.NewExponentialBackOff(), maxRetries),
+	); err != nil {
+		return output, errors.Wrapf(err, "failed to probe interface '%s': %s", inf, stderr.String())
```
muhamadazmy, line 85: The udhcpc command already retries and you can control how many times and delay between sends via it's argument (the `-t` and `-T`) options I am wondering if it's necessary to implement another retry mechanism on top

muhamadazmy (changes requested):
I think this is the wrong place to implement retries. If you check the `internet` binary you will see that it implements its own retry on top of this `prob` The probe is has it's own retry mechansim with udhcpc command.

It's up to the user of the probe to decide. for example `internet` never gives up until it has an ip. Other users might give up or retry periodically that's where they can retry the Probe!

Issue #2439, opened 2024-10-21
Compare the node clock to tfchain instead of third party API


scottyeager:
We are attempting to use `worldtimeapi.org` to check that nodes have functioning NTP. However, the service seems pretty unreliable and there are a lot of errors in node logs about not being able to connect to it. I noticed in a quick test just now that I got a 502 error when trying to load the json response and the website itself was very slow to respond.

Instead of introducing a dependency on an external service, why not just compare the node's clock against the timestamps of blocks on tfchain? It's accurate enough and also highly available.

xmonader:
multiple strategies got added including tfchain in #2457

PR #2457, opened 2024-10-31
support multiple timeservers for the ntpcheck

xmonader:
### Description

Supports multiple time servers along with worldtimeapi

### Changes

Adds timeapi and worldclockapi 

### Related Issues

- #2439 
- #2219 
- #2464

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/pkg/perf/healthcheck/ntp.go
+++ b/pkg/perf/healthcheck/ntp.go
@@ -50,7 +50,6 @@ func RunNTPCheck(ctx context.Context) {
 func ntpCheck(ctx context.Context) error {
 	z := zinit.Default()
 	zcl, err := perf.TryGetZbusClient(ctx)
-
 	if err != nil {
 		return fmt.Errorf("ntpCheck expects zbus client in the context and found none %w", err)
 	}
@@ -76,8 +75,14 @@ func getCurrentUTCTime(zcl zbus.Client) (time.Time, error) {
 		Func func() (time.Time, error)
 	}
 
-	// List of time servers
+	// List of time servers, and here not in the global vars, so we can inject zcl to pass to getTimeChainWithZCL
 	var timeServers = []TimeServer{
+		{
+			Name: "tfchain",
+			Func: func() (time.Time, error) {
+				return getTimeChainWithZCL(zcl)
+			},
+		},
 		{
 			Name: "worldtimeapi",
 			Func: getWorldTimeAPI,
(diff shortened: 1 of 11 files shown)


xmonader:
Another consideration could be that we get the block at the current height 

```
// GetBlockAtCurrentHeight retrieves the block at the current chain height.
func GetBlockAtCurrentHeight() (*types.SignedBlock, error) {
	substrateclient := .....
    // Get the current block height.
	height, err := substrateclient.GetCurrentHeight()
	if err != nil {
		return nil, err
	}

	// Fetch the block by its number.
	return substrateclient.GetBlockByNumber(types.U32(height))
}
```

and from there we can compare the timestamps

xmonader:
image

cc @sameh-farouk

sameh-farouk:
> cc @sameh-farouk

You can use something like this but I prefer to make it a fallback approach, instead of depending mainly on it.


```go
func GetTimestampFromTFChain(tfchainClient *tfchain.Substrate) (time.Time, error) {
	api, meta, err := tfchainClient.GetClient()
	if err != nil {
	}
	var timestamp types.Moment
	key, err := types.CreateStorageKey(meta, "Timestamp", "Now", nil, nil)
	if err != nil {
	}
	// Retrieve the timestamp.now storage value
	ok, err := api.RPC.State.GetStorageLatest(key, &timestamp)
	if err != nil {
	}
	if !ok {
	}
	return timestamp.Time, nil
}
}
```

EDIT:
just call `Time()` method on `tfchain.Substrate`. It will do the same as the code I wrote above.

rawdaGastan (dismissed):

sameh-farouk:
Great job overall!
I think we could implement it as a backup approach to help lighten the load on our RPC nodes. Even though they might not be fully utilized at the moment, it could be really beneficial down the line!

xmonader:
> Great job overall! I think we could implement it as a backup approach to help lighten the load on our RPC nodes. Even though they might not be fully utilized at the moment, it could be really beneficial down the line!

I applied your suggestion :)

pkg/perf/healthcheck/ntp.go
```diff
@@ -57,20 +68,97 @@ func ntpCheck() error {
...
+			Name: "timeapi.io",
+			Func: getTimeAPI,
+		},
+		{
+			Name: "tfchain",
+			Func: func() (time.Time, error) {
+				return getTimeChainWithZCL(zcl)
+			},
+		},
+	}
```
muhamadazmy: Can this be defined as a global `var` (in ntp.go)

muhamadazmy (dismissed):
Amazing work! I left a minor comment. Feel free to ignore

pkg/perf/healthcheck/ntp.go
xmonader: Problem is I couldn't find a nicer way to inject zcl or otherwise I would have changed the signature of all of them to accept zbus client :(

Omarabdul3ziz (approved):

rawdaGastan (approved):

Issue #2480, opened 2024-11-10
add get location endpoint


Omarabdul3ziz:
original issue: https://github.com/threefoldtech/tfgrid-sdk-go/issues/1248

PR #2481, opened 2024-11-10
add location endpoint

Omarabdul3ziz:
### Description
- write node location info on a file while registering
- add location endpoint to get this file info

### Related Issues
- https://github.com/threefoldtech/zos/issues/2480

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/pkg/registrar/register.go
+++ b/pkg/registrar/register.go
@@ -3,8 +3,10 @@ package registrar
 import (
 	"context"
 	"crypto/ed25519"
+	"encoding/json"
 	"fmt"
 	"net"
+	"os"
 	"time"
 
 	"github.com/centrifuge/go-substrate-rpc-client/v4/types"
@@ -144,6 +146,10 @@ func registerNode(
 		City:      info.Location.City,
 	}
 
+	if err := writeLocationOnFile(info.Location, geoip.LocationFile); err != nil {
+		return 0, 0, errors.Wrap(err, "failed to set location on disk")
+	}
+
 	log.Info().Str("id", mgr.NodeID(ctx).Identity()).Msg("start registration of the node")
 	log.Info().Msg("registering node on blockchain")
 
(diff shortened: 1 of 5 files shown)


pkg/registrar/register.go
```diff
@@ -231,3 +237,13 @@ func ensureTwin(ctx context.Context, substrateGateway *stubs.SubstrateGatewayStu
 
 	return twinID, nil
 }
+
+func writeLocationOnFile(loc geoip.Location, filepath string) error {
+	file, err := os.OpenFile(filepath, os.O_RDWR|os.O_CREATE, 0644)
```
iwanbk: is it possible for the file to already exists?
If possible, what is the default behavior? Is truncate or append?

pkg/zos_api/location.go
```diff
@@ -0,0 +1,26 @@
...
+	"context"
+	"encoding/json"
+	"os"
+
+	"github.com/pkg/errors"
+	"github.com/threefoldtech/zos/pkg/geoip"
+)
+
+func (g *ZosAPI) locationGet(ctx context.Context, payload []byte) (interface{}, error) {
+	if _, err := os.Stat(geoip.LocationFile); err != nil {
```
iwanbk: is it needed?

xmonader (changes requested):
This should be addressed differently, I don't see a problem calling that function from the api and returning its result to the user

pkg/zos_api/location.go
```diff
@@ -2,25 +2,10 @@ package zosapi
...
-	f, err := os.Open(geoip.LocationFile)
-	if err != nil {
-		return nil, errors.Wrap(err, "couldn't get the location info")
-	}
-	defer f.Close()
-
-	var loc geoip.Location
-	err = json.NewDecoder(f).Decode(&loc)
-	return loc, err
+	return geoip.Fetch()
```
xmonader: at this point you can add caching if needed

Omarabdul3ziz:
i will draft until we finish releasing today

pkg/zos_api/location.go
muhamadazmy: I agree, the node location should be updated like once per day and then cached for full day before it's fetched again.

muhamadazmy (dismissed):
Thank @Omarabdul3ziz for implementing this. It looks good to me. It's yet a good idea to cache the location of the node to avoid un-necessary load on the location service as per @xmonader comment

Eslam-Nawara (approved):

iwanbk (approved):

xmonader (approved):

Issue #2491, opened 2024-11-17
Review rollingupdate Substrate usage


xmonader:
It should be using substrate only through api-gateway, not sure if that's dependent of noded or not?

PR #2493, opened 2024-11-17
use api gateway instead of chain client in getting version in zos upgrade

rawdaGastan:
### Related Issues

- https://github.com/threefoldtech/zos/issues/2491

### DIFF
--- a/pkg/upgrade/upgrade.go
+++ b/pkg/upgrade/upgrade.go
@@ -20,9 +20,10 @@ import (
 	"github.com/threefoldtech/0-fs/meta"
 	"github.com/threefoldtech/0-fs/rofs"
 	"github.com/threefoldtech/0-fs/storage"
-	substrate "github.com/threefoldtech/tfchain/clients/tfchain-client-go"
+	"github.com/threefoldtech/zbus"
 	"github.com/threefoldtech/zos/pkg/app"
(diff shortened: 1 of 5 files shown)


pkg/upgrade/upgrade.go
```diff
@@ -112,6 +107,15 @@ func NoZosUpgrade(o bool) UpgraderOption {
 	}
 }
 
+// ZbusClient option, adds a zbus client to the upgrader
+func ZbusClient(cl zbus.Client) UpgraderOption {
```
xmonader, line 111: WithZbusClient

pkg/upgrade/upgrade.go
```diff
@@ -56,20 +57,13 @@ type ChainVersion struct {
 	Version       string `json:"version"`
 }
 
-func getRolloutConfig(env environment.Environment) (ChainVersion, []uint32, error) {
+func getRolloutConfig(ctx context.Context, gw *stubs.SubstrateGatewayStub) (ChainVersion, []uint32, error) {
```
xmonader, line 60: Are we sure that this is where it belongs? not on the api-gateway itself?

pkg/upgrade/upgrade.go
xmonader: let's stick with it for consistency

xmonader (approved):

Issue #2361, opened 2024-06-14
see if we can switch to retryable http 


xmonader:
https://github.com/hashicorp/go-retryablehttp

for cases like retrieving nodes ffor iperf or submitting uptime reports or whatever that can be retried (needs to have max number of retries, we don't want it to be stuck forever retrying)

Nabil-Salah:
# 17-12-2024

Exploring zos codebase to find out where I need to switch to retriable HTTP

tried to ssh to local zos node

Nabil-Salah:
# 18-12-2024

Implementation retryable http to needed ones

Investigated openstack

Nabil-Salah:
# 19-12-2024

resolved pr comments

complete Investigation in OpenStack

![Screenshot from 2024-12-19 16-50-26](https://github.com/user-attachments/assets/9a8dfc80-79eb-45d1-b2e8-f223e6ab937d)

zos working fine on qemu

Nabil-Salah:
# 22-12-2024

Complete investigation on OpenStack components

Nabil-Salah:
# 24-12-2024

Complete investigation on OpenStack components
Started preparing slides

Nabil-Salah:
# 25-12-2024

Introduce openstack session with salma
worked on a demo for the openstack

Nabil-Salah:
# 29-12-2024
Worked on openstack hands on
worked on openstack potential ideas report

Nabil-Salah:
# 1-1-2025

Implementation retryable http to needed ones on zos4

![Screenshot from 2025-01-01 18-17-11](https://github.com/user-attachments/assets/0e48e0f6-8319-4773-80c5-2e99dab3203b)

Used retryable HTTP in more cases on zos

![image](https://github.com/user-attachments/assets/df80fbd2-d442-4588-8bec-9cffbb3f3bd6)

PR #2515, opened 2024-12-18
Switch to retriable HTTP

Nabil-Salah:
### Description

This pr makes use of retriable HTTP requests instead of the default HTTP server 

### Changes
HTTP server in :
`pkg/perf/publicip/publicip_task.go`
`pkg/environment/config.go`
`pkg/flist/flist.go`
`pkg/gateway/metrics.go`
`pkg/provision/engine.go`
`pkg/upgrade/hub/hub.go`
`pkg/vm/client.go`

### Related Issues

[issue#2361](https://github.com/threefoldtech/zos/issues/2361
)
### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/pkg/vm/client.go
+++ b/pkg/vm/client.go
@@ -45,7 +45,7 @@ func (c *Client) Shutdown(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	response, err := c.client.Do(&retryablehttp.Request{Request: request})
+	response, err := c.client.StandardClient().Do(request)
 	if err != nil {
 		return errors.Wrap(err, "error calling machine shutdown")
 	}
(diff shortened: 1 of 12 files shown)


pkg/vm/client.go
```diff
@@ -43,7 +45,7 @@ func (c *Client) Shutdown(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	response, err := c.client.Do(request)
+	response, err := c.client.Do(&retryablehttp.Request{Request: request})
```
ashraffouda: I see from the docs we can still use the standard client with retryable client. if so, no need to wrap the request like the above
https://github.com/hashicorp/go-retryablehttp#getting-a-stdlib-httpclient-with-retries
so if we needed to change the client to something else no need to change those lines again

ashraffouda (changes requested):

ashraffouda (approved):

Issue #2523, opened 2025-01-12
add zos4 farms for the managed farms list


Omarabdul3ziz:
farms:
- 3997 on mainnet
- 43 on qa

add on:
- ssh keys in zos4
- test farms in zos-config

PR #2524, opened 2025-01-12
add LiriaFarm as a managed mainnet farm

Omarabdul3ziz:
### Description

- add LiriaFarm as a managed mainnet farm

### issues
- https://github.com/threefoldtech/zos/issues/2523


Omarabdul3ziz:
no need to add it in zos3 src

PR #2544, opened 2025-02-27
remove test.yaml workflow

Omarabdul3ziz:
- running pkg/ tests done on zosbase
- building cmds/ done on publish.yaml


Omarabdul3ziz:
we agreed that pkg related to zos only will be returned here, so it is better to keep the workflow

Issue #2571, opened 2025-05-26
let's get config from zos-config instead of hard coded values


ashraffouda:
currently we have alot of config managed from environment.go like chainurls, activation urls ....
it is better if we kept those values as default if we didn't find values in zos-config and depend on zos config instead which will keeps changes to min when urls changed/added or removed wihtout changing everytime in codebase

Eslam-Nawara:
### WIP
- updated config.go and environment.go to have all configuration from zos-config
- removed all hardcoded urls and updated the node to get urls from zos-config
- updated registrar module to reload config on every registration retry
- working on updating the node so that if config changes with time, it restarts the services depending on the updated config.

Eslam-Nawara:
### WIP
- now all urls are fully configurable from zos-config
- I think we need to continue working on restarting the modules with changes to config in a new pr as this one has many changes
- need to discuss with ashraf the changes we need to do to be able to update the node urls if the config changed while the node is running.

Eslam-Nawara:
wip
- updated substrate gateway to be able to update the open connection used by the module
- updated zos-api module so we can update the service if any of substrate url or relay url changed

Eslam-Nawara:
work completed, 
- updated zosbase to load config from zos-config on demand
- updated zos module api-gateway and noded to monitor changes in sub urls and relay urls and update modules if needed

Eslam-Nawara:
updated the pr https://github.com/threefoldtech/zosbase/pull/33 with the newest changes to main and added back the multiple []hub URLs and testing the node features

>> Note: this expects hub URLs to be []string not only string

deploying VM:

<img width="1712" height="510" alt="Image" src="https://github.com/user-attachments/assets/d33dba30-f498-4692-a337-9c9a05c94768" />

and the VM is accessible

<img width="974" height="675" alt="Image" src="https://github.com/user-attachments/assets/08014c23-65ab-4c21-a317-89037e908fea" />

Node logs:

<img width="1905" height="1077" alt="Image" src="https://github.com/user-attachments/assets/f12edbab-06d0-4a75-a6d6-7848c274d7dd" />

zui

<img width="748" height="491" alt="Image" src="https://github.com/user-attachments/assets/7ac73012-27ee-4e75-86f3-32f2e632cad4" />

Eslam-Nawara:
Updated 3 pr to the latest on main and was able to  boot a node with the new changes

[...]

PR #2583, opened 2025-07-02
Update modules to update with config

Eslam-Nawara:
### Description

support modules updating substrate/relay connections if the url in zos-config was updated.

### Changes

- update api-gateway module to if sub/relay urls changed
- update noded module to update sub connection in evnets package

### Related Issues

- https://github.com/threefoldtech/zos/issues/2571

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/bootstrap/bootstrap/src/hub.rs
+++ b/bootstrap/bootstrap/src/hub.rs
@@ -14,7 +14,7 @@ struct ZosConfig {
     hub_urls: Vec<String>,
 }
 
-fn get_hub_url(runmode: &RunMode) -> Result<Vec<String>> {
+fn get_hub_urls(runmode: &RunMode) -> Result<Vec<String>> {
     let base_url = "https://github.com/threefoldtech/zos-config/raw/main/";
     let config_filename = match runmode {
         RunMode::Prod => "production.json",
@@ -43,10 +43,10 @@ fn get_hub_url(runmode: &RunMode) -> Result<Vec<String>> {
         Err(_) => return Ok(fallback),
     };
 
-    if config.hub_url.is_empty() {
+    if config.hub_urls.is_empty() {
         Ok(fallback)
     } else {
-        Ok(config.hub_url)
+        Ok(config.hub_urls)
     }
 }
 
@@ -86,7 +86,7 @@ impl Repo {
         T: AsRef<str>,
     {
         let config = crate::config::Config::current()?;
-        let hub = get_hub_url(&config.runmode)?;
+        let hub = get_hub_urls(&config.runmode)?;
         Ok(Repo {
             name: String::from(name.as_ref()),
             hub,
--- a/bootstrap/bootstrap/src/hub.rs
+++ b/bootstrap/bootstrap/src/hub.rs
@@ -11,7 +11,7 @@ use std::path::Path;
 
 #[derive(Deserialize)]
 struct ZosConfig {
-    hub_url: Vec<String>,
+    hub_urls: Vec<String>,
 }
 
 fn get_hub_url(runmode: &RunMode) -> Result<Vec<String>> {
--- a/cmds/modules/api_gateway/main.go
+++ b/cmds/modules/api_gateway/main.go
@@ -111,12 +111,10 @@ func action(cli *cli.Context) error {
 	bo.MaxElapsedTime = 0
 
 	// this ctx is used to allow the node to restart the peer without leaving any unwanted open connections
-	peerCtx, cancel := context.WithCancel(ctx)
-	defer cancel()
-
+	currentPeerCtx, cancel := context.WithCancel(ctx)
 	backoff.Retry(func() error {
 		_, err = peer.NewPeer(
-			peerCtx,
+			currentPeerCtx,
 			hex.EncodeToString(pair.Seed()),
 			manager,
 			router.Serve,
@@ -125,6 +123,9 @@ func action(cli *cli.Context) error {
 			peer.WithInMemoryExpiration(6*60*60), // 6 hours
 		)
 		if err != nil {
+			if cancel != nil {
+				cancel()
+			}
 			return fmt.Errorf("failed to start a new rmb peer: %w", err)
 		}
 
@@ -143,16 +144,6 @@ func action(cli *cli.Context) error {
 			return nil
 			// check if we need to run an update on the peer and only do the update if all the changes are done successfully
 		case <-time.After(10 * time.Minute):
-
-			log.Debug().Msg("checking if node is maintaining a healty substrate connection")
-			cl, _, err := manager.Raw()
-			if err == nil {
-				// skip update if the connection is working properly
-				log.Debug().Msg("the open connection is healty, no update needed")
-				cl.Client.Close()
-				continue
-			}
-
 			env, err := environment.Get()
 			if err != nil {
 				// skip update if we can't get env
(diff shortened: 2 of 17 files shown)


Eslam-Nawara:
blocked on https://github.com/threefoldtech/zosbase/pull/38

bootstrap/bootstrap/src/hub.rs
```diff
@@ -11,10 +11,10 @@ use std::path::Path;
 
 #[derive(Deserialize)]
 struct ZosConfig {
-    hub_url: Vec<String>,
+    hub_urls: Vec<String>,
```
Omarabdul3ziz, line 14: i see it is still `hub_url` in qa and beyond, it is only changed in dev. so we need to be careful releasing this to other nets https://github.com/threefoldtech/zos-config/blob/51b732620739b3392ed118f593a528fe2f15c1d8/qa.json#L85

Omarabdul3ziz: or i can suggest saving a backward compatible state, by having both `hub_url` and `hub_urls` in the config file, until all nets are using the plural field then we can remove the other

cmds/modules/api_gateway/main.go
```diff
@@ -128,6 +137,85 @@ func action(cli *cli.Context) error {
...
+
+			env, err := environment.Get()
+			if err != nil {
+				// skip update if we can't get env
+				log.Debug().Err(err).Msg("failed to load node environment")
+				continue
+			}
+
+			updatedSubURLs := env.SubstrateURL
+			updatedRelayURLs := environment.GetRelaysURLs()
```
Omarabdul3ziz: i think we don't need a special handling for getting relay anymore, since all config now are fetched from the config repo, we can have updated relay from `env.relaysUrls`. this will make another fetch for config. right?

cmds/modules/api_gateway/main.go
```diff
@@ -128,6 +137,85 @@ func action(cli *cli.Context) error {
...
+			// skip update if substrate and relay urls did not change
+			if slices.Equal(subURLs, updatedSubURLs) && slices.Equal(relayURLs, updatedRelayURLs) {
+				log.Debug().Msg("zos-config doesn't have updated config to update the node with")
+				continue
+			}
+
+			log.Debug().Strs("relays_urls", updatedRelayURLs).Strs("substrate_urls", updatedSubURLs).Msg("detected new update in configuration")
+			newPeerCtx, newCancel := context.WithCancel(ctx)
+
+			newManager, err := environment.GetSubstrate()
```
Omarabdul3ziz: this also make another fetch to config, can't we pass the subUrls already fetched?

cmds/modules/api_gateway/main.go
```diff
@@ -128,6 +137,85 @@ func action(cli *cli.Context) error {
...
+					newCancel()
+					continue
+				}
+			}
+
+			cancel()
+			peerCtx = newPeerCtx
+			cancel = newCancel
+
+			manager = newManager
```
Omarabdul3ziz: correct me if i got it wrong, but this doesn't grantee the manger passed to NewZosApi is updated? it will still have the old manager with the old subUrls?

cmds/modules/provisiond/main.go
```diff
@@ -365,6 +365,7 @@ func action(cli *cli.Context) error {
 	// call the runtime upgrade before running engine
 	if err := provisioners.Initialize(ctx); err != nil {
 		log.Error().Err(err).Msg("failed to run provisioners initializers")
+		return err
```
Omarabdul3ziz, line 368: i know it make sense, but it was like that for a long time. are we sure we need to return here

cmds/modules/api_gateway/main.go
```diff
@@ -128,6 +137,85 @@ func action(cli *cli.Context) error {
...
+			if !slices.Equal(subURLs, updatedSubURLs) {
+				err = gw.UpdateSubstrateGatewayConnection(newManager)
+				if err != nil {
+					log.Debug().Err(err).Msg("failed to update substrate gateway with new manager")
+					newCancel()
+					continue
+				}
+			}
+
+			cancel()
```
Omarabdul3ziz: we should call this `cancel` closing old peer, before calling `peer.NewPeer`. if the old peer was still active it will break both connections

cmds/modules/api_gateway/main.go
```diff
@@ -128,6 +137,85 @@ func action(cli *cli.Context) error {
...
+				router.Serve,
+				peer.WithKeyType(peer.KeyTypeEd25519),
+				peer.WithRelay(updatedRelayURLs...),
+				peer.WithInMemoryExpiration(6*60*60), // 6 hours
+			); err != nil {
+				log.Debug().Err(err).Msg("failed to start a new rmb peer")
+				continue
+			}
+
+			if !slices.Equal(subURLs, updatedSubURLs) {
```
Omarabdul3ziz: newPeer already created with the newManger, so we need to update the gw to use the newManger anyway?

cmds/modules/noded/main.go
```diff
@@ -245,6 +247,50 @@ func action(cli *cli.Context) error {
...
+			newSubURLs := newEnv.SubstrateURL
+			slices.Sort(subURLs)
+			slices.Sort(newSubURLs)
+
+			if slices.Equal(subURLs, newSubURLs) {
+				log.Debug().Msg("zos-config doesn't have updated config to update the node with")
+				continue
+			}
+
+			newSub, err := environment.GetSubstrate()
```
Omarabdul3ziz: correct me, i see GetSubstrate do the exact checks above? fetch urls, assure urls changed, check the connection, and neturn new manager

Omarabdul3ziz (changes requested):
can you check please, if i got something wrong

bootstrap/bootstrap/src/hub.rs
Eslam-Nawara: I think yes we can do that, have both hub_url and hub_urls in all networks until it's released in all networks

cmds/modules/api_gateway/main.go
Eslam-Nawara: yes, I agree

cmds/modules/provisiond/main.go
Eslam-Nawara: yes I discussed it with azmy and ashraf and ashraf agreed we will enforce retry here for now

cmds/modules/api_gateway/main.go
Eslam-Nawara: why woould it break the old connection? it was done like that to make sure we don't stop the old peer until the new one is created and working fine

cmds/modules/api_gateway/main.go
Eslam-Nawara: the new manager is created only if the urls changed

bootstrap/bootstrap/src/hub.rs
ashraffouda: lets change that and keep the json file have both fields

Omarabdul3ziz: still not updated in zos-config

cmds/modules/api_gateway/main.go
```diff
@@ -128,6 +138,75 @@ func action(cli *cli.Context) error {
...
+		case <-time.After(10 * time.Minute):
+			env, err := environment.Get()
+			if err != nil {
+				// skip update if we can't get env
+				log.Debug().Err(err).Msg("failed to load node environment")
+				continue
+			}
+
+			updatedSubURLs := env.SubstrateURL
+			updatedRelayURLs := environment.MustGet().RelaysURLs
```
Omarabdul3ziz: `env.RelayURLs`?

cmds/modules/api_gateway/main.go
Omarabdul3ziz: it is a limitation with relay/peer, if a peer have two connections with different relay at the same time it will keep overriding each other and eventually broke

cmds/modules/noded/main.go
```diff
@@ -245,6 +247,38 @@ func action(cli *cli.Context) error {
 		}
 	}()
 
+	// monitor node env updates in substrate url
+	go func() {
```
Omarabdul3ziz, line 251: afai understand, 
- api_gateway: check for changes on substrate/relay urls changes => update the rmb peer
- noded: check for changes on substrate urls => update the substrate manager

what do you think about having this check run separately to check once, and send update signal on channel like this routine does.

Omarabdul3ziz (commented):
i added some non-critical comments

bootstrap/bootstrap/src/hub.rs
ashraffouda: https://github.com/threefoldtech/zos-config/blob/main/development.json#L96

Omarabdul3ziz (approved):
agreed to do the enhancements later

Issue #2593, opened 2025-07-21
Errors while rendering Zos console


scottyeager:
A farmer reported seeing this error flash on screen while booting up a Zos node on mainnet:

<img width="1435" height="459" alt="Image" src="https://github.com/user-attachments/assets/d3f9112d-f538-45e4-9a7b-29b7d26edf3b" />

SalmaElsoly:
### Work Completed:
Based on my investigation, this issue occurred due to rendering a table widget whose rows were not initialized. I believe it's a timing issue related to re-rendering widgets (services and headers) while others are still uninitialized (netgrid and resources).
I added a check to render only initialized widgets

SalmaElsoly:
Work Completed:
handled in zos4 & zoslight

PR #2602, opened 2025-08-03
Error rendering zos console

SalmaElsoly:
### Description

Describe the changes introduced by this PR and what does it affect

### Changes

added check for rendering initialized widgets in zui/main.go

### Related Issues

https://github.com/threefoldtech/zos/issues/2593#issue-3249839203

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/cmds/modules/zui/main.go
+++ b/cmds/modules/zui/main.go
@@ -108,6 +108,7 @@ func action(ctx *cli.Context) error {
 		log.Error().Err(err).Msg("failed to start services renderer")
 	}
 
+	var isInitialized atomic.Bool
 	go func() {
 		<-done
 
@@ -125,10 +126,11 @@ func action(ctx *cli.Context) error {
 		if err := netRender(client, netgrid, &flag); err != nil {
 			log.Error().Err(err).Msg("failed to start net renderer")
 		}
-
+		
 		if err := resourcesRender(client, resources, &flag); err != nil {
 			log.Error().Err(err).Msg("failed to start resources renderer")
 		}
+		isInitialized.Store(true)
 	}()
 
 	mod := zui.New(ctx.Context, errorsParagraph, &flag)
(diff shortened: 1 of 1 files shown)


cmds/modules/zui/main.go
```diff
@@ -142,6 +144,10 @@ func action(ctx *cli.Context) error {
 	}()
 
 	render := func() {
+		if !isInitialized.Load() {
```
ashraffouda, line 147: I'm not sure if u were able to reproduce this or not 
r u sure the issue is in netgrid, resources or not?
as a suggestion if we r not sure which element produces this we may harden all widgets with something like this 
```
render := func() {
    // Check if widgets are initialized before rendering
    widgets := make([]ui.Drawable, 0, 5)
    
    if header != nil {
        widgets = append(widgets, header)
    }
    if services != nil {
        widgets = append(widgets, services)
    }
    if netgrid != nil {
        widgets = append(widgets, netgrid)
    }
    if resources != nil {
        widgets = append(widgets, resources)
    }
    if errorsParagraph != nil {
        widgets = append(widgets, errorsParagraph)
    }
    
    if len(widgets) > 0 {
        ui.Render(widgets...)
    }
}
```
I'm also not able to reproduce but let's harden all?
also maybe we don't have to add isintialiezed bool in this case

ashraffouda (changes requested):

cmds/modules/zui/main.go
ashraffouda: also this need to be handled in zoslight/zos4

SalmaElsoly: The issue occurs when rendering a table on the screen. A re-render is triggered while a nested table widget is still uninitialized with it's rows. Specifically, the package attempts to access index 0 of the table’s rows before they are ready.and the part which initializes tables in a go routine is netgrid and resources, that's why I added isinitialized to make sure that even nested components are completely initialized

ashraffouda (approved):

ashraffouda:
plz handle this on zoslight/zos4

